Issue #24, opened 2016-08-05
Symbol search doesn't list expected items


DanTup:
I don't think the method I'm using is the correct one; since it doesn't show all the things inside a class (I guess the clue was in the name, "top level").

devoncarew:
I think this is the best analyzer API to use, to map to the symbol search. We may still want the API extended to show more things (`Foo.ba` would show something like `Foo.bar`).

DanTup:
Do you think it's worthwhile me sending two requests (as mentioned [here](https://groups.google.com/a/dartlang.org/forum/#!topic/analyzer-discuss/conkKX7XiKI)) in the meantime? I think Code users would probably expect class methods/properties to appear in the search and it feels a bit limited as it is.

devoncarew:
I haven't used the other APIs - are they used for searching for a member
given a class? Or for any member in a project?

There's a separate vscode api for symbols within a file. That may make the
most sense to map to the analyzers outline api.

On Aug 6, 2016 7:27 AM, "Danny Tuppeny" <notifications@github.com> wrote:

Do you think it's worthwhile me sending two requests (as mentioned here
<https://groups.google.com/a/dartlang.org/forum/#!topic/analyzer-discuss/conkKX7XiKI>)
in the meantime? I think Code users would probably expect class
methods/properties to appear in the search and it feels a bit limited as it
is.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/DanTup/Dart-Code/issues/24#issuecomment-238025785>, or mute
the thread

DanTup:
findMemberDeclarations looks very similar to the one I'm currently using; it's just global search for declerations with that name (though it might only do full matches and no pattern). I might have a go at merging results from both and see how it feels; certainly the current behaviour doesn't feel as useful as I would expect. I guess I could also compare results with some sample code on both TS and Dart.

DanTup:
I did also post to [analyzer-discuss](https://groups.google.com/a/dartlang.org/forum/#!topic/analyzer-discuss/conkKX7XiKI) if it's more helpful for others if the discussion happens there.

DanTup:
@devoncarew I've implemented merging of these two for now. Seems to work ok for me (though I discovered a crazy bug in the process - https://github.com/Microsoft/vscode/issues/10446) but if you have a minute to test it I'd be interested in your opinion.

[...]

PR #27, opened 2016-08-06
Make workspace symbol search case-insensitive

devoncarew:
- allow case insensitive search in the workspace symbol dialog
- add some common logic to convert from analysis server locations to vscode ones (fix https://github.com/DanTup/Dart-Code/issues/10)

screen shot 2016-08-06 at 12 00 20 am

### DIFF
--- a/src/dart_diagnostic_provider.ts
+++ b/src/dart_diagnostic_provider.ts
@@ -2,6 +2,7 @@
 
 import { DiagnosticCollection, Diagnostic, DiagnosticSeverity, Uri, Range, Position } from "vscode";
 import { Analyzer } from "./analyzer";
+import { toRange } from "./utils";
 import * as as from "./analysis_server_types";
 
 export class DartDiagnosticProvider {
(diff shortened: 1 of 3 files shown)


DanTup:
Great, thanks!

I think I'm actually using the wrong method here (top-level) as I would expect it to find methods in classes (for example). I presume this logic all applies the same if I change that? (#24).

DanTup:
When I re-read this code, I was worried about the wildcards. It means "cat" will match "cart". However, VS Code seems to do further filtering on the results that come back and only allow missing characters if they're lowercase and preceed an uppercase character.

Eg:

- "TA" matches "TwitterApi"
- "twittera" matches "TwitterApi"
- "twitera" does match "TwitterApi" in the analyzer (I see the results come back in the debug window) but VS Code appears to hide the result

I think this actually works out really well for the user though I'm wondering whether on a large codebase it'd generate a lot of unnecessary results over the wire. Any thoughts @devoncarew?

devoncarew:
Yeah, we're filtering results - all the top level symbols - and then vscode is applying a fuzzy match on top of that. I do like how it works out - it'll often show you the match you were thinking about w/o you having to be too precise.

I'm not super worried about the wire traffic - there are other parts of the protocol that are pretty verbose as well. We could measure it and get a sense of whether it's impacting the performance.

DanTup:
Cool; you're more familiar with it than I so if you're not concerned, sounds good to me!

I'm not sure about only doing top-level results though, I posted in the [analyser-discuss group here](https://groups.google.com/a/dartlang.org/forum/#!topic/analyzer-discuss/conkKX7XiKI), would be interested in your opinion!

PR #36, opened 2016-08-06
adjust console logging; set editing prefs

devoncarew:
- specify to use tabs w/ the project
- add a flag to control analysis server traffic logging
- minor tweaks to the logging

Some of the changes are more style related; happy to iterate on this patch.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
+    "editor.tabSize": 4,
+    "editor.insertSpaces": false,
     "files.exclude": {
         "out": true,
         "**/node_modules": true
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -20,7 +20,8 @@ export class Analyzer extends AnalyzerGen {
 
 		this.analyzerProcess.stdout.on("data", (data: Buffer) => {
 			let message = data.toString();
-			if (verbose && message.trim().length != 0) console.log(`<== ${message}`);
+			if (verbose && message.trim().length != 0)
+				console.log(`<== ${message}`);
 
 			// Add this message to the buffer for processing.
 			this.messageBuffer.push(message);
@@ -57,7 +58,8 @@ export class Analyzer extends AnalyzerGen {
 
 	private sendMessage<T>(req: Request<T>) {
 		let json = JSON.stringify(req);
-		if (verbose) console.log(`==> ${json}`);
+		if (verbose)
+			console.log(`==> ${json}`);
 		this.analyzerProcess.stdin.write(json);
 		this.analyzerProcess.stdin.write("\n");
 	}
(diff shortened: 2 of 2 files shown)


.vscode/settings.json
```diff
@@ -1,5 +1,7 @@
 // Place your settings in this file to overwrite default and user settings.
 {
+    "editor.tabSize": 4,
+    "editor.insertSpaces": false,
```
DanTup: I actually just noticed there was a weird mismatch in my files.. I think I set `insertSpaces: false` but the default behaviour scans the existing file and my `extension.ts` (and somehow `util.ts`) have spaces.

4-char tabs is my preference and I planned on sorting this

:+1: !

src/analyzer.ts
```diff
@@ -18,16 +20,14 @@ export class Analyzer extends AnalyzerGen {
 
 		this.analyzerProcess.stdout.on("data", (data: Buffer) => {
 			let message = data.toString();
-			console.log(`RCV: ${message}`);
+			if (verbose && message.trim().length != 0) console.log(`<== ${message}`);
```
DanTup: I'm really not a fan of `if`s all on one line like this, easy to miss things (possibly Kiln's bad indenting in code reviews is to blame, it sometimes indents completely differently to the source file!).

I don't go as far as putting braces around single-line ifs, though I've discovered in Dart I've had to start doing that to stop the formatter doing this!

Maybe I'll learn to deal with it; I don't know :D

DanTup:
All seems fine to me (though I'd prefer to break the ifs onto two lines).

Dumping things to the console was useful when I started, but the debug console in Code is so bad (doesn't show long things, is really sluggish and copy/paste doesn't work properly for more than the visible height) that I'm thinking of changing this to dump to disk. Hopefully there's a way to only do this from the dev extension host; doesn't make much sense on users machines (unless they're trying to help diagnose something).

DanTup:
Thanks! :)

src/analyzer.ts
devoncarew: I moved the if bodies to their own lines. Happy to adjust to whatever coding style - definitely more important that the code in a repo shares a similar style.

devoncarew:
> that I'm thinking of changing this to dump to disk

You could set up a hidden pref to specify a file to log to (`dart.development.serverLog`) - that way you could log during development but it wouldn't be something that's exposed to users -

DanTup:
Does Code support hidden config options? I get completion suggestions on the others, haven't seen anything about hiding them. Sounds like a good option if there is. I've created #37 as a task to figure something out.

devoncarew:
If you define them in package.json they'll show up in that overall, synthesized settings file. I _think_ you can read from any value though, so reading from `dart.fooBar` but not listing it in the package spec should effectively create a hidden pref.

DanTup:
Gotcha!

PR #38, opened 2016-08-06
filter todos from the problems view

devoncarew:
- filter todos from the problems view (and add a preferences for it)
- add a hidden flag to start the analysis server diagnostic server up on a specified port

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,6 +70,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Forces indenting with two spaces when Dart files are opened. This is on by default because VS Code doesn't support per-language settings and most people use tabs/4 spaces for other languages."
+				},
+				"dart.showTodos": {
+					"type": "boolean",
+					"default": false,
+					"description": "Show TODOs in the Problems view."
 				}
 			}
 		}
--- a/src/analyzer.ts
+++ b/src/analyzer.ts
@@ -4,6 +4,7 @@ import * as vscode from "vscode";
 import * as child_process from "child_process";
 import * as as from "./analysis_server_types";
 import { AnalyzerGen } from "./analyzer_gen";
+import { getConfig } from "./utils";
 
 let verbose: boolean = true;
 
@@ -15,8 +16,17 @@ export class Analyzer extends AnalyzerGen {
 
 	constructor(dartVMPath: string, analyzerPath: string) {
 		super();
+
 		console.log(`Starting Dart analysis server...`);
-		this.analyzerProcess = child_process.spawn(dartVMPath, [analyzerPath]);
+
+		let args = [analyzerPath];
+
+		// Optionally start the analyzer's diagnostic web server on the given port.
+		let port = getConfig('analyzerDiagnosticsPort');
+		if (port)
+			args.push(`--port=${port}`);
+
+		this.analyzerProcess = child_process.spawn(dartVMPath, args);
 
 		this.analyzerProcess.stdout.on("data", (data: Buffer) => {
 			let message = data.toString();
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -70,6 +70,11 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Forces indenting with two spaces when Dart files are opened. This is on by default because VS Code doesn't support per-language settings and most people use tabs/4 spaces for other languages."
+				},
+				"dart.showTodos": {
+					"type": "boolean",
+					"default": false,
```
DanTup: Slightly torn about this default... It could be quite noisy if you have a lot, but it's also a nice reminder you should deal with this stuff. If we default to off I suspect most people will never know the option exists, but without a way to filter it in the Problems window maybe it'll just be more annoying than anything.

I just Google'd to see if I can find what other languages do.. Didn't find any evidence of any showing TODOs, but I did find this (watch the animated GIF) which looks quite cool (but doesn't seem like something a Dart-specific extension would be expected to do!)

https://github.com/MattiasPernhult/vscode-todo

Let's leave as is for now and hope maybe we'll get a filter or something in future. I guess the file-search is a reasonable substitute if you do want to go and fix them (which I really need to on this codebase already!).

src/analyzer.ts
```diff
@@ -15,8 +16,17 @@ export class Analyzer extends AnalyzerGen {
...
+
 		console.log(`Starting Dart analysis server...`);
-		this.analyzerProcess = child_process.spawn(dartVMPath, [analyzerPath]);
+
+		let args = [analyzerPath];
+
+		// Optionally start the analyzer's diagnostic web server on the given port.
+		let port = getConfig('analyzerDiagnosticsPort');
+		if (port)
+			args.push(`--port=${port}`);
```
DanTup: What is this magic? I can't see any mention of a diagnostics port on the docs? !

package.json
devoncarew: > Let's leave as is for now

I think you're right about the discovery issue - people won't know to turn it on.

That todo extension looks nice. It does seem like you want to work with todos occasionally, and not all the time like errors or warnings.

DanTup: What about collapsing the TODOs into a single line?

> You have 123 TODOs. Set dart.showTodos in settings to have them displayed here

I think having one item in the list won't be a problem and it aids discovery. If we can hook the user clicking on it (I suspect not) we could even do "click to show" and then fill them in.

src/analyzer.ts
devoncarew: hmm, it may not be documented - it's not really a user facing feature though - just used by the analysis server team when things go badly wrong.  It's useful when trying to get enough info to diagnose a problem. Also useful when building an analysis server client - you can see what it thinks the state of the overlays are, the priority sources, ...

package.json
DanTup: Not sure how well this'd work - we have to give Code a `Uri` for a set. It might not accept a null or fake filename; possibly could attribute them against pubspec or similar, but then it's starting to feel a bit hacky :(

devoncarew: I guess I'm leaning towards enabling it by default, and assuming that people will see the setting if they want to disable it.

Right now it takes a restart before the setting really takes effect - it filters as the analysis errors are created, and doesn't remember them to re-report to vscode.

src/analyzer.ts
DanTup: That sounds really useful! Before I realised you were using 1-based lines/characters I was sending updates badly and getting really strange behaviour - being able to see the servers understanding of a file would've made that much easier =D

I'll have a nose next time I'm at PC!

DanTup:
Merging this in because I think it's all good. I'll raise an issue about investigating options for showing TODOs.

package.json
DanTup: I'm happy for the default to be on until we have a better option (and like you say, people more likely to look for an option if they don't like it). Not applying immediately sucks a little, but we could send an `analysis.reanalyze` from `vscode.workspace.onDidChangeConfiguration`? (though it looks like we don't get told which values changed; we'd have to always do it of stash a copy of the previous value)

devoncarew: +1 to reanalyzing if that setting changes, it'll be nice to not have the UI
out of sync with the pref setting

DanTup: Ok, sounds like default=on and reanalyze on change seems best for now. I've updated #39 for handling this.

PR #42, opened 2016-08-07
Show analysis status in status bar (+ other tweaks)

### LINKED ISSUE #39: Default to showing TODOs; reanalyze when config changes
See discussion in #38.

devoncarew:
- show analysis status (`Analyzing...`) in the status line when the analysis server is busy
- default showing todos to true (fix #39)
- reanalyze sources when the setting for showing todos changes
- don't call `serverShutdown` - if I understand correctly it's basically a no-op
- fix an issue w/ tooltips where some elements (string constants) would show a small, no content tooltip
- only send overlays for files that the analysis server needs to be aware of

@DanTup, it's a grab bag of work. Probably it for this weekend! I'm excited that Dart's getting a plugin for vscode :)

### DIFF
--- a/src/dart_diagnostic_provider.ts
+++ b/src/dart_diagnostic_provider.ts
@@ -17,7 +17,7 @@ export class DartDiagnosticProvider {
 
 	private handleErrors(notification: as.AnalysisErrorsNotification) {
 		let errors = notification.errors;
-		if (!getConfig('showTodos'))
+		if (!getConfig<boolean>('showTodos'))
 			errors = errors.filter((error) => error.type != 'TODO');
 		this.diagnostics.set(
 			Uri.file(notification.file), 
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -1,23 +1,29 @@
 "use strict";
 
-import * as vscode from "vscode";
 import * as path from "path";
 import * as util from "./utils";
+import * as vscode from "vscode";
 import { Analyzer } from "./analyzer";
-import { DartFormattingEditProvider } from "./dart_formatting_edit_provider";
-import { DartHoverProvider } from "./dart_hover_provider";
 import { DartCompletionItemProvider } from "./dart_completion_item_provider";
 import { DartDefinitionProvider } from "./dart_definition_provider";
 import { DartDiagnosticProvider } from "./dart_diagnostic_provider";
+import { DartFormattingEditProvider } from "./dart_formatting_edit_provider";
+import { DartHoverProvider } from "./dart_hover_provider";
+import { DartIndentFixer } from "./dart_indent_fixer";
 import { DartWorkspaceSymbolProvider } from "./dart_workspace_symbol_provider";
 import { FileChangeHandler } from "./file_change_handler";
-import { DartIndentFixer } from "./dart_indent_fixer";
+import { ServerStatusNotification } from "./analysis_server_types";
 
 const DART_MODE: vscode.DocumentFilter = { language: "dart", scheme: "file" };
 const stateLastKnownSdkPathName = "dart.lastKnownSdkPath";
 
 let dartSdkRoot: string;
 let analyzer: Analyzer;
+
+let showTodos: boolean;
+
+let statusBarItem: vscode.StatusBarItem;
+let statusShowing: boolean = false;
 
 export function activate(context: vscode.ExtensionContext) {
 	console.log("Dart Code activated!");
@@ -41,6 +47,11 @@ export function activate(context: vscode.ExtensionContext) {
 	}
 
 	analyzer = new Analyzer(path.join(dartSdkRoot, util.dartVMPath), path.join(dartSdkRoot, util.analyzerPath));
+	analyzer.serverSetSubscriptions({
+		subscriptions: ['STATUS']
+	});
+	analyzer.registerForServerStatus(handleServerStatus);
+
 	// TODO: Check if EventEmitter<T> would be more appropriate than our own.
 
 	// Set up providers.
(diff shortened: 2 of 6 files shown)


src/dart_diagnostic_provider.ts
```diff
@@ -17,7 +17,7 @@ export class DartDiagnosticProvider {
 
 	private handleErrors(notification: as.AnalysisErrorsNotification) {
 		let errors = notification.errors;
-		if (!getConfig('showTodos'))
+		if (!getConfig<boolean>('showTodos'))
```
DanTup: I can't remember why, I tried to standardise on double quotes for strings. I seemed to be randomly mixing them and it kinda bugged me. Maybe I thought I'd be putting apostrophes in English strings more and I guess they're also slightly more distinguishable from template-strings (backticks).

You don't have to worry about changing this but I thought it worth mentioning (or maybe there's a reason I should switch!?)

DanTup: I'd also been putting config names into constants; though currently they're littered around the place and probably should go together somewhere; I'll probably pull them all together soon into a strongly-typed class.

src/extension.ts
```diff
@@ -75,6 +86,41 @@ export function activate(context: vscode.ExtensionContext) {
 	let dartIndentFixer = new DartIndentFixer();
 	context.subscriptions.push(vscode.window.onDidChangeActiveTextEditor(td => dartIndentFixer.onDidChangeActiveTextEditor(td)));
 	dartIndentFixer.onDidChangeActiveTextEditor(vscode.window.activeTextEditor); // Handle already-open file.
+
+	context.subscriptions.push(vscode.workspace.onDidChangeConfiguration(handleConfigurationChange));
+}
+
+function handleServerStatus(status: ServerStatusNotification) {
```
DanTup: I think it'd be worth pulling all this into another file/class to keep `extension.ts` clean as it's a good place to understand how everything it's wired up but this will clutter it a little.

src/file_change_handler.ts
```diff
@@ -71,4 +80,12 @@ export class FileChangeHandler {
 			id: "" // TODO: Fix this, should be optional!
 		}
 	}
+
+	private isDartLike(document: vscode.TextDocument): boolean {
```
DanTup: I posted [a question about this on the analyzer-discuss list](https://groups.google.com/a/dartlang.org/forum/#!topic/analyzer-discuss/AOPDiiBjxKI) and Brian suggested a couple more file types are monitored. I was sending them all for now thinking the service would just drop those it didn't care about (other than with the multi-cursor hack the traffic is minimal).

If you think it's worth doing, I think we should include all the files Brian mentioned [here](https://groups.google.com/a/dartlang.org/d/msg/analyzer-discuss/AOPDiiBjxKI/XM7naJ8KBwAJ).

DanTup:
Looks great! I've added a couple of minor comments. If I get some enough time today I might merge this in and tweak those things myself; would be nice to get another build out with all the changes!

Minor nitpick: I tend to prefer smaller individual commits for functionally different things (easier to describe them in the commit message and to attribute them to different issues). Don't feel that strongly about it (esp for such small changesets) but thought it worth noting. Maybe I'll add some notes to the contributing file.

Thanks again; really appreciate the effort (and it's great to have an experienced Dart dev's input, I'm still a Dart noob). I'm also really keen on a good IDE integration; I've really not got on with the other IDEs (and bad IDE support was actually one of the big factors in us not adopting Dart at work).. I was sad when CDE was abandoned (I still am a little, since Code won't run on ChromeOS!).

src/dart_diagnostic_provider.ts
devoncarew: > I'll probably pull them all together soon into a strongly-typed class.

Sounds great - I have the same string literal (`showTodos`) in 2-3 places now; having a class for the pref name would be more robust.

DanTup: Yep; just working on this now.. Looks something like this:

```ts
import { config } from "./config";

if (config.showTodos)
```

:)

src/file_change_handler.ts
devoncarew: I think it'd be useful to filter - to just create overlays for things we know are dart/dartlike files. +1 to improving this list to what's in that email: `'*.dart', '*.html', '*.htm', '.analysis_options', and 'analysis_options.yaml'`.

src/dart_diagnostic_provider.ts
DanTup: [Sorted](https://github.com/DanTup/Dart-Code/commit/e06b06ab7df3f817d5a82da6e7da3e3a996e31b4).

I also moved `verbose` to config (hidden). This way you can turn on and off without a recompile, it applies immediately. It's off by default (since hiddens don't have defaults) but that seems reasonable.

devoncarew:
> Minor nitpick: I tend to prefer smaller individual commits for functionally different things (easier to describe them in the commit message and to attribute them to different issues).

Totally makes sense. I generally try and be hygienic w/ my PRs and just send one per topic. Easy enough to do going forward :)

src/dart_diagnostic_provider.ts
devoncarew: > it applies immediately

ah, that's very nice. the analysis server output can be very useful but high volume. It's great to be able to toggle it live.

src/file_change_handler.ts
DanTup: [Done](https://github.com/DanTup/Dart-Code/commit/f81cb14a0844d4f69966cba728d88a67001b6a6e) this using the langageID and added `basename()` around the filename since it was getting the full path.

src/extension.ts
DanTup: [Done this](https://github.com/DanTup/Dart-Code/commit/e3211ed86ed827679bb7263a78e620c20ae24afc); hope it seems reasonable!

DanTup:
Thanks :)

I've tweaked some stuff based on my comments above, believe everything mentioned has been done in this PR. Thanks again!

Issue #20, opened 2016-08-05
Add basic analytics to track extension usage


DanTup:
It'd be good to have some idea of how often features are used. Investigate sending some basic events to Analytics counting activations and use of major features.

DanTup:
`env.machineId` might be useful.

PR #43, opened 2016-08-07
Add basic Analytics

DanTup:


DanTup:
@devoncarew I added some basic analytics (to see how many people are using and which versions - I don't thing VSCode auto-updates or prompts uses to). After the discussion about TODOs I also added an event for TODOs being enabled/disabled (this is easily extendable to other toggles).

I also set an isDevelopment flag when in the dev host (not sure if I did this the best way...) so I can filter these out in Analytics.

The numbers are obviously 0 for now, but if you're interested in access to the Analytics account let me know.

DanTup:
This fixes #20.

src/analytics.ts
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	logActivation() { this.log(EventCategory.Extension, EventAction.Activated); }
+	logShowTodosToggled(enabled: boolean) { this.log(EventCategory.TODOs, enabled ? EventAction.Enabled : EventAction.Disabled); }
+
+	private log(category: EventCategory, action: EventAction) {
+		let data = {
+			v: "1", // API Version.
+			tid: "UA-2201586-19",
+			cid: env.machineId,
```
devoncarew: I don't know if this value will work. When I run this locally I see `'someValue.machineId'. analytics wants this value to be something like a UUID  v4 - https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#clientId.

There's some sample code for creating one here: https://github.com/dart-lang/usage/blob/master/lib/src/uuid.dart.

devoncarew:
Awesome, analytics in great to have. It can be sensitive for some, so probably best to add an opt-out as a preference (VSCode has an opt-out for this as well).

Adding a preference would also allow us to filter out dev hosts as well.

src/analytics.ts
devoncarew: It could be that env.machineId _is_ normally in a good form for analytics to use, but theres something funky going on on my machine -

DanTup: I actually say the same thing in the debugger and wondered if it was the dev host. I made a note to double-check what's really being sent from the release version but haven't had chance yet. I'll be sure to do it tonight!

If it doesn't work, I'll create a random ID and stash it in the extension context to use instead (and raise a bug with MS!)

DanTup:
I did add something to the readme but hoped it was basic and anonymous enough people wouldn't mind. I was trying to decide whether it's reasonable to demand some basic analytics in return for using a free extension, though if GA is logging IPs then maybe it's less anonymous than I hoped :(

There is a dev flag passed (as long as the version ends with `-dev`, but if MS implement this I'll use it to detect extension host: https://github.com/Microsoft/vscode/issues/10272#issuecomment-238144406). In the analytics account I have a view for "All" (which includes dev, to aid testing) and "Live" (filters dev). Turning off in dev would mean having to turn on (and the log live stats) while testing which isn't ideal.

DanTup:
I've added a setting to opt-out of analytics.

There is a dev flag passed to GA (based on version number until MS implement [this](https://github.com/Microsoft/vscode/issues/10272#issuecomment-238144406)).

In the analytics account I have a view for "All" (which includes dev, to aid testing) and "Live" (filters dev) so you don't need to disable it if you're just worried about polluting stats (I think missing live usage would be worse), though obviously the option is there if you wish!

Looking into the weird machine ID thing now...

src/analytics.ts
DanTup: Ok, confirmed that `machineId` is a big long GUID in the real host; it's only that placeholder in the extension host. That might make a better check than my "-dev" version check so I'll see if I can get MS to confirm if that's by design and stable.

devoncarew: Ah, cool; it's nice to not have to write code to generate one.

PR #44, opened 2016-08-07
Use instance fields

devoncarew:
- move module globals to fields
- tweak no sdk installed error message
- prefer double quotes


src/analyzer_status_reporter.ts
```diff
@@ -4,15 +4,15 @@ import { window, StatusBarItem, Disposable } from "vscode";
...
-let statusShowing: boolean;
-
 export class AnalyzerStatusReporter extends Disposable {
+	statusBarItem: StatusBarItem;
+	statusShowing: boolean;
+
 	constructor(analyzer: Analyzer) {
-		statusBarItem = window.createStatusBarItem();
-		statusBarItem.text = 'Analyzing…';
+		super(() => this.statusBarItem.dispose());
```
DanTup: Heh; I put this further down thinking the reference wouldn't exist till I'd created it but of course it's not called untill later so that make no sense!

src/analyzer_status_reporter.ts
```diff
@@ -4,15 +4,15 @@ import { window, StatusBarItem, Disposable } from "vscode";
 import { Analyzer } from "./analyzer";
 import { ServerStatusNotification } from "./analysis_server_types";
 
-let statusBarItem: StatusBarItem;
-let statusShowing: boolean;
-
 export class AnalyzerStatusReporter extends Disposable {
+	statusBarItem: StatusBarItem;
```
DanTup: I think I should've actually made these private (I'm used to C#, things default to private!) so I'll try changing that later.

src/extension.ts
```diff
@@ -31,8 +31,8 @@ export function activate(context: vscode.ExtensionContext) {
 
 	dartSdkRoot = util.findDartSdk(<string>context.globalState.get(stateLastKnownSdkPathName));
 	if (dartSdkRoot == null) {
-		vscode.window.showErrorMessage("Dart Code: Could not find a Dart SDK to use. " +
-			"Please add it to your PATH or set it in the extensions settings and reload");
+		vscode.window.showErrorMessage("Could not find a Dart SDK to use. " +
+			"Please install one (www.dartlang.org) add configure the 'dart.sdkPath' setting.");
```
DanTup: If you use an installer that puts it on your PATH you shouldn't need to configure it so I think this is a bit misleading (the option is really a fallback in case the detection fails (or override to test others), which I think works fine now)

DanTup:
I think it's worth tweaking the message to not suggest the path needs configuring (if we spent effort making it less work for the user, no point telling them to do the extra work) but otherwise this all looks great 👍

src/extension.ts
devoncarew: Makes sense; I'll revert this part of the PR -

src/analyzer_status_reporter.ts
devoncarew: I updated this to use `private`.

src/extension.ts
devoncarew: update the text here

devoncarew:
updated :) ptal

DanTup:
Thanks! :)

PR #45, opened 2016-08-08
implement a document symbol provider

devoncarew:
Implement a document symbol provider based on the outline view notification. This is the 'go to symbol' command.

screen shot 2016-08-07 at 5 14 51 pm


devoncarew:
It's ~equivalent to the outline view in other tools, but w/o the UI being persistent (I'm used to using an outline view as an always open bit of UI on the right side of the editor).

src/dart_document_symbol_provider.ts
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	provideDocumentSymbols(document: TextDocument, token: CancellationToken): Thenable<SymbolInformation[]> {
+		let file = document.fileName;
+
+		this.analyzer.analysisSetSubscriptions({
+			subscriptions: { 'OUTLINE': [file] }
+		});
+
+		// TODO: Improve the presentation to make the file structure clearer?
```
DanTup: What does this relate to; all seems fine to me (I think)!

src/dart_document_symbol_provider.ts
```diff
@@ -0,0 +1,64 @@
...
+			});
+		});
+	}
+
+	private transcribeOutline(symbols: SymbolInformation[], parent, outline) {
+		let element = outline.element;
+
+		let name = element.name;
+		if (element.kind == "CONSTRUCTOR" || element.kind == "METHOD" || element.kind == "FUNCTION")
+			name = `${name}()`;
```
DanTup: In [the docs](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#type_Element) there's a `parameters` property which seems it eliminate the need for this if but also provide the parameter list?

> **parameters ( optional String )**
> The parameter list for the element. If the element is not a method or function this field will not be defined. If the element doesn't have parameters (e.g. getter), this field will not be defined. If the element has zero parameters, this field will have a value of "()".

src/dart_document_symbol_provider.ts
```diff
@@ -0,0 +1,64 @@
...
+
+		symbols.push({
+			name: name,
+			kind: getSymbolKindForElementKind(element.kind),
+			location: {
+				uri: Uri.file(element.location.file),
+				range: toRange(element.location)
+			},
+			containerName: parent == null ? null : parent.name
```
DanTup: I think Dart supports `?.` now so this could be simplified?
(though you may have copy/pasted from me; I only discovered it Dart had it recently! 😀).

src/dart_workspace_symbol_provider.ts
```diff
@@ -15,11 +15,11 @@ export class DartWorkspaceSymbolProvider implements WorkspaceSymbolProvider {
...
-				return '[]()\\-'.indexOf(c) == -1;
-    });
+			return '[]()\\-'.indexOf(c) == -1;
+		});
 		chars = chars.map((c: string) => {
-				if (c.toUpperCase() == c.toLowerCase()) return c;
-				return `[${c.toUpperCase()}${c.toLowerCase()}]`;
+			if (c.toUpperCase() == c.toLowerCase()) return c;
+			return `[${c.toUpperCase()}${c.toLowerCase()}]`;
 		});
```
DanTup: What's the reason for this change? Is the server already case-insensitive?

DanTup:
@devoncarew Looks good! I made a comment about parameters which I think would make it better?

Also - the mapping you moved to `Analyzer.ts`; does it all look good to you? I struggled to map many, so ended up picking fairly randomly (eg. `module` = `compilation unit`?!).

src/dart_document_symbol_provider.ts
devoncarew: It's not 100% clear from looking at the list when a member is contained in a class. It's a note to myself - we could remove it as well.

src/dart_document_symbol_provider.ts
devoncarew: Ah, that would probably work great.

src/dart_document_symbol_provider.ts
devoncarew: Ah, `parent?.name` is valid?

src/dart_workspace_symbol_provider.ts
devoncarew: This was just correcting some indent - a ws change.

devoncarew:
I'll take another spin through this tomorrow -

src/dart_workspace_symbol_provider.ts
DanTup: Whoops, I totally misread this - I blame the lack of side-by-side diffs (despite never using them in even in Kiln ;))

src/dart_document_symbol_provider.ts
DanTup: I haven't tried it but I read [this](http://blog.sethladd.com/2015/07/null-aware-operators-in-dart.html) and it works in DartPad (I'm not at PC to test for real) :)

devoncarew: Yeah, we have them in Dart, but I don't know about typescript :)

src/dart_document_symbol_provider.ts
DanTup: Oh, I see, I thought you meant the code presentation. Couldn't we prefix with the class name? Or isn't that what the containerName is supposed to do; does it not get rendered?

src/dart_document_symbol_provider.ts
DanTup: LOL! This is what happens when you do code reviews before getting out of bed =)

You are correct, TypeScript doesn't support it (and the request for it was denied). Sorry, my bad :)

One day this extension will be written in Dart! :D

src/dart_document_symbol_provider.ts
devoncarew: Updated - this works great (using the parameters field).

devoncarew:
I took a look at the mapping - I updated one mapping for an enum_value from a constant to an enum. I'm not sure what a `label` in the analysis server represents. The other choices look good :) I think this is mostly about mapping things for the purpose of displaying icons, so rough approximations are ok :)

src/dart_document_symbol_provider.ts
devoncarew: I removed this TODO: - looking at this, I think we're returning the right info from the POV of vscode; we'll rely on them to present things well so that user's can easily navigate the file.

devoncarew:
updated -

src/dart_document_symbol_provider.ts
```diff
@@ -0,0 +1,62 @@
...
+			});
+		});
+	}
+
+	private transcribeOutline(symbols: SymbolInformation[], parent: as.Element, outline: as.Outline) {
+		let element = outline.element;
+
+		let name = element.name;
+		if (element.parameters)
+			name = `${name}()`;
```
DanTup: What I meant here was, can't we do ``name = `${name}${parameters}``` (depending on TS handling of undef/null, it *might* not need the if?). This way the results will show the parameters of the method too!

DanTup:
I can't merge this inline due to conflicts, will do when at PC tonight. I don't mind testing the params change to see how it looks if I get to it before you.

devoncarew:
Merged to master -

src/dart_document_symbol_provider.ts
devoncarew: The names might get a little long, but it'd be worth experimenting with, to see how it looked. You may also want to see what other language integrations do here -

DanTup: Yeah; I'll both compare with TS and see how it looks (and also review the workspace one for consistency, I don't think I added parens or anything there).

DanTup:
I think AppVeyor got an ISE from GitHub downloading the TS defs!

```
> dart-code@0.6.3-dev postinstall C:\projects\dart-code
> node ./node_modules/vscode/bin/install
Detected VS Code engine version: ^1.4.0
Error installing vscode.d.ts: Error: Request returned status code: 500
Details: Internal Server Error
npm ERR! Windows_NT 6.3.9600
npm ERR! argv "C:\\Program Files (x86)\\nodejs\\node.exe" "C:\\Program Files (x86)\\nodejs\\node_modules\\npm\\bin\\npm-cli.js" "install"
npm ERR! node v4.4.6
npm ERR! npm  v2.15.5
```

There's nothing like a reliable build system! Forcing a rebuild..

DanTup:
(Currently the only thing it's doing is running the TS compiler, though seeing if I can run automated tests there is on my list; though I'm concerned I might not be able to Code there... we'll see).

Unsurprisingly passes this time. Thanks!

src/dart_document_symbol_provider.ts
devoncarew: What I've done in full outline views in the past (where I was concerned about horizontal scrolling), was:
- `fooBar` (a field or getter)
- `fooBar()` (a method w/o args)
- `fooBar(...)` (a method w/ args; the ellipsis is used to save space but indicate that there are params there)

But probably we should just aim for consistency w/ other language integrations into vscode -

DanTup: Hmmm, TS doesn't even show parens on methods!

Typescript

What do you think? Copy, or be better?

DanTup: Also looks like they're quoting when the containing element is a filename, yet not displaying the filename for classes :/

devoncarew: Hmm, the parens help me know when something's a method vs a field - I guess I'd want to keep those.

DanTup: Here's a comparison with/without params. I think without is a bit weird (makes them look like they don't have params) however I think without colours the right one is just too busy. Prefer the left?

Params

DanTup: And the other suggestion:

Elipsis

devoncarew: I'm good w/ either (the ellipsis or the full params). The full params are useful, if a bit busy.

DanTup: ok, let's go with the full params (I guess it's most accurate, even if lack of overloads make it less confusing) and we'll see how it goes. You're likely to use it more/on bigger files than me in the short-term so if you decide it's bad, do shout (or just change).

PR #54, opened 2016-08-10
add command pub.get, pub.upgrade

devoncarew:
- add command pub.get, pub.upgrade
- have their output go to a dedicated output view at the bottom (this may not be the best long term UX)
- expose the commands on the navigator context menu, editor context menu, and in the editor toolbar when the pubspec.yaml file is open

screen shot 2016-08-10 at 5 17 04 am

screen shot 2016-08-10 at 5 17 14 am

screen shot 2016-08-10 at 5 17 21 am

These commands are active when editing a yaml file; I couldn't find a way to scope it to just a pubspec.yaml file.

### DIFF
--- a/src/pub.ts
+++ b/src/pub.ts
@@ -1,0 +1,44 @@
+'use strict';
+
+import * as channels from "./channels";
+import * as child_process from "child_process";
+import * as os from "os";
+import * as path from "path";
+import * as project from "./project";
+import * as vs from "vscode";
+
+export class PubManager {
+	private sdk: string;
+
+	constructor(sdk: string) {
+		this.sdk = sdk;
+	}
+
+	registerCommands(context: vs.ExtensionContext) {
+		context.subscriptions.push(vs.commands.registerCommand("pub.get", selection => {
+			this.runPub("get", selection);
+		}));
+		context.subscriptions.push(vs.commands.registerCommand("pub.upgrade", selection => {
+			this.runPub("upgrade", selection);
+		}));
+	}
+
+	private runPub(command: String, selection?) {
+		let root = vs.workspace.rootPath;
+		let projectPath = selection
+			? path.dirname(selection.path)
+			: project.locateBestProjectRoot();
+		let shortPath = path.join(path.basename(root), path.relative(root, projectPath));
+		let channel = channels.getCreateChannel("Pub");
+		channel.show(true);
+
+		// TODO: Add a wrapper around the Dart SDK? It could do things like
+		// return the paths for tools in the bin/ dir. 
+		let pubPath = os.platform() == "win32"
+			? path.join(this.sdk, "bin", "pub.bat")
+			: path.join(this.sdk, "bin", "pub");
+		channel.appendLine(`[${shortPath}] pub ${command}`);
+		let process = child_process.exec(`${pubPath} ${command}`, { "cwd": projectPath });
+		channels.runProcessInChannel(process, channel);
+	}
+}
--- a/media/pull-inverse.svg
+++ b/media/pull-inverse.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><rect x="3" y="2" width="10" height="5" fill="#C5C5C5"/><polygon points="9,8 7,8 7,11 5,9 5,11 8,14 11,11 11,9 9,11" fill="#75BEFF"/></svg>
--- a/media/pull.svg
+++ b/media/pull.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="16" height="16"><rect x="3" y="2" width="10" height="5" fill="#424242"/><polygon points="9,8 7,8 7,11 5,9 5,11 8,14 11,11 11,9 9,11" fill="#00539C"/></svg>
--- a/package.json
+++ b/package.json
@@ -30,7 +30,9 @@
 	"icon": "media/icon.png",
 	"activationEvents": [
 		"onLanguage:dart",
-		"workspaceContains:pubspec.yaml"
+		"workspaceContains:pubspec.yaml",
+		"onCommand:pub.get",
+		"onCommand:pub.upgrade"
 	],
 	"main": "./out/src/extension",
 	"contributes": {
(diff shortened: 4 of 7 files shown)


DanTup:
Awesome; this was next on my list to read up on!

> have their output go to a dedicated output view at the bottom (this may not be the best long term UX)

I guess we need to surface it to the user somewhere, this doesn't sound so bad to me.

> These commands are active when editing a yaml file; I couldn't find a way to scope it to just a pubspec.yaml file.

I've posted a question at https://github.com/Microsoft/vscode/issues/10401 about this, seems crazy if there's no way; really sucks to have `pub get` in the command list all the time!

Won't be around much tonight, but try and test this out and merge tomorrow. Do you think it's useful to automatically run `pub get` on saving `pubspec.yaml`?

devoncarew:
> Do you think it's useful to automatically run pub get on saving pubspec.yaml?

Yeah, I think that would be good. I was skeptical about it when we first did it for the dart editor, but users (and myself) ended up relying on and expecting the behavior :)

Having the commands in the command palette dialog is ok I think; the command looks up a directory from the current active file looking for a pubspec.yaml, and then executes in that directory.

But it would be nice to not show context menu actions on non-pubspec.yaml files. That seems a little misleading to me; you really only expect actions that are scoped to that specific file.

devoncarew:
And this PR adds a bunch of new files to the directory; we may want to think about some re-organizing.

DanTup:
> Having the commands in the command palette dialog is ok I think

It feels a bit crap them appearing if you have a Python project and they're there though!

DanTup:
I'm gonna take a look at this tonight and merge it in.

I might make a tweak though to remove the commands on the onActivation section. In https://github.com/Microsoft/vscode/issues/10401 MS confirmed there's no way to have commands only apply with a Dart project open. I checked the PowerShell extension (which adds some commands, like "expand alias") and it seems to not activate on these commands, so you just get a built-in "Command not found" error from Code when you choose these. It feels weird to load our extension just to show a nicer message. MS have marked the case as a feature request so this may improve in future (and if they have any better ideas in the meantime, I'll look at them).

We could remove the commands from non-pubspec files by creating a Pubspec`` language. I had this originally but you pointed out it was the cause I didn't get MS's syntax-highlighting. If syntax-highlighting is the only thing having `pubspec.yaml` fall back to MS's ỳaml` language gives us then we could just copy their syntax and ship it as our Pubspec language, then change the language filters to use pubspec instead of yaml?

Finally, the `locateBestProjectRoot` code... if the `pubspec` is not the root of the project, won't all the analysis be bad anyway? If so, should we just assume it's correct? If not, do we need to apply this logic more than just here?

src/pub.ts
```diff
@@ -0,0 +1,44 @@
...
+		let projectPath = selection
+			? path.dirname(selection.path)
+			: project.locateBestProjectRoot();
+		let shortPath = path.join(path.basename(root), path.relative(root, projectPath));
+		let channel = channels.getCreateChannel("Pub");
+		channel.show(true);
+
+		// TODO: Add a wrapper around the Dart SDK? It could do things like
+		// return the paths for tools in the bin/ dir. 
+		let pubPath = os.platform() == "win32"
```
DanTup: I have another platform check (for `dart.exe`/`dart`) so should probably consolidate these (I think the other is in `util`).

Actually, maybe this one isn't required (Windows checks certain extensions automatically, inc. `.bat`.. We don't need to use `dart.exe` when spwaning the server (only use it for symlink detection where we need the real filename).

DanTup:
> > Do you think it's useful to automatically run pub get on saving pubspec.yaml?
> 
> Yeah, I think that would be good

Is there a way to know if it needs running (eg. on a fresh clone of a project)? Would be nice to run at activation if we know it's needed!

devoncarew:
> I might make a tweak though to remove the commands on the onActivation section.

sgtm :)

> We could remove the commands from non-pubspec files by creating a Pubspec

My only concern here is the technical debt of forking the yaml highlighting file and then having to maintain it. It would make it easier to hang specific actions off a pubspec.yaml file.

> Finally, the locateBestProjectRoot code... if the pubspec is not the root of the project, won't all the analysis be bad anyway? If so, should we just assume it's correct? If not, do we need to apply this logic more than just here?

It's fine for there to be multiple pub packages in one project, or for there not to be a pub package at the root of a project. The analysis server will handle this case by recursing into the given analysis root and locating the child pub packages.

src/pub.ts
devoncarew: sgtm re: consolidation - I wasn't aware of the other check

devoncarew:
> Is there a way to know if it needs running (eg. on a fresh clone of a project)? Would be nice to run at activation if we know it's needed!

Yeah, you can do a timestamp check for `pubspec.yaml` vs `.packages`. If the pubspec file is newer, you know things are out-of-date.

DanTup:
I've added new issues for the things discussed here. Merging this and making the `activateOn` change. I may also shuffle some files around so you'll want to be sure to update before doing further work (you might have to resolve conflicts when updating/rebasing existing dev branches (cc @ikhwanhayat)).

src/pub.ts
DanTup: Don't need this one here, I've removed and it works fine (also means we won't break if pub ever became an exe or something).

src/pub.ts
```diff
@@ -0,0 +1,44 @@
...
+		}));
+		context.subscriptions.push(vs.commands.registerCommand("pub.upgrade", selection => {
+			this.runPub("upgrade", selection);
+		}));
+	}
+
+	private runPub(command: String, selection?) {
+		let root = vs.workspace.rootPath;
+		let projectPath = selection
+			? path.dirname(selection.path)
```
DanTup: I've had to change this from `.path` to `.fsPath` since `path` is for urls and on Windows returns things like `m:/work/` which then got prefixed with `c:\` and it all went bad. I believe it'll still work on Mac/Linux this way (the Code codebase has lots of instances of `fsPath` for stuff like this) but it'd be good to confirm.

src/pub.ts
devoncarew: I'm running on mac, so I can check this.

Issue #33, opened 2016-08-06
Check if we need to subscribe to analysis.flushResults


DanTup:
Looks like [this](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#notification_analysis.flushResults) would be sent if a file was deleted. Need to check if Code already removes listed errors for deleted files (if not, we should subscribe this and then set an empty set of errors for that file path).

ikhwanhayat:
From my observation:

1) If the file is currently active in the editor, when deleted, it's error list is cleared.

2) If the file is not opened, or opened but not active, when deleted, it's error list is NOT cleared.

If guess the safest is to listen to `analysis.flushResults` notification and remove the errors manually.

I'm currently working on this one.

DanTup:
Good stuff; thanks! :-)

PR #55, opened 2016-08-10
Remove diagnostics for deleted files.

ikhwanhayat:
Subscribe to analysis.flushResults and remove all diagnostics for deleted files.

For issue #33

### DIFF
--- a/src/dart_diagnostic_provider.ts
+++ b/src/dart_diagnostic_provider.ts
@@ -14,6 +14,9 @@ export class DartDiagnosticProvider {
 		this.diagnostics = diagnostics;
 
 		this.analyzer.registerForAnalysisErrors(es => this.handleErrors(es));
+
+		// Fired when files are deleted
+		this.analyzer.registerForAnalysisFlushResults(es => this.flushResults(es));
 	}
 
 	private handleErrors(notification: as.AnalysisErrorsNotification) {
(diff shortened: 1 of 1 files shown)


src/dart_diagnostic_provider.ts
```diff
@@ -48,4 +51,12 @@ export class DartDiagnosticProvider {
 				throw new Error("Unknown severity type: " + severity); 
 		}
 	}
+
+	private flushResults(notification: as.AnalysisFlushResultsNotification) {		
+		if (notification.files) {
```
DanTup: Do we need this `if`? The docs don't say this is optional and it seems unlikely this event would be raised without any files?

src/dart_diagnostic_provider.ts
```diff
@@ -48,4 +51,12 @@ export class DartDiagnosticProvider {
...
+			notification.files.forEach(file => {
+				this.diagnostics.delete(Uri.file(file));
```
DanTup: I think it might be better to use the [`set(entries: [Uri, Diagnostic[]][])`](https://code.visualstudio.com/docs/extensionAPI/vscode-api#DiagnosticCollection) method here (and just `.map` on files to convert). This way Code gets all the updates in one go and might avoid updating the list multiple times (though admittedly, it's probably unlikely we'd get a huge number of files in this event).

DanTup:
Good job! I've made a few minor comments but I don't mind merging and doing them myself if I get to it before you tonight. Thanks!

src/dart_diagnostic_provider.ts
ikhwanhayat: Yeah I guess no need for extra check there.

ikhwanhayat:
I've pushed changes based on your recommendations.

DanTup:
Excellent, works great. Thanks!

PR #65, opened 2016-08-12
implement a document highlighting provider

### LINKED ISSUE #64: Implement a document highlighting provider
It shows references to a symbol within a file.

devoncarew:
- implement a document highlighting provider
- fix #64 
- wrap some long lines

This uses the same technique as the document symbol provider - it turns on an analysis server notification briefly, waits for some results, turns off the notification, and converts the results to the expected vscode format.

screen shot 2016-08-11 at 4 51 16 pm

### DIFF
--- a/src/providers/dart_highlighting_provider.ts
+++ b/src/providers/dart_highlighting_provider.ts
@@ -1,0 +1,70 @@
+
+"use strict";
+
+import {
+	DocumentHighlightProvider, DocumentHighlight, TextDocument, Position, CancellationToken, Range
+} from "vscode";
+import { Analyzer } from "../analysis/analyzer";
+import * as as from "../analysis/analysis_server_types";
+
+export class DartDocumentHighlightProvider implements DocumentHighlightProvider {
+	private analyzer: Analyzer;
+	constructor(analyzer: Analyzer) {
+		this.analyzer = analyzer;
+	}
+
+	provideDocumentHighlights(
+		document: TextDocument, position: Position, token: CancellationToken
+	): Thenable<DocumentHighlight[]> {
+		let file = document.fileName;
+		let offset = document.offsetAt(position);
+
+		this.analyzer.analysisSetSubscriptions({ subscriptions: { "OCCURRENCES": [file] } });
+
+		return new Promise<DocumentHighlight[]>((resolve, reject) => {
+			let disposable = this.analyzer.registerForAnalysisOccurrences(n => {
+				if (n.file != file)
+					return;
+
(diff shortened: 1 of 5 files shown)


src/providers/dart_highlighting_provider.ts
```diff
@@ -0,0 +1,69 @@
...
+		return new Promise<DocumentHighlight[]>((resolve, reject) => {
+			let disposable = this.analyzer.registerForAnalysisOccurrences(n => {
+				if (n.file != file)
+					return;
+
+				this.analyzer.analysisSetSubscriptions({ subscriptions: { "OCCURRENCES": [] } });
+				disposable.dispose();
+
+				let highlights: DocumentHighlight[] = [];
+				for (let occurrence of n.occurrences) {
```
DanTup: I'm not familiar with this - is it expected to take just the first occurrence and not all of them?

src/providers/dart_highlighting_provider.ts
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	private buildOccurrences(
+		highlights: DocumentHighlight[], document: TextDocument, position: number, occurrences: as.Occurrences
+	) {
+		let element = occurrences.element;
+		let offsets: number[] = occurrences.offsets;
+		let length: number = occurrences.length;
+
+		for (let i = 0; i < offsets.length; i++) {
```
DanTup: It took me a while to understand what was going on here; it might benefit from some comments!

DanTup:
> This uses the same technique as the document symbol provider - it turns on an analysis server notification briefly, waits for some results, turns off the notification

Do you think it's worth raising feature requests for `getHighlights` in the AS to reduce the amount of data here? Seems like the data for a big document could be massive since it's for every instance of every reference?

(I'm on a Chromebook; haven't been able to test/monitor this, I'm just thinking out loud.. I'll still merge as-is as I think the feature is great!)

src/providers/dart_highlighting_provider.ts
devoncarew: Yup, can add some comments in a follow up.

devoncarew:
The API could definitely be improved here. In two places we're adapting a notification centric API to a more request/response one. For this particular API, we are getting occurrences for every symbol in the file, and then discarding all but the one for the element under the current cursor location. I can clarify that in the code, but a more semantic API would be great :)

DanTup:
Ok, I've raised requests for both (https://github.com/dart-lang/sdk/issues/27062 and https://github.com/dart-lang/sdk/issues/27064). Feel free to add comments if I described anything badly!

PR #66, opened 2016-08-12
Travis testing

devoncarew:
- add support for testing on travis
- run tsc
- run the unit tests
- test on mac and linux

@DanTup, the switch still needs to be flipped over at [travis-ci.org](https://travis-ci.org). Once that's done, we should put the markdown badge in the readme file :)


DanTup:
Sweet! I've been meaning to get the tests running on AV too (not that there are any yet). I'll try and get a release out first, then will work on trying to get these both up and running with tests (and ensure they fail a PR that fails tests). Thanks!

DanTup:
I remembered why I never set Travis up before... Their permissions suck; you can't build a public repo without handing over access to all of your private repos. It's not that I don't trust them; but there are things like API keys for various services in some of my private repos and the less servers/services with access the better!

I think I might have a workaround; just testing it..

.travis.yml
```diff
@@ -0,0 +1,20 @@
...
+before_install:
+  - if [ $TRAVIS_OS_NAME == "linux" ]; then
+      export CXX="g++-4.9" CC="gcc-4.9" DISPLAY=:99.0;
+      sh -e /etc/init.d/xvfb start;
+      sleep 3;
+    fi
+
+install:
+  - npm install
+  - npm run vscode:prepublish
```
DanTup: I think this prepublish and tsc are doing the same thing (I've got AppVeyor working only with the prepublish) so I'm gonna try tweaking that. Think this is all working now across both Travis and AppVeyor. I'll merge some tweaks to master shortly.

Thanks!

devoncarew:
Re: permissions - you may have been looking at the commercial offering? travis-ci.org can only test against public repos - travis-ci.com (I believe) needs the public/private repo read/write permission. I have not agreed to that myself and only use the .org version. It's an unfortunate limitation of the github API - the permissions lack some granularity.

DanTup:
Sorry, maybe I worded badly. To use .org (free OSS) you need to authenticate with GH, and the GH scope they request gives them access to all private repos (even though you can't build them there!).

AppVeyor has a much more limited scope and don't have access to your private repos :)

I've worked around it by promoting this repo to an org (@Dart-Code), which I was thinking of doing anyway and having a new user that's an admin of the org. It all works fine, it's just a bit annoying. I did mention it to GH support this morning asking if they'll fix this (by allowing you to limit repo access at the GH end if the third party is asking for "god permissions") and they said they can't comment on unreleased stuff but they are "well aware" of the issues, so maybe in future it'll be better :)

DanTup:
Ugh, I think you're right.. I think it's been fixed! I re-authorised with Travis and it no longer has access to privates. The previous authorisation did (it's ironic that I didn't like that, but had actually left it on my account).

Maybe I can ditch the other account now! :)

Issue #56, opened 2016-08-11
add the ability to run dart CLI apps


devoncarew:

devoncarew:
@DanTup, heads up as per https://github.com/DanTup/Dart-Code/blob/master/CONTRIBUTING.md,  I may take a look at this -

DanTup:
Awesome :)

I was thinking about this a little (function, not implementation), but I don't know anything about how this stuff works in Code. I guess a Dart project can have many entry points:

- many `.dart` files in `bin` folder
- many `.html` files in `web` folder(?)

Currently it looks like launching stuff goes into `launch.json`(?) but it seems a bit weird to tell users to go and configure lots of stuff there so they can just press `F5`. Is there a better way? Eg. let's say a user clones a program with two HTML files in `web` and two Dart files in `bin` and hits `F5`; what's the experience?

Also, for web stuff (I know this is very separate to running Dart CLI progs), I think it'd be nice to run Dart2JS/DDC(?) in the background when the user hits F5 (they do incremental compile, right?) so the user could make a change in Code, hit Save, then just Refresh in their browser. This seems to be how TS works (first time you hit F5/Build it starts a `Task` which runs and watches for future changes).

Interested to know your thoughts/plan on this; as if you're just doing the CLI stuff maybe I can investigate the web stuff. I guess the web stuff has no debugging (you'd do that in the browser) and is therefore easy?

(Actually, I'm thinking about DDC and using the browser to debug, but I guess Dartium is still a thing - should we be using that or is DDC+rela browser the expected way forwards?)

devoncarew:
Yeah, I was just scoping this to think about dart cli apps :) It's great to think about web aps as well.

I think the current state of the art in vscode is for the user to edit their launch.json file (except for the first launch perhaps, where it's created automatically from the launch info). It would be great to  think about making this easier - perhaps a context menu item to run, or offer to create a launch.json entry?

Re: web, it's probably best to think about a DDC based flow; DDC's really where we see the future of dart for the web. It will leverage existing web workflows as much as possible - it's just generating ES6 modules, and can be consumed in any JS gulp/grunt/browserify/webpack workflows. For debugging, it emits sourcemaps, so you can use the regular devtools debugging. Though I think there is a vscode plugin to let you debug chrome w/ sourcemaps.

devoncarew:
> Interested to know your thoughts/plan on this; as if you're just doing the CLI stuff maybe I can investigate the web stuff.

[...]

PR #77, opened 2016-08-13
Dart run

devoncarew:
- add the ability to run Dart CLI apps (fix https://github.com/Dart-Code/Dart-Code/issues/56)

This is the first part of a debugger, modulo any debugging. Running the first app (F5) creates a new entry in the launch.json file; output goes to the debug console. If the app starts with something other than `bin/main.dart` the user needs to update the starting script in their launch.json file. In order to run more than one app, the user needs to add more entries to their launch.json file - the UI picks up the new entries automatically.


DanTup:
Awesome, this looks great!

I don't understand a lot of this, so I'll have a proper read through and do some testing when at the PC :)

package.json
```diff
@@ -132,13 +133,64 @@
 					"description": "Note: We only send a few very basic events and the extension version number/platform :-)"
 				}
 			}
+		},
+		"debuggers": [{
+			"type": "dart-cli",
+			"label": "Dart command line",
+			"enableBreakpointsFor": {
```
DanTup: If this is just Run for now (without debug, and assuming we might ship this before Debug is done) would it be better to comment this out for now - I think people might get confusing if they can add breakpoints and run in a way that it looks like the debugger is attached, but they don't get hit?

DanTup:
Also - re: the `launch.json`... Can we manipulate this easily? Eg., if you added a new file to the `bin` folder and it wasn't already in `launch.json`, it'd be neat to add a stub (copied from `initialConfigurations`) for that file. Maybe I should look to see if any other extensions do things like this; if Code users generally expect to manage the file manually, we probably don't want to confuse things.

src/debug/debug_impl.ts
```diff
@@ -0,0 +1,197 @@
...
+	private sourceFile: string;
+	private cwd: string;
+	private childProcess: child_process.ChildProcess;
+	private processExited: boolean = false;
+
+	public constructor() {
+		super();
+
+		this.setDebuggerLinesStartAt1(false);
+		this.setDebuggerColumnsStartAt1(false);
```
DanTup: Looks like the default for these is false? ([source](https://github.com/Microsoft/vscode-debugadapter-node/blob/848947f938b7327689362832f1846aed01f92e88/adapter/src/debugSession.ts#L267))

src/debug/debug_impl.ts
```diff
@@ -0,0 +1,197 @@
...
+		// TODO:
+		// if (debug) {
+		// 	appArgs.push("--enable-vm-service:0");
+		// 	appArgs.push("--pause_isolates_on_start=true");
+		// }
+		appArgs.push(this.sourceFile);
+		if (args.args)
+			appArgs = appArgs.concat(args.args);
+
+		// TODO: dart sdk path - send a request to the client for it
```
DanTup: I guess we should resolve this before shipping; currently it's not a requirement for Dart to be in the path to use Dart-Code (you could just set it in the options).

src/debug/debug_impl.ts
```diff
@@ -0,0 +1,197 @@
...
+		let process = child_process.spawn("dart", appArgs, {
+			cwd: args.cwd
+		});
+
+		this.childProcess = process;
+
+		process.stdout.setEncoding("utf8");
```
DanTup: I think the default encoding on child_process is also UTF8, so we could remove these lines?

DanTup:
Just had a quick test of this, and it works great! :) First press of F5 asked me what to use (I picked "Dart command line") then told me I need to configure `launch.json` and created it, pre-populated with your stuff from the `package.json` and after I tweaked the filename, etc. it worked great!

I made a couple of minor comments on the code, though I also have one other comment/question... In Visual Studio if I ran/debugged a console app, I would see the console app open up. Here it seems the process is "hidden" and just routed into the Debug window. I don't think this is necessarily wrong, but it wasn't what I expected. I don't know what other IDEs do (or Code does for command line apps of other languages) so thought it worth mentioning.

I think with the SDK Path fixed, this would be good to ship as a feature and debugging could be added later.

I'm gonna start looking at DDC stuff, though I think I have too many questions (I'll probably post on dart-misc shortly) to start that yet (We might also need to tweak the default `launch.json` to support web? it's a shame we can't provide a default config programmatically based on presence of `bin`/`web` folders!).

DanTup:
Related - if we jeep current behaviour, we should show the Debug window at launch (if you open Problems view, then that just stays there).

That said, I'm now thinking about STDIN - if someone wants to read input, doing it in here will complicate things?

package.json
devoncarew: Yup, makes sense - removed.

src/debug/debug_impl.ts
devoncarew: Looks to be the case - I'll remove.

src/debug/debug_impl.ts
devoncarew: I don't think there's a way for the debug adapter to query the host vscode. The dart sdk location would have to be set in every launch config, which would be pretty annoying for users. We really want a way to pass a group of settings through automatically at launch (`dart.*`). The language servers have a feature like this -

src/debug/debug_impl.ts
devoncarew: done

devoncarew: oops, removing these means we don't get stdout output from the process

devoncarew:
re: manipulating the launch.json file - it's something we could do, but might not be exactly what user's expect. Perhaps on an explicit user action (right-click, run)? I'm hoping we can lean on the default vscode behavior here, and anticipate it improving a bit over time.

I do see the debug console open for me. There have been a few times in the past - when running a vscode extension - that the window hasn't opened, or has been obscured by another open view in that area (perhaps I'd closed the debug console recently?). I think the debug console is the best place for the output, and we can hope that any issues w/ the view not showing up are bugs in vscode. Using the debug console for output will mean that the debugger will be able to support a repl in that view (once we have more of the debugger hooked up).

devoncarew:
Thanks for taking a look! Updated -

src/debug/debug_impl.ts
DanTup: Oh, I thought the debug adapter was just running in Code and could import our settings etc. (I haven't read anything about it at all yet!).

I don't think it'd be terrible if we required the SDK in PATH, but if so, I think we should:

1. Catch when the user tries to run and tell them it's not in path and debugging will fail (I don't know if we can catch the attempt to run?)
2. Inform the user at activation that Dart is not in path and launching won't work (or even fail to activate)

What do you think? (we can break this into an issue and deal with later/before next release to avoid dragging this PR out if you want).

src/debug/debug_impl.ts
DanTup: Removing the encoding? I meant the two `setEncoding` lines, but not the events. But, there's another comment further down about launching the app properly (in which case I'm not sure having stdout missing from the debug window would be incorrect?)

DanTup:
> re: manipulating the launch.json file - it's something we could do, but might not be exactly what user's expect.

Yeah, I think you're right. And at least we can get feedback (or use it ourselves) a little and see how it works out.

>Perhaps on an explicit user action (right-click, run)?

That could be good too.

I've created #81 to look at options for this.

DanTup:
> I do see the debug console open for me. There have been a few times in the past - when running a vscode extension - that the window hasn't opened, or has been obscured by another open view in that area (perhaps I'd closed the debug console recently?).

I did see it first time (I had no other views open), but not when I had the Problems window open. Might've just been a one-off, I'll re-test and can raise a separate issue if I can repro.

> I think the debug console is the best place for the output, and we can hope that any issues w/ the view not showing up are bugs in vscode. Using the debug console for output will mean that the debugger will be able to support a repl in that view (once we have more of the debugger hooked up).

I think a repl would be good, but it still feels weird for the process to be inside the debug window. Eg. if the app is reading input from `stdin`, do we have to proxy it all through the debug window? If so, how does that interact with the repl? (I'm also concerned about the performance of large amounts of text into Code debug windows; the debug window I was logging the AS output into performed terribly and didn't allow copy/pasting properly!) :cry:

My experience with IDEs is limited to mostly VS, so I'm open to ideas, but what I would've expected is that a new console spawns running the program (as if I'd done `WinKey ->Run -> dart.exe blah.dart`) and I can interact with it there as normal (and even terminate it!). Debug window could function as a repl while execution is paused and for debug output (though I don't know if Dart has any equiv of this - in C# we have Debug.WriteLine() that goes to an output window, whereas stdout exists only in the real window executing the process).

src/debug/debug_impl.ts
devoncarew: We can probably fail w/ an error message that `dart` was not found, and they should adjust their path. Eventually, it would be great to be able to get the value from the user's settings. That way running and debugging will work as long as they have configured vscode w/ an sdk.

DanTup: It's really dirty, but... we could stash the sdk path in a temp file in %temp% and read it from the debug adapter. It stinks, but possibly the improve experience offsets it?

Otherwise, is there a hook to detect when the user tries to start a debug session to display the message?

src/debug/debug_impl.ts
devoncarew: From reading the APIs, setting the encoding changes the stream to return strings instead of `Buffer`s; I think we want to set the encoding here, but happy to change if we find a different way.

devoncarew:
OK, I _think_ the expectation is for stdout output and such to show up in the debug console. It'd be good to explore how other plugins do it as well for comparison. Perhaps we can land this as it is currently, and survey other plugins / vscode behaviors?

src/debug/debug_impl.ts
DanTup: Ah yes, I remember. You can call `toString()` on the buffer to get as string (which defaults to reading it as `utf8`) but I don't know if we'd be saving anything (except removing the utf8 string), so whatever is fine.

DanTup:
Ok, I've raised #82 and #83 for the two points I think are outstanding. I'll try and get some time to look at them tomorrow; would be nice to resolve them both as we ship this in 0.8.

Thanks again; feels like things are starting to come together! Already feels like a much better integration than I got to with DartVS! :)

src/debug/debug_impl.ts
devoncarew: I don't think there's a hook before the user runs; it may be something that vscode could add.

> we could stash the sdk path in a temp file in %temp% and read it from the debug adapter

That would work, and sounds like a good solution until there's a more official way to access user prefs.

DanTup: Ok; I'll investigate this and also raise a case about support in Code and see what MS say.

Issue #78, opened 2016-08-13
expose organize directives


devoncarew:
There's an analysis server command for organizing directives; I'd love to have it available - I use it all the time in other tools.

DanTup:
I looked for this and didn't find it; I guess I was looking for "Sort" or similar!

In C#, I normally have it set to do both "Remove and Sort Usings" and "Format Document". Does it seem reasonable to have some options for similar things:

- `dart.formatOnSave` - Formats the document automatically upon save
- `dart.organiseDirectivesOnFormat` - Automatically sorts (and removes unused) directives when formatting the document

I think we could default the second one to try and it wouldn't upset anyone; though auto-formatting should probably be off (I've seen what it did to my code! if you're not expecting it and don't have source control (some don't, I know!) you might get mad if we trash your perfectly formatted file).

devoncarew:
- format on save makes sense, 👍 
- organize on format also would make sense; I have run into some situations where organize directives doesn't do a great job w/ the code. Perhaps is some items are commented out? In any case, for myself, it'd probably remain an explicit user action.

DanTup:
> I have run into some situations where organize directives doesn't do a great job w/ the code. Perhaps is some items are commented out?

I wonder if these are things that could be handled better in the analyzer?

PR #79, opened 2016-08-13
implement organize directives

devoncarew:
- implement organize directives
- fix https://github.com/Dart-Code/Dart-Code/issues/78

screen shot 2016-08-12 at 7 29 19 pm

### DIFF
--- a/src/editors.ts
+++ b/src/editors.ts
@@ -1,0 +1,10 @@
+"use strict";
+
+import * as vs from 'vscode';
+
+export function hasActiveDartEditor(): boolean {
+  if (!vs.window.activeTextEditor)
+  	return false;
+
+  return vs.window.activeTextEditor.document.fileName.endsWith(".dart");
+}
--- a/src/editors.ts
+++ b/src/editors.ts
@@ -6,5 +6,5 @@ export function hasActiveDartEditor(): boolean {
   if (!vs.window.activeTextEditor)
   	return false;
 
-  return vs.window.activeTextEditor.document.fileName.endsWith(".dart");
+  return vs.window.activeTextEditor.document.languageId === "dart";
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -76,7 +76,7 @@ export function activate(context: vs.ExtensionContext) {
 	context.subscriptions.push(new AnalyzerStatusReporter(analyzer));
 
 	// Set up commands for Dart editors.
-	new DartCommands(context, analyzer);
+	context.subscriptions.push(new DartCommands(context, analyzer));
 
 	// Set up diagnostics.
 	let diagnostics = vs.languages.createDiagnosticCollection("dart");
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -7,6 +7,7 @@ import { analytics } from "./analytics";
 import { Analyzer } from "./analysis/analyzer";
 import { AnalyzerStatusReporter } from "./analyzer_status_reporter";
 import { config } from "./config";
+import { DartCommands } from "./providers/dart_commands";
 import { DartCompletionItemProvider } from "./providers/dart_completion_item_provider";
 import { DartDefinitionProvider } from "./providers/dart_definition_provider";
 import { DartReferenceProvider } from "./providers/dart_reference_provider";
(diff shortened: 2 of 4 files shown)


src/editors.ts
```diff
@@ -0,0 +1,10 @@
+"use strict";
+
+import * as vs from 'vscode';
+
+export function hasActiveDartEditor(): boolean {
+  if (!vs.window.activeTextEditor)
+  	return false;
+
+  return vs.window.activeTextEditor.document.fileName.endsWith(".dart");
```
DanTup: In other places we use the `languageId` on the `editor`(`document`?). This means if the user changes the language for a document (done in the status bar) it'll honour their language (though whether it's likely someone will change a `.dart` file to `HTML` and then expect the Dart things to vanish and HTML ones to appear, I don't know!)

src/extension.ts
```diff
@@ -74,6 +75,9 @@ export function activate(context: vs.ExtensionContext) {
 	context.subscriptions.push(vs.languages.registerDocumentHighlightProvider(DART_MODE, new DartDocumentHighlightProvider(analyzer)));
 	context.subscriptions.push(new AnalyzerStatusReporter(analyzer));
 
+	// Set up commands for Dart editors.
+	new DartCommands(context, analyzer);
```
DanTup: It feels really weird calling a constructor and not assigning the result, and I think I've tried to make sure only `extension.ts` is doing things like adding to `context.subscriptions`.

I think it'd be better to make `DartCommands` disposable and put it into `context.subscriptions` here and have its dispose handle its children, or make this a function (`registerCommands`?) that sets them up and returns disposables that are put into the collection here.

(Happy to discuss if you think the idea is weird though!)

src/providers/dart_commands.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as editors from "../editors";
+import * as vs from "vscode";
+import { Analyzer } from "../analysis/analyzer";
+
+export class DartCommands  {
+	private analyzer: Analyzer;
+	constructor(context: vs.ExtensionContext, analyzer: Analyzer) {
+		this.analyzer = analyzer;
+
+        context.subscriptions.push(
```
DanTup: Not sure if it's GitHub (I'm not at PC to check), but the indenting in this file looks a bit wonky.

src/providers/dart_commands.ts
```diff
@@ -0,0 +1,47 @@
...
+    organizeDirectives(editor: vs.TextEditor, editBuilder: vs.TextEditorEdit) {
+        if (!editors.hasActiveDartEditor()) {
+            vs.window.showWarningMessage("No active Dart editor.");
+            return;
+        }
+
+        this.analyzer.editOrganizeDirectives({ file: editor.document.fileName }).then((response) => {
+            let edit: as.SourceFileEdit = response.edit;
+			if (edit.edits.length == 0) {
+                vs.window.showInformationMessage("No changes from organize directives.");
```
DanTup: I think maybe it'd be best to just be silent if there are no changes to make. We wouldn't alert the user if a Format Document operation didn't make any changes, so I don't think this is much different (IDEs I've used don't alert here, but I haven't used many!).

Thoughts?

DanTup:
This sounds great! I don't think TS has this, I've been using `Ctrl+Shift+2` to sort a selection for me TS stuff which is tedious =D

I've added a few comments, don't think anything major. Thanks!

src/editors.ts
devoncarew: Updated to use the `languageId` field

src/extension.ts
devoncarew: No, this does look odd to me; I'll refactor.

src/providers/dart_commands.ts
devoncarew: removed some spaces; I reformatted the file as well, but am not sure whether that converts spaces ==> tabs

src/providers/dart_commands.ts
devoncarew: It's nice to let the user know that the operation did run, so they don't think it didn't take or the key binding wasn't correct. The toast vscode uses is persistent though, so the user would have to dismiss it each time. I'll remove the notification here and we can see how it feels.

devoncarew:
Updated!

src/providers/dart_commands.ts
DanTup: I didn't think it does, but they do look like tabs (in GH at least) now. You can always force it be clicking in the status bar where it tells you what the indenting is (which then puts the options up in the command palette which is a bit weird!).

src/providers/dart_commands.ts
DanTup: Yeah, I was kinda torn, but since we don't do it for things like format, I think just saying out the way is best.

That said, I'm not against creating a little function that shows an info message for x seconds and using that, and then applying the same to reformat (and anywhere else we want to give feedback without a persistent message). What do you think?

DanTup:
All looks great! I'll merge as-is; if you think we should add an x-second notification for the failures let me know.

PR #85, opened 2016-08-13
display analysis server crashes

devoncarew:
- display analysis server crashes

In rare cases, the analysis server can throw fatal errors (stack traces from its own code). We dump these to the devtools console; if the user has the `analyzerDiagnosticsPort` setting on, we also display the error in a dialog.


DanTup:
Great; nice idea showing an error message in the main Code instance too! :)

Though I think this adds more weight to the idea of ditching `verbose` since both that and the port are being used for similar things. Unless you think there's a good reason to keep `verbose`, I'll rip it out.

Issue #37, opened 2016-08-06
Add ability to log analysis service communication to a file


DanTup:
There are some related comments in #36.

PR #86, opened 2016-08-14
Logging

devoncarew:
- only log output w/ the dart.verbose setting
- log analysis server traffic to a file (fix https://github.com/Dart-Code/Dart-Code/issues/37)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 out
 node_modules
 .vscode-test/
+analyzer.log
--- a/src/analysis/analyzer.ts
+++ b/src/analysis/analyzer.ts
@@ -5,6 +5,7 @@ import * as child_process from "child_process";
 import * as as from "./analysis_server_types";
 import { AnalyzerGen } from "./analyzer_gen";
 import { config } from "../config";
+import { log } from "../utils";
 
 export class Analyzer extends AnalyzerGen {
 	private analyzerProcess: child_process.ChildProcess;
@@ -15,7 +16,7 @@ export class Analyzer extends AnalyzerGen {
 	constructor(dartVMPath: string, analyzerPath: string) {
 		super();
 
-		console.log(`Starting Dart analysis server...`);
+		log(`Starting Dart analysis server...`);
 
 		let args = [analyzerPath];
 
@@ -117,7 +118,7 @@ export class Analyzer extends AnalyzerGen {
 	}
 
 	stop() {
-		console.log(`Stopping Dart analysis server...`);
+		log(`Stopping Dart analysis server...`);
 
 		this.analyzerProcess.kill();
 	}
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 out
 node_modules
 .vscode-test/
+analyzer.log
```
DanTup: This should go in the vscode ignore file too, so `vsce package` and `vsce publish` don't include it if it's on the disk.

src/analysis/analyzer.ts
```diff
@@ -69,11 +71,26 @@ export class Analyzer extends AnalyzerGen {
...
+		if (config.analyzerLogFile) {
+			if (!this.logStream)
+				this.logStream = fs.createWriteStream(config.analyzerLogFile);
+			if (message.length > max)
+				this.logStream.write(message.substring(0, max) + "...\n");
+			else
+				this.logStream.write(message);
+		} else if (!config.analyzerLogFile && this.logStream) {
+			// Turn off logging.
+			this.logStream.close();
```
DanTup: I don't know if it's important, but this `logStream` will only be closed if the setting is turned off and another request sent. Could this result in a locked file if you reload window / change workspace folder? Would it be better closing in the `close` method if open too?

(though really I wonder if this class should've been disposable and adding to `vscode.subscriptions` instead of a close method!)

src/utils.ts
```diff
@@ -93,3 +93,8 @@ export function isWithinRootPath(file: string) {
 
 	return workspace.rootPath != null && file.startsWith(workspace.rootPath + path.sep); 
 }
+
+export function log(message: any): void {
+	if (config.verbose)
```
DanTup: I wonder if there's any value in `config.versbose` anymore. The console (presumably) only logs with the debugger attached so I don't think there's much value in turning it off? What's left is very low traffic (and we do some logging without checking this, like warnings in error responses, so maybe it's also applied a bit inconsistently).

src/analysis/analyzer.ts
devoncarew: I'll close this in the `close` method.

Changing this class to Disposable makes sense to me -

.gitignore
devoncarew: will do

src/utils.ts
devoncarew: I have no problem dropping `verbose` :) I'll keep this PR as is and we can remove in future work.

devoncarew:
(updated)

src/analysis/analyzer.ts
DanTup: Added #90 for making `Disposable`.

src/utils.ts
DanTup: Added #89 to remove `verbose`; also suggested we tweak the logging (that sends warnings to warning dialogs) to use the same "developer" detection as analytics.

DanTup:
Great; thanks! 👍

Issue #83, opened 2016-08-13
Improve passing of Dart SDK path to debugger (command variables)


DanTup:
~~The rest of the plugin works without it, but launching does not. We should review whether we can give the user a message when they hit F5 or not (if not, we might need to notify them at activation that launching won't work).~~

~~Alternatively, see if we can find a way to get the SDK path to the debug adapter.~~

This was fixed, but this issue is now about moving over to command variables so it plays nicer with multiple instances.

DanTup:
I've opened https://github.com/Microsoft/vscode/issues/10503 to see if there's a better way (or Code might add one), but for now maybe try the temp file.

DanTup:
This can be done much better and support multiple windows, see:

see section 'Debug Extension Authoring: Command Variables' in the [June release notes](https://code.visualstudio.com/updates/June_2016) and the [mock debugger](https://github.com/Microsoft/vscode-mock-debug/blob/master/src/extension.ts).

Issue #18, opened 2016-08-05
Detect when sdkPath is changed to avoid reloading


DanTup:
There's an `onDidChangeConfiguration` event.

DanTup:
Need to ensure this also updates the file used to pass the path to the debugger (if it still exists).

DanTup:
Automatic reloading is tricky (because we'd have to re-send all open file buffers, etc.) so instead I've done #219 which prompts the user to save and reload if this value changes.

PR #88, opened 2016-08-14
Stash dart SDK path in a temp file so debugger can read.

DanTup:
@devoncarew How does this look? We stash the SDK during activation and then read it back in the debug adapter. I created a file to hold the read/write functions (although only one is needed on each side it made sense to keep the path in one place) and added a comment about it being used in both places (it's important not to pull in `vscode` etc. or the debugger fails to launch with no useful error!).

If you think is good, we can merge as the solution to #83  in lieu of something better.

### DIFF
--- a/src/debug/sdk_path.ts
+++ b/src/debug/sdk_path.ts
@@ -1,0 +1,30 @@
+"use strict"
+
+// This file exists to write the SDK path from config in the extension and read it from the
+// debug adapter. Everything imported here must be available in both places (eg. no vscode).
+import * as os from "os";
+import * as path from "path";
+import * as fs from "fs";
+
+const sdkPathStoreName = "dart-code.sdk";
+const sdkPathStore = path.join(os.tmpdir(), sdkPathStoreName);
+
+export function readSdkPath(): string {
+	// If this file doesn't exist, return null. We'll have to hope it's
+	// in PATH.
+	try {
+		return fs.readFileSync(sdkPathStore, "utf8").trim();
+	}
+	catch (e) {
+		return null;
+	}
+}
+
+export function writeSdkPath(sdkPath: string) {
+	try {
+		fs.writeFileSync(sdkPathStore, sdkPath, "utf8");
+	}
+	catch (e) {
+		console.warn(e);
+	}
+}
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -22,6 +22,7 @@ import { FileChangeHandler } from "./file_change_handler";
 import { OpenFileTracker } from "./open_file_tracker";
 import { PubManager } from "./commands/pub";
 import { ServerStatusNotification } from "./analysis/analysis_server_types";
+import * as debug from "./debug/sdk_path"
 
 const DART_MODE: vs.DocumentFilter = { language: "dart", scheme: "file" };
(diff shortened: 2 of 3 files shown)


DanTup:
I just got this when I added the C# extension in Code:

Path

So, I'm totally happy to just do the same and ditch support fot it not being on the path altogether. I thought maybe switching to newer pre-release SDKs etc. the current behaviour might be useful, but if MS are happy to refuse to work without it it's good enough for me.

Let me know if you have a preference and I'll either merge this or abandon and change our startup to do the same.

src/debug/sdk_path.ts
```diff
@@ -0,0 +1,30 @@
+"use strict"
+
+// This file exists to write the SDK path from config in the extension and read it from the
+// debug adapter. Everything imported here must be available in both places (eg. no vscode).
+import * as os from "os";
+import * as path from "path";
+import * as fs from "fs";
+
+const sdkPathStoreName = "dart-code.sdk";
+const sdkPathStore = path.join(os.tmpdir(), sdkPathStoreName);
```
devoncarew: `tempdir()`, or somewhere in user home (`~/.dartcode/.settings`)? I have no preference but wanted to call out the option.

src/extension.ts
```diff
@@ -41,6 +42,7 @@ export function activate(context: vs.ExtensionContext) {
 		return; // Don't set anything else up; we can't work like this!
 	}
 	context.globalState.update(stateLastKnownSdkPathName, dartSdkRoot);
+	debug.writeSdkPath(dartSdkRoot); // Write the SDK path for the debugger to find.
```
devoncarew: We do this on startup, but we should also write the value when the preference changes.

We'll run into some subtle issues if the user has multiple copies of vscode open, with different values for the dart sdk path in each. `run` will pull the sdk value from the various copies of vscode non-deterministically. This may not be a common issue; perhaps future changes to the debugging api would let us address this.

devoncarew:
lgtm! I think this improves the situation for users.

One thought is, we could also use the project's `.vscode/` directory to store our own settings file, a copy of the dart sdk path. I'm not sure if a debug adapter knows where the project root is; if it does it could then read from that file.

User's may not appreciate that as it would be a dirty file from git's perspective, and could have a path local to their machine. Using the global tempdir may be the best option here -

src/debug/sdk_path.ts
DanTup: Since it's just throwaway info, I think temp is better.. I hate my home folder being clogged up with crap from apps; it's *my* home folder! IMO if I don't know what it's for, it shouldn't be in my home folder!. I don't feel too strongly if you think there are reasons to put elsewhere though (but having what looks like another settings file separate to the VS one might seem weird to the user).

src/extension.ts
DanTup: Yeah, this needs doing for more than just this though - I created #18 to handle this. We'll need to terminate and restart the analyzer, etc.

I also added a note in that case about this, so I think we can just deal with it there (I think it should be fixed by v1).

DanTup:
Yeah, I think we'll stick with global temp for now. If it was something we expected users to change (or even care about) it might make sense where they can see it, but for all intents and purposes this is a temporary hack to pass data around; not something the user should really be concerned with.

Issue #62, opened 2016-08-11
Improve information shown in workspace search results


devoncarew:
I think this is a result of merging two API call results?

<img width="423" alt="screen shot 2016-08-11 at 3 43 45 pm" src="https://cloud.githubusercontent.com/assets/1269969/17607440/9741cf9e-5fda-11e6-90e5-96f4a343a4eb.png">

DanTup:
Strange; I was sure I tested this and it didn't happen. I just tried now searching for a top-level class but still don't see it. Could you confirm what your Dart file looks like?

![Search results](https://cloud.githubusercontent.com/assets/1078012/17621582/2314ebb2-608c-11e6-9016-cfb0b9c7d560.png)

DanTup:
I wonder if something weird is happening with the results, maybe #63 is related. Are you able to capture the service output when this happens? (this might be tricky until there's an option to log to file)

DanTup:
@devoncarew I'm gonna do a little more testing and then do a build for 0.7. I've pushed some minor things back to 0.8 as I'd like to release the other stuff; this the only remaining case. If you can repro this or think it's potentially an issue, if you let me know before I build, I'll revert that change until 0.8.

DanTup:
I went ahead and pushed 0.7. If we can repro this I can always push an update :)

devoncarew:
Cool; I can re-test and try and narrow down the issue this evening.

devoncarew:
Hmm; I had two `foo` methods, and two `FooBar` classes in that sample project, in separate libraries :( User error -

devoncarew:
If we change the containerName in the workspace search results to be the library path, we get this:

<img width="615" alt="screen shot 2016-08-12 at 7 37 33 pm" src="https://cloud.githubusercontent.com/assets/1269969/17641031/53249c7a-60c4-11e6-940f-fa53cfe4a225.png">

which could help users ID where the symbols are coming from.

DanTup:
Seems like a good change to me! 👍

DanTup:
That said; are we then losing the container name; eg:

````
/// test.dart

class Foo {
    string name;
}

class Bar {
    string name;
}
```

Would these both appear as `name (test.dart)`? It's less of an issue in the document one because of the highlighting, but maybe worth doing something here. We could show `Bar.name` but I guess with nested classes etc., the issue might continue.

I think the service returns a full path of parents; maybe `containerName` could be all parents up the tree?

[...]

Issue #74, opened 2016-08-12
Workspace symbol-search should not show private items outside of your project


DanTup:
I think we should only include public-API for things in packages.

DanTup:
See comments in #91.

DanTup:
@devoncarew I opened that project but I couldn't actually repro.. only one result for "dir" and no privates for "di".

![Search](https://cloud.githubusercontent.com/assets/1078012/17673347/8eb979b2-6318-11e6-9ea7-0888049d03fa.png)

I've be interested to know if you can repro this (ideally on current master).

devoncarew:
Looks fixed to me:

<img width="620" alt="screen shot 2016-08-15 at 5 52 53 pm" src="https://cloud.githubusercontent.com/assets/1269969/17684519/4cd309ac-6311-11e6-8fa9-9e9fdc96c29b.png">

<img width="617" alt="screen shot 2016-08-15 at 5 53 01 pm" src="https://cloud.githubusercontent.com/assets/1269969/17684520/4cdf0c66-6311-11e6-9068-5f9981bbd172.png">

DanTup:
I pushed another fix last night; I was checking parent names for underscores so `_Thing.thing` would still appear which I figured wasn't expected (though now I'm wondering why you're start a class with `_` vs putting inside the `src` folder?)

devoncarew:
I `_` class would be library private. `src/` is more of an informal convention - it's not enforced. The access for underscores is enforced by the runtime.

DanTup:
Gotcha!

PR #91, opened 2016-08-15
tweak the workspace symbols display

devoncarew:
- work towards https://github.com/Dart-Code/Dart-Code/issues/62
- tweak the workspace symbols display

screen shot 2016-08-14 at 7 16 37 pm

screen shot 2016-08-14 at 7 17 17 pm

The results are for running in the VSCode Insiders (nightly) build. This uses `package:foo/bar.dart` for package libraries, uses relative paths for other libraries, and displays the container hierarchy in the name for elements (`Foo.bar` for class members; `foo` for top-level symbols).

### DIFF
--- a/src/providers/dart_workspace_symbol_provider.ts
+++ b/src/providers/dart_workspace_symbol_provider.ts
@@ -18,7 +18,7 @@ export class DartWorkspaceSymbolProvider implements WorkspaceSymbolProvider {
 		return new Promise<SymbolInformation[]>((resolve, reject) => {
 			Promise.all([
 				this.searchTopLevelSymbols(query),
-				this.searchmemberDeclerations(query)
+				this.searchMemberDeclarations(query)
 			]).then(results => resolve(this.combineResults(results)), e => { console.warn(e.message); reject(); });
 		});
 	}
@@ -44,7 +44,7 @@ export class DartWorkspaceSymbolProvider implements WorkspaceSymbolProvider {
 		});
 	}
 
-	private searchmemberDeclerations(query: string): PromiseLike<as.SearchResult[]> {
+	private searchMemberDeclarations(query: string): PromiseLike<as.SearchResult[]> {
 		return new Promise<as.SearchResult[]>((resolve, reject) => {
 			// TODO: Change this if the regex "support" gets fixed.
 			let pattern = this.makeCaseInsensitiveFuzzyRegex(query);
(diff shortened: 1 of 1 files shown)


devoncarew:
Strangely enough, we don't get top-level symbol results for things in `dart:` libraries (`List`, `String`, `Future`, ...).

DanTup:
This looks great!

Do you think we need to make the document symbol list (`Ctrl+Shift+O`) look more like this?

How did you get Code to display those dots without stripping stuff?

> Strangely enough, we don't get top-level symbol results for things in `dart:` libraries (`List`, `String`, `Future`, ...).

Maybe we should raise this as a request? (possible [we might end up with a new search API](https://groups.google.com/a/dartlang.org/d/msg/analyzer-discuss/conkKX7XiKI/Qmw7o3baCAAJ)?)

From your screenshot I see my `shouldIncludeResult` isn't working correctly - you have some package stuff starting with underscores - as I understood, they should be filtered out for not being in the project root. Do you know what happened there? (did you possibly have them open from inside the workspace, which maybe made them come back with workspace paths in the search). If so, should I tweak my `shouldIncludeResult` to treat `packages` the same way as things outside the root?

I'll give this a little test and merge tonight. Thanks!

devoncarew:
> Do you think we need to make the document symbol list (Ctrl+Shift+O) look more like this?

Possibly just to convert it to `Class.member` style? We really wouldn't use the `containerName` fields much at all then -

> How did you get Code to display those dots without stripping stuff?

I think this boils down to using the Insiders build. I stil saw the issue w/ the regular stable build. I went back and forth over whether the issue was fixed in Insiders - I'd be interested i what you see locally -

> From your screenshot I see my shouldIncludeResult isn't working correctly - you have some package stuff starting with underscores

I'm not sure why - from looking at the code it looks like the filter should have worked here. The only thing I did differently was to open the parent of the dart project - it's a project that contains both a dart and a java sub-project (https://github.com/dart-lang/vm_service_drivers).

DanTup:
> Possibly just to convert it to Class.member style? We really wouldn't use the containerName fields much at all then

Yeah, probably. It did look good as-is, but I think being consistent is probably better. What do you think?

> I went back and forth over whether the issue was fixed in Insiders - I'd be interested i what you see locally 

Based on comments from MS, it should've been fixed two weeks ago. I'll try Insiders. Maybe worth you re-checking to see if maybe you'd just confused the two (are they visibly different?). If you can still repro on insiders (and it's up-to-date), might be worth commenting back on the MS case (since it's closed now).

> I'm not sure why - from looking at the code it looks like the filter should have worked here.

I've reopened #74 and will take a look, ta.

Issue #31, opened 2016-08-06
Add a RenameProvider to support rename refactor


DanTup:
[VSCode doc](https://code.visualstudio.com/docs/extensionAPI/vscode-api#RenameProvider)
[Analysis Server doc - getRefactorings](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_edit.getAvailableRefactorings)
[Analysis Server doc - getRefactoring](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#request_edit.getRefactoring)

ikhwanhayat:
Just a heads up, I'm currently doing this.

DanTup:
Looks like @devoncarew may have started on this yesterday; not sure of the status though .. ?

https://github.com/Dart-Code/Dart-Code/commit/32aeba46ddd0219f3ffcf20c09fa3d9ce88d4cbc

devoncarew:
Hey - I took a stab at it, but didn't ping this issue. Happy to hold off on my implementation. @ikhwanhayat, feel free to mine bits of it out that are helpful.

DanTup:
@ikhwanhayat:
> In VS Code for TypeScript, it just opens all the affected file. For undo, you need to undo in each of the files individually. My implementation just follows that.

This sounds reasonable to me then. Maybe in future Code will support prompting etc. for us.

> Not quite sure on how to handle error messages though.

What you've done all looks good to me; what's your concern with this? If it's displaying all the messages, maybe just show the first (that's what Devon's did). Depending on the messages, we might need to display them differently though.

If you open a PR when you think it's basically working, I'll have a test and see how it feels.

Thanks!

ikhwanhayat:
> What you've done all looks good to me; what's your concern with this? If it's displaying all the messages, maybe just show the first (that's what Devon's did). Depending on the messages, we might need to display them differently though.

I am thinking whether these kind of messages are better shown using the popup alert (`vscode.window.showErrorMessage(...)`) or is it better if can create a "Refactor" output channel that we can standardize to be used for all kind of refactorings.

DanTup:
Not sure; maybe worth seeing what others do. I haven't actually tried to do this in TypeScript yet despite all the coding in it recently!

devoncarew:
[...]

PR #92, opened 2016-08-15
Implement RenameProvider

ikhwanhayat:
For issue #31

Feel free to comment on code structure/performance/etc.

### DIFF
--- a/src/providers/dart_rename_provider.ts
+++ b/src/providers/dart_rename_provider.ts
@@ -4,19 +4,19 @@ import { window, workspace, RenameProvider, WorkspaceEdit, TextDocument, Positio
 import { Analyzer } from "../analysis/analyzer";
 import * as as from "../analysis/analysis_server_types";
 
-export class DartRenameProvider implements RenameProvider {	
+export class DartRenameProvider implements RenameProvider {
 	private analyzer: Analyzer;
 	constructor(analyzer: Analyzer) {
 		this.analyzer = analyzer;
 	}
-	
-	provideRenameEdits(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {		
+
+	provideRenameEdits(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
 		return this.doRename(document, position, newName, token);
 	}
 
 	private doRename(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
 		return new Promise<WorkspaceEdit>((resolve, reject) => {
-			var wordRange = document.getWordRangeAtPosition(position);				
+			var wordRange = document.getWordRangeAtPosition(position);
 			this.analyzer.editGetRefactoring({
 				kind: "RENAME",
 				file: document.fileName,
@@ -25,8 +25,8 @@ export class DartRenameProvider implements RenameProvider {
 				validateOnly: false,
 				options: {
 					newName: newName
-				}			
-			}).then(resp => {				
+				}
+			}).then(resp => {
 				let workspaceEdit = new WorkspaceEdit();
 
 				let hasError = this.handleProblem(
@@ -36,34 +36,34 @@ export class DartRenameProvider implements RenameProvider {
 				);
 
 				if (hasError) {
-					reject("Rename cancelled.");
+					reject("");
 					return;
 				}
 
-				let promises = [];				
+				let promises = [];
 				resp.change.edits.forEach(changeEdit => {
 					changeEdit.edits.forEach(fileEdit => {
 						let uri = Uri.file(changeEdit.file);
 						let promise = workspace.openTextDocument(uri)
 						promises.push(promise);
-						promise.then(document => {											
+						promise.then(document => {
 							workspaceEdit.replace(
 								uri,
 								new Range(
-									document.positionAt(fileEdit.offset), 
+									document.positionAt(fileEdit.offset),
 									document.positionAt(fileEdit.offset + fileEdit.length)),
 								fileEdit.replacement
 							);
 						});
-					});					
+					});
 				});
 
 				// Wait all openTextDocument to finish
 				Promise.all(promises).then(() => {
-					window.showInformationMessage(resp.change.message); 
+					window.showInformationMessage(resp.change.message);
 					resolve(workspaceEdit)
 				});
-				
+
 			}, e => console.warn(e.message));
 		});
 	}
--- a/src/providers/dart_rename_provider.ts
+++ b/src/providers/dart_rename_provider.ts
@@ -15,12 +15,13 @@ export class DartRenameProvider implements RenameProvider {
 	}
 
 	private doRename(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
-        return new Promise<WorkspaceEdit>((resolve, reject) => {
+		return new Promise<WorkspaceEdit>((resolve, reject) => {
+			var wordRange = document.getWordRangeAtPosition(position);				
 			this.analyzer.editGetRefactoring({
 				kind: "RENAME",
 				file: document.fileName,
 				offset: document.offsetAt(position),
-				length: position.character,
+				length: wordRange.end.character - wordRange.start.character,
 				validateOnly: false,
 				options: {
 					newName: newName
--- a/src/providers/dart_rename_provider.ts
+++ b/src/providers/dart_rename_provider.ts
@@ -58,7 +58,10 @@ export class DartRenameProvider implements RenameProvider {
 				});
 
 				// Wait all openTextDocument to finish
-				Promise.all(promises).then(() => resolve(workspaceEdit));
+				Promise.all(promises).then(() => {
+					window.showInformationMessage(resp.change.message); 
+					resolve(workspaceEdit)
+				});
 				
 			}, e => console.warn(e.message));
 		});
--- a/src/providers/dart_rename_provider.ts
+++ b/src/providers/dart_rename_provider.ts
@@ -20,7 +20,7 @@ export class DartRenameProvider implements RenameProvider {
 			this.analyzer.editGetRefactoring({
 				kind: "RENAME",
 				file: document.fileName,
-				offset: document.offsetAt(position),
+				offset: document.offsetAt(wordRange.start),
 				length: wordRange.end.character - wordRange.start.character,
 				validateOnly: false,
 				options: {
--- a/src/providers/dart_rename_provider.ts
+++ b/src/providers/dart_rename_provider.ts
@@ -1,8 +1,9 @@
 "use strict";
 
-import { window, workspace, RenameProvider, WorkspaceEdit, TextDocument, Position, CancellationToken, Uri, TextEdit, Range } from "vscode";
+import { window, workspace, RenameProvider, OutputChannel, WorkspaceEdit, TextDocument, Position, CancellationToken, Uri, TextEdit, Range } from "vscode";
 import { Analyzer } from "../analysis/analyzer";
 import * as as from "../analysis/analysis_server_types";
+import * as channel from "../commands/channels"
 
 export class DartRenameProvider implements RenameProvider {
 	private analyzer: Analyzer;
@@ -28,14 +29,19 @@ export class DartRenameProvider implements RenameProvider {
 				}
 			}).then(resp => {
 				let workspaceEdit = new WorkspaceEdit();
+				let outputChannel = channel.getChannel("Refactorings");
+
+				outputChannel.appendLine(`[INFO] ${resp.change.message}...`);
 
 				let hasError = this.handleProblem(
 					resp.initialProblems
 						.concat(resp.optionsProblems)
-						.concat(resp.finalProblems)
+						.concat(resp.finalProblems),
+					outputChannel
 				);
 
 				if (hasError) {
+					outputChannel.appendLine("[INFO] Rename aborted.");
 					reject("");
 					return;
 				}
@@ -60,7 +66,7 @@ export class DartRenameProvider implements RenameProvider {
 
 				// Wait all openTextDocument to finish
 				Promise.all(promises).then(() => {
-					window.showInformationMessage(resp.change.message);
+					outputChannel.appendLine("[INFO] Rename successful.");
 					resolve(workspaceEdit)
 				});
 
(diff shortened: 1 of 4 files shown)


src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,92 @@
...
+				let workspaceEdit = new WorkspaceEdit();
+
+				let hasError = this.handleProblem(
+					resp.initialProblems
+						.concat(resp.optionsProblems)
+						.concat(resp.finalProblems)
+				);
+
+				if (hasError) {
+					reject("Rename cancelled.");
```
ikhwanhayat: When there an error, messages such as below is shown.
image

If reject is called without parameter, the "Info" is still shown with message "No result". I rather have no "Info" message shown, just the "Error". I am not sure how to do this.

src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,92 @@
...
+import * as as from "../analysis/analysis_server_types";
+
+export class DartRenameProvider implements RenameProvider {	
+	private analyzer: Analyzer;
+	constructor(analyzer: Analyzer) {
+		this.analyzer = analyzer;
+	}
+	
+	provideRenameEdits(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {		
+		return workspace.saveAll(false).then(() => {
```
DanTup: Do we actually need to `saveAll` first? Seems unexpected, and the analysis service has all of the in-memory edits. If I was just working in a new file and renaming stuff, I'm not sure I'd expect that (obviously if it's required for some reason, it's not a huge issue!).

src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,92 @@
...
+			return this.doRename(document, position, newName, token);
+		});
+	}
+
+	private doRename(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
+        return new Promise<WorkspaceEdit>((resolve, reject) => {
```
DanTup: Indenting looks weird here - `return` shows as only one char past `private`?

src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,92 @@
...
+			this.analyzer.editGetRefactoring({
+				kind: "RENAME",
+				file: document.fileName,
+				offset: document.offsetAt(position),
+				length: position.character,
```
DanTup: I'm not sure this is correct; `position.character` is the number of characters into the current line? If you're 60 characters into a line, won't this send a length of 60?

src/providers/dart_rename_provider.ts
DanTup: Weird, the docs do say:

> The lack of a result can be signaled by returning undefined or null.

What happens if you pass an empty array instead?

I'll try and have a play around with this later and see if I can find a way to make this better.

src/providers/dart_rename_provider.ts
ikhwanhayat: > the analysis service has all of the in-memory edits

Oh, I'm not aware of that. Yeah agree it would be unexpected, especially for unsaved files that's not affected by the rename. I've removed the `saveAll`, seems to work better this way.

src/providers/dart_rename_provider.ts
ikhwanhayat: Accidentally used space for indention. Fixed it.

src/providers/dart_rename_provider.ts
ikhwanhayat: You are actually right. I misunderstood the `position` parameter. Fixed it.

Whats weird was analysis server did the refactoring right even with wrong params! I think analysis server can read my mind.

src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,91 @@
...
+									document.positionAt(fileEdit.offset + fileEdit.length)),
+								fileEdit.replacement
+							);
+						});
+					});
+				});
+
+				// Wait all openTextDocument to finish
+				Promise.all(promises).then(() => {
+					window.showInformationMessage(resp.change.message);
```
ikhwanhayat: Having second thoughts about showing success message everything rename is done. Do you the popup would be annoying?

src/providers/dart_rename_provider.ts
DanTup: Yeah; this is how we get realtime errors/etc. We send all edits straight over to the analysis server.

There's a hidden config option (`analyzerDiagnosticsPort`?) that lets you see a bunch of stuff, including the "overlays" it has in memory if you ever need to debug what the server thinks a file looks like in the editor.

src/providers/dart_rename_provider.ts
DanTup: The edits will have had their own ranges, so I guess the AS server just used the first symbol it found at the offset (maybe the length doesn't matter for this refactoring; it would be more important for things like "extract method" if such a thing exists).

src/providers/dart_rename_provider.ts
DanTup: Possibly. I'm gonna go through other 0.8 stuff then I'll grab this code and have a play around (and do some comparison to TS) and see how it feels.

src/providers/dart_rename_provider.ts
```diff
@@ -0,0 +1,91 @@
...
+		this.analyzer = analyzer;
+	}
+
+	provideRenameEdits(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
+		return this.doRename(document, position, newName, token);
+	}
+
+	private doRename(document: TextDocument, position: Position, newName: string, token: CancellationToken): Thenable<WorkspaceEdit> {
+		return new Promise<WorkspaceEdit>((resolve, reject) => {
+			var wordRange = document.getWordRangeAtPosition(position);
```
DanTup: Docs say for rename:

> It is an error if the range contains anything other than all or part of the name of a single function (including methods, getters and setters), variable (including fields, parameters and local variables), class or function type.

Possibly it's safe to pass 0 or 1 as the length here to avoid relying on `getWordRangeAtPosition` doing a good thing (though maybe if it doesn't, we can improve language options to fix it).

DanTup:
Didn't get time to look at this tonight.. I pushed 0.8 out with the other things that were done and will try to look at this in next few days.

DanTup:
@ikhwanhayat I haven't forgotten about this! I've been off work unwell today so haven't been at a PC. With any luck I'll get to it tomorrow. Sorry for the delay!

ikhwanhayat:
Now worries. I'm rather busy this week, cant help much I think. Get well soon!

DanTup:
Sorry for the delay; been unwell longer than I expected (hoped?)! Looking at this now!

First I did some testing of TypeScript, and man was I disappointed..

- Renaming to something that already exists - no error, just does it anyway (then gives compile errors)
- Renaming to something invalid (rename a class to `foo.bar`) - no error, just does it anyway (then gives compile errors)

It's nice to be consistent, but let's not copy this :)

Anyway, done some testing, here's some feedback:

### 1. Incorrect refactor of constructor

I have this code:

```dart
twitter = new TwitterApi(liveConsumerKey, liveConsumerKeySecret,
    liveAccessToken, liveAccessTokenSecret);
```

I hit `F2` on `TwitterApi` and appended a 2 (`TwitterApi2`). It fails with `Warn: Constructor name should start with a lowercase letter`. Seems a bit strange, so I tried renaming it to `twitterApi2` and it ended up applying an edit like this:

```dart
twitter = new TwitterApi.twitterApi2(liveConsumerKey, liveConsumerKeySecret,
    liveAccessToken, liveAccessTokenSecret);
```

So I think something has gone wrong (possibly with the offset... apologies if that was my fault!).

### 2. Info messages seem worthless

I agree with your here, seems annoying to show the info popup (especially when it persists) when the rename is fine. Maybe we should console.log these for debugging purposes (maybe we should console.`log`/`warn`/`error` all of the responses) but not bother the user with them.

I think the error works fine (eg. if I try to rename to something with a dot in it).

### 3. Redundant errors

Some of the errors that come back seem to overlap a lot (eg. I tried to rename a variable to a name already used and I got `Library already declares top level variable with name queueFile` and `Library already declares getter with name 'queueFile'`. I think we should probably just display the first error (or if we can add newlines, join them into one message, or show "(+ 1 more)" on the end. Making the user dismiss a bunch of errors could be frustrating.


Other than that, this is excellent! I don't mind tweaking `2`/`3` after merging, but I think `1` needs investigating first. If you can't repro lmk and I'll try in a clean file I can easily share.

ikhwanhayat:
> ### 1. Incorrect refactor of constructor

I've only tested doing renames at the definition, never at invocations. 

So when I tried it, funcs/methods/fields/vars seems to be working ok. Rename at class instantiation will turn out like what you've tested. In fact, I can hit `F2` on keyword `new` and it will run the same refactoring. 

On further test, apparently VSCode allows `F2` on keywords. On some keyword Analysis Server just throws error like `Unable to create refactoring`. But on some AS will run certain refactoring.

Like for `import`, these are my observation.

`import './somefile.dart';`

If I `F2` at `import`, and put in `import2` it refactors to

`import "./somefile.dart" as import2;`

I can do an `F2` on `as` as well, analysis server will rename the library prefix (`import2`).

I'm really not sure how should we handle this.

> ### 2. Info messages seem worthless

I will remove the success error message.

> ### 3. Redundant errors

I will just show the first error message like @devoncarew suggested in the other thread. I think I will create an output channel dedicated for refactorings to user can still see the other messages.

ikhwanhayat:
I can observe the same behavior for renaming `import` and for constructor in Atom.

src/utils.ts
```diff
@@ -105,3 +105,21 @@ function checkIsDevelopment() {
 export function log(message: any): void {
 	console.log(message);
 }
+
+const _keywords = [
```
DanTup: It feels a bit naff having to maintain this list here; the analysis service is supposed to stop us needing to know things like this.

What happens without this? Wouldn't we just fail the refactor with an error as normal?

DanTup:
Ok, this all sounds reasonable to me! I think the behaviour on `imports` sounds good. I'll post a question to the analyzer group about the constructor rename to see if it's intended.

I've added a comment about the keyword "blacklist" though, not really keen on the idea of maintaining that - what are the implications of not having it? If it's just showing an error message I don't think that's too bad (though I wonder what TypeScript does here - seems like maybe there should be an API for the IDE to check if something can be renamed... if TS handles this poorly, maybe we can raise a feature request).

ikhwanhayat:
Please view this screencast on my findings (this is **without** the keyword/blacklist checking done):

https://immetra.tinytake.com/sf/OTIxMjA0XzM4OTUxNDE

In summary:

1. When rename at keyword `import`, AS will add a library prefix, or rename the prefix if already exists.
2. Rename at keyword `as` and the prefix yield the same result as (1)
3. Rename at `class` and the class name (definition) will rename the class name.
4. Rename at class instantiation will try to rename a named constructor, or will change the default constructor to a named constructor.
5. For (4) above, if no constructor is defined, AS replied with `Rename Constructor '' to 'something'...` but no edits are returned.
6. Rename at keyword `new` yield the same result as (4).

Regarding on doing a rename at `import` to add a library prefix, I can see some value in that. But maybe this should be done via a specific refactoring, not through rename. I wonder if CodeActionProvider can be used for this?

The problem I see is that when user do rename at keyword like `import`, `class`, and `new`, the popup textbox is populated with the keyword, hence it's confusing. 

I also dislike having to maintain the list of keywords, might be problematic in the future. It is best if VS Code itself don't allow rename on reserved keywords. 

Options we have:

1. Just list down the problematic keywords like `import`, `class`, and `new`. Other keywords seems safe since AS will throw error on renaming them.
2. Keep whole list of reserved words to make the behavior consistent.  How often does reserved keywords change anyway :)
3. Don't do any checking at all, let AS handle it as it sees fit.

DanTup:
Thanks for the video, this is all much clearer!

I think one of these is a bug (the missing edits on constructor rename), and the other actions a bit weird (seemingly renaming the "nearest thing")..

I've started a discussion about these [in the analyzer-discuss group](https://groups.google.com/a/dartlang.org/d/msg/analyzer-discuss/PLdn5Y4DWaQ/kxdoUYRMDAAJ). We'll see what responses we get and then decide what to do. Possibly we'll have to go with the blacklist for now and maybe in the future the AS will have something that can support how Code's renames work a little better.

src/utils.ts
DanTup: Btw, underscore prefixes for private are Dart, TypeScript uses private, I think this can be changed (I think both @devoncarew and I have both done this too... writing one language while thinking about another is confusing!).

DanTup:
Ok, overnight it occurred to me that the blacklist won't work reliably either, consider:

```dart
main() {
}

import() {
}
```

It's totally valid to have a function called `import` and it collides with our blacklist. In order to work around this, we'd have to start getting really clever with how we blacklist stuff.

I'm starting to think maybe we should park this and hope we can get some changes to the analyzer to better support Code, unless you or @devoncarew have any good ideas?

ikhwanhayat:
I think we should just release it with notes on the wonky parts. At least people can start to use the rename feature where it is normally needed. In practice, unless accidentally, I don't think people would hit rename on the keywords anyway. Dart Atom released the rename feature with same behavior.

About whether to include the "blacklist" word checking, I'm divided on this. I will follow your judgement.

DanTup:
I don't think the blacklist works because of colliding with legit variable names etc., but without it, this behaviour is worst :(

My concern with shipping rename in this state is that we can't guarantee we can fix this for a v1 and I don't want to ship v1 with known bugs (even if they're documented; this behaviour looks really bad IMO) :(

It might be more work, but I wonder if we can make the blacklist more reliable by not just being a hard-coded list but using something like hover/highlight information to know what's actually at the cursor (see if it really is a keyword)? Though maybe we should see what response we get in the analyser group first before devoting time?

devoncarew:
Hey, jumping in at the end of this discussion, but I think the question is about providing faster feedback when the user tries to rename keywords? If that's the case, I think it's OK to try and do the rename and fail when the analysis server returns an error. It'll be a pretty infrequent use case, and doing it this way matches the current VSCode UX. My 0.02 :)

DanTup:
Unfortunately it's not just about failing - the AS does extra (and in some cases unexpected) thinks (but does not error).

Eg., imagine this:

```dart
var myThing = new MyThing();
```

If you put your cusor on `MyThing` and hit `F2` and change what's in the box from `MyThing` to `MyThing2` you end up with:

```dart
var myThing = new MyThing.MyThing2();
```

It adds a named constructor. This isn't really compatible with how Code made it look (pre-populating with the class name) :(

Unless the AS is changed, I think whatever we do is going to suck or look really silly to the user.

DanTup:
I might be completely mad; but I'm going to have a little play around with the analysis server code and see how hard it is to tweak. We have a growing number of open issues that would need changes to the server (that only really benefit Dart Code and probably not other integrations) so I think being able to send PRs (or run on a fork) could be useful.

ikhwanhayat:
Yay, go for it! 😉

DanTup:
Related discussion: https://github.com/Microsoft/vscode/issues/7340
My comments: https://github.com/Microsoft/vscode/issues/7340#issuecomment-242308838

DanTup:
I'm merging this into a new branch (`rename`) so we have it in the main repo. I really want to ship this but I'm just not happy with how it is.

I don't think we'll get AS changes soon, but I think we can look at some alternatives (such as using highlight or other information to figure out if we're on the problematic items).

DanTup:
@ikhwanhayat Forgot to notify you, I shipped this in v0.10 last night. Thanks for the work! :)

ikhwanhayat:
@DanTup Yeah  I noticed, i just got really busy with work these past weeks. Good job in getting that oldName-newName solution.

PR #106, opened 2016-08-17
Fixed issue when sdkpath has spaces

jrote1:


src/commands/pub.ts
```diff
@@ -36,7 +36,7 @@ export class PubManager {
 		// return the paths for tools in the bin/ dir. 
 		let pubPath = path.join(this.sdk, "bin", "pub");
 		channel.appendLine(`[${shortPath}] pub ${command}`);
-		let process = child_process.exec(`${pubPath} ${command}`, { "cwd": projectPath });
```
DanTup: Although it's probably unlikely, it's actually possible to have double quotes in directory names in Linux!

```
danny@raspberrypi:~$ touch "\""
danny@raspberrypi:~$ ls -l
total 12
-rw-r--r-- 1 danny danny    0 Aug 17 17:45 "
drwxr-xr-x 6 danny danny 4096 Aug  6 19:03 Documents
drwxr-xr-x 4 danny danny 4096 Aug 14 20:32 Downloads
drwxr-xr-x 3 danny danny 4096 Jul 25 20:42 Scripts
danny@raspberrypi:~$ 
```

So I wonder if we can fix this in a way that will always work? Apparently `execFile` doesn't have this flaw, but its behaviour is slightly different:

> The child_process.execFile() function is similar to child_process.exec() except that it does not spawn a shell. Rather, the specified executable file is spawned directly as a new process making it slightly more efficient than child_process.exec().

I'm not sure if this will work (I don't know the full implications of that difference), maybe worth trying?

DanTup: Did some testing on Raspbian, and I think just escaping the quotes with blackslash will resolve this:

```
`"${pubPath.replace(/"/g, "\\\""")}" ${command}`
```

jrote1:
@DanTup I have added this change still works fine on windows aswell. I will not be able to test on linux until tonight

DanTup:
Ok, this looks fine to me if we can confirm it's ok on linux. Let me know if you manage to test, and if not I'll try and fix up my VM tomorrow. Thanks!

jrote1:
I can confirm that it works on linux

DanTup:
Great; thanks!

Issue #97, opened 2016-08-16
Add common dart snippets


DanTup:
[https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributessnippets](VS Code docs)

DanTup:
@devoncarew Are there any "standard"-ish snippets anywhere we can start with?

PR #108, opened 2016-08-17
Implement Snippets

ikhwanhayat:
Issue #97

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -45,14 +45,6 @@
 				"}"
 			]
 		},
-		"assert": {
-			"prefix": "assert",
-			"body": "assert($1);"
-		},
-		"print": {
-			"prefix": "print",
-			"body": "print($1);"
-		},
 		"for": {
 			"prefix": "for",
 			"body": [
@@ -64,8 +56,8 @@
 		"for in": {
 			"prefix": "fori",
 			"body": [
-				"for (var x in ${1:list}) {",
-				"  $2",
+				"for (var ${1:item} in ${2:items}) {",
+				"  $3",
 				"}"
 			]
 		},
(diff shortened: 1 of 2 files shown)


snippets/dart.json
```diff
@@ -0,0 +1,121 @@
...
+			"body": [
+				"for (var i = 0; i < ${1:count}; i++) {",
+				"  $2",
+				"}"
+			]
+		},
+		"for in": {
+			"prefix": "fori",
+			"body": [
+				"for (var x in ${1:list}) {",
```
DanTup: Should we make this `x` a label the user can tab to to rename?

snippets/dart.json
```diff
@@ -0,0 +1,121 @@
...
+		},
+		"do while": {
+			"prefix": "do",
+			"body": [
+				"do {",
+				"  $2",
+				"} while ($1);"
+			]
+		},
+		"void": {
```
DanTup: Many of these don't seem to work for me (`print`, `void`), because the completion provider is also providing completion for them and seems to take precedence, so I think maybe we should remove those ones? (I also think it's slightly confusing that sometimes you tab and get the semicolon on the end, and sometimes don't?

DanTup: Strangely, the `import` one works over the completion one (which I kinda like, since I was using double-quotes :D).

snippets/dart.json
```diff
@@ -0,0 +1,121 @@
...
+				"${3:void} ${1:name}(${2:args}) {",
+				"  $4",
+				"}"
+			]
+		},
+		"import": {
+			"prefix": "import",
+			"body": "import \"$1\";"
+		},
+		"class": {
```
DanTup: The `class` one has the same issue that the completion provider takes precedence, which is pretty annoying. I don't think we should remove it though, user will just have to select it (we could filter it out in the completion provider, but that's probably a bit weird!).

DanTup:
These work great! I added a couple of minor comments for you to review. Thanks!

ikhwanhayat:
@DanTup I made changes based on your comment.

DanTup:
Sorry for still not looking at this; still rather unwell. As soon as I can sit at a PC a while comfortably, catching up with these PRs and things are top of the list!

ikhwanhayat:
Take your time @DanTup ... Hope you get well soon.

Issue #57, opened 2016-08-11
add the ability to debug dart cli apps


devoncarew:

devoncarew:
I'd like to take a stab at this; targeted for 1.0? post 1.0?

DanTup:
I put it in post-v1 because I don't know how big a job it is or when we'd start but it'd be neat to have for v1. The future milestones are a bit wolly (just goals, but I'll shuffle things around that are done/not done nearing a release).

I'll move to v1 if you think it's reasonable to do (there's no real plan for when v1 should be, other than it should have at least all the basics and not feel unfinished) :)

DanTup:
This is now in master and will be in 0.10. Thanks!

PR #112, opened 2016-08-18
Debugger implementation

devoncarew:
A fairly complete implementation of a debugger:

- breakpoints
- stepping
- async stepping
- breaking at exceptions
- support multiple isolates
- exp. evaluation on hover
- watchpoint expressions
- repl exp. eval

This may need to be iterated on a bit. I'm having some trouble with an import of a JS `ws` packages. Probably best to land this after a major release? I can also push a PR to a branch on the Dart-Code repo, so we can iterate more easily there.

@DanTup

### DIFF
--- a/package.json
+++ b/package.json
@@ -146,6 +146,11 @@
 		"debuggers": [{
 			"type": "dart-cli",
 			"label": "Dart command line",
+			"enableBreakpointsFor": {
+				"languageIds": [
+					"dart"
+				]
+			},
 			"program": "./out/src/debug/debug_entry.js",
 			"runtime": "node",
 			"configurationAttributes": {
@@ -194,7 +199,8 @@
 	},
 	"dependencies": {
 		"vscode-debugadapter": "^1.11.0",
-		"vscode-debugprotocol": "^1.11.0"
+		"vscode-debugprotocol": "^1.11.0",
+		"ws": "^1.1.1"
 	},
 	"devDependencies": {
 		"typescript": "^1.8.5",
--- a/src/debug/debug_impl.ts
+++ b/src/debug/debug_impl.ts
@@ -1,4 +1,4 @@
-'use strict';
+"use strict";
 
 import * as child_process from "child_process";
 import * as path from "path";
@@ -9,6 +9,18 @@ import {
 } from "vscode-debugadapter";
 import { DebugProtocol } from "vscode-debugprotocol";
 import { readSdkPath } from "./sdk_path"
+import { PackageMap, uriToFilePath, fileToUri, PromiseCompleter } from "./utils";
+import {
+	ObservatoryConnection, VMEvent, VMIsolateRef, RPCError, DebuggerResult, VMStack, VMSentinel, VMObj,
+	VMFrame, VMFuncRef, VMInstanceRef, VMScriptRef, VMScript, VMSourceLocation, VMErrorRef, VMBreakpoint,
+	VMInstance, VMResponse, VMClassRef
+} from "./debug_protocol";
+
+// TODO: modules? ModuleEvent
+// TODO: supportsSetVariable
+// TODO: class variables
+// TODO: library variables
+// TODO: break large arrays into subarrays
 
 export interface DartLaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
 	cwd: string;
@@ -22,10 +34,15 @@ export class DartDebugSession extends DebugSession {
 	private childProcess: child_process.ChildProcess;
 	private processExited: boolean = false;
 	private sdkPath = readSdkPath();
-	private dartPath = this.sdkPath != null ? path.join(this.sdkPath, "bin", "dart") : "dart"; 
+	private dartPath = this.sdkPath != null ? path.join(this.sdkPath, "bin", "dart") : "dart";
+	observatory: ObservatoryConnection;
+	private threadManager: ThreadManager;
+	private packageMap: PackageMap;
 
 	public constructor() {
 		super();
+
+		this.threadManager = new ThreadManager(this);
 	}
 
 	protected initializeRequest(
@@ -34,11 +51,11 @@ export class DartDebugSession extends DebugSession {
 	): void {
 		response.body = {
 			supportsConfigurationDoneRequest: true,
-			// supportsEvaluateForHovers: true,
-			// exceptionBreakpointFilters: [
-			// 	{ filter: "All", label: "All Exceptions", default: false },
-			// 	{ filter: "Unhandled", label: "Uncaught Exceptions", default: true }
-			// ]
+			supportsEvaluateForHovers: true,
+			exceptionBreakpointFilters: [
+				{ filter: "All", label: "All Exceptions", default: false },
+				{ filter: "Unhandled", label: "Uncaught Exceptions", default: true }
+			]
 		};
 		this.sendResponse(response);
 	}
@@ -48,15 +65,16 @@ export class DartDebugSession extends DebugSession {
 		this.sourceFile = path.relative(args.cwd, args.program);
 		this.sendEvent(new OutputEvent(`dart ${this.sourceFile}\n`));
 
+		this.packageMap = new PackageMap(PackageMap.findPackagesFile(args.program));
+
 		this.sendResponse(response);
 
 		let debug = !args.noDebug;
 		let appArgs = [];
-		// TODO:
-		// if (debug) {
-		// 	appArgs.push("--enable-vm-service:0");
-		// 	appArgs.push("--pause_isolates_on_start=true");
-		// }
+		if (debug) {
+			appArgs.push("--enable-vm-service:0");
+			appArgs.push("--pause_isolates_on_start=true");
+		}
 		appArgs.push(this.sourceFile);
 		if (args.args)
 			appArgs = appArgs.concat(args.args);
@@ -69,8 +87,16 @@ export class DartDebugSession extends DebugSession {
 
 		process.stdout.setEncoding("utf8");
 		process.stdout.on("data", (data) => {
-			// TODO: look for 'Observatory listening on...'
-			this.sendEvent(new OutputEvent(data, "stdout"));
+			let match: RegExpExecArray;
+			if (!this.observatory) {
+				match = ObservatoryConnection.portRegex.exec(data);
+			}
+
+			if (match) {
+				this._initObservatory(match[1]);
+			} else {
+				this.sendEvent(new OutputEvent(data, "stdout"));
+			}
 		});
 		process.stderr.setEncoding("utf8");
 		process.stderr.on("data", (data) => {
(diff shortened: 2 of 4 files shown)


devoncarew:
(fix https://github.com/Dart-Code/Dart-Code/issues/57)

DanTup:
Awesome! I'll try and run through the code on GH later though probably won't get time to test it out until tomorrow.

The `ws` import thing is a bit weird; maybe it's path-related. I don't think we use `require` anywhere else, so wonder if we can change it. I'll see if I can figure out what's up.

I guess to send to another branch, I need to create it first? I'll try and do that shortly and let you know.

DanTup:
@devoncarew There's now a `debugger` branch for this. I can probably just pull your changes and push them there? Not sure if that'll affect PRs in any way!

DanTup:
Apparently I can change the branch!

Though if I merge this PR, where would be best to handle any feedback?

devoncarew:
If you want to merge it, you can add feedback to this original PR, and we can iterate on that branch.

DanTup:
Sorry for still not looking at this; still rather unwell. As soon as I can sit at a PC a while comfortably, catching up with these PRs and things are top of the list!

devoncarew:
Yeah, no hurry. It's a big change and I'm sure will need plenty of time to bake :)

DanTup:
The only way I can satisfy the compiler over the `ws` package is with this change:

WS import

However, even after that change, if I put a breakpoint on the first line of my app (inside `main`) and hit F5 it just runs without ever stopping.

I tried stripping it down to a simple script but same behaviour:

```dart
main() {
  print("test");
}
```

Not sure where to start with debugging (breakpoints in the debug adapter don't seem to be hit).

devoncarew:
Can you land that change and I'll sync up?

Re: the breakpoints, I suspect that it's an issue with Windows paths.
Breakpoints do work for me on macos! (And I suspect they would on Linux as
well)

On Aug 20, 2016 7:45 AM, "Danny Tuppeny" <notifications@github.com> wrote:

The only way I can satisfy the compiler over the ws package is with this
change:

[image: WS import]
<https://cloud.githubusercontent.com/assets/1078012/17831852/aa889e34-66ec-11e6-9cba-5835f910d62d.png>

However, even after that change, if I put a breakpoint on the first line of
my app (inside main) and hit F5 it just runs without ever stopping.

I tried stripping it down to a simple script but same behaviour:

main() {
  print("test");
}

Not sure where to start with debugging (breakpoints in the debug adapter
don't seem to be hit).

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Dart-Code/Dart-Code/pull/112#issuecomment-241203980>,
or mute the thread

DanTup:
Sure, done. Not sure if it's the best fix, though it makes sense we can't have a type on `socket` without bindings for `ws`. I do think DefinitelyTyped has some, but in the past we've had issues with their bindings (it's a big grab bag, not versioned very well, etc.) so might not be worth the effort.

Didn't think about paths - I can try and do some debugging tonight or tomorrow on that if it's hard for you to diagnose on Mac.

DanTup:
There are a few places that have paths that I wonder if might have the wrong separators (like `uriToFilePath`, `fileToUri`). Is there a better way to debug this than littering the file with appending to files to see what's going on?

devoncarew:
I added this: https://github.com/Dart-Code/Dart-Code/pull/118 to print out the debugger traffic (to and from the VM).

I suspect something like the VM is executing the code using windows style paths (`C:\foo\bar\baz.dart`), but we're setting breakpoints using unix style paths (`/foo/bar/baz.dart`). So none of our breakpoints actually match scripts the VM thinks are running.

You should be able to see the actual paths the VM is using for the scripts by looking at the wire traffic. The solution _may_ be to improve those uri methods, or it may be to do more targeted work when we encode and decode script paths.

screen shot 2016-08-20 at 12 24 22 pm

devoncarew:
screen shot 2016-08-20 at 12 34 19 pm

DanTup:
Great; thanks! If I don't get a minute to try it shortly, I'll have a dig in the morning and see what's up.

DanTup:
Trying a few things, but no matter how I send the `scriptUri`, it always seems to report as unresolved:

```json
==> {"id":"3","method":"addBreakpointWithScriptUri","params":{"isolateId":"isolates/911301492","scriptUri":"file://m:/Coding/Applications/NewsWoWBot/bin/danny.dart","line":2}}

<== {"jsonrpc":"2.0", "result":{"type":"Breakpoint","fixedId":true,"id":"breakpoints\/1","breakpointNumber":1,"resolved":false,"location":{"type":"UnresolvedSourceLocation","scriptUri":"file:\/\/m:\/Coding\/Applications\/NewsWoWBot\/bin\/danny.dart","line":2}},"id":"3"}
```

Will keep digging...

DanTup:
Trying to figure out from [the spec](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md) whether there's a way I can get the VM to return a uri for me, so I can see exactly what it thinks they should be. If you know of anything, let me know!

DanTup:
I tried calling `this.callMethod("getSourceReport", { isolateId: isolateId, reports: [ "PossibleBreakpoints" ] });` which I found in the spec but I didn't get any uris back in the results (though they got truncated by Code's log; I can't tell if they would've been in there otherwise).

Out of time for tonight, but will try and dig more tomorrow. If you know any reliable way I can get the VM to tell me what scripts are loaded then that might help track it down.

DanTup:
Reading more... I think I can probably send `addBreakpointAtEntry` then `getStack` later (or even just `getStack` in place of the `resume`) and examine the SourceLocation in the resulting stack frames.

I'll give it a go tomorrow. I might have no idea what I'm doing!

devoncarew:
I'm afk right now, but the stream of info you get just starting up will
include a few script paths (or it should, to the best of my recollection).

There are also some calls to get info about the current VM, and to get some
data about the current isolate. The isolate one should have some references
to scripts.

DanTup:
So, the initial stuff from the VM doesn't seem to include any paths:

```
==> {"id":"0","method":"streamListen","params":{"streamId":"Isolate"}}
==> {"id":"1","method":"streamListen","params":{"streamId":"Debug"}}
<== {"jsonrpc":"2.0","id":"0","result":{"type":"Success"}}
<== {"jsonrpc":"2.0","id":"1","result":{"type":"Success"}}
<== {"jsonrpc":"2.0","method":"streamNotify","params":{"streamId":"Isolate","event":{"type":"Event","kind":"IsolateRunnable","isolate":{"type":"@Isolate","fixedId":true,"id":"isolates\/428829441","name":"danny.dart$main","number":"428829441"},"timestamp":1471775516227}}}
==> {"id":"2","method":"setExceptionPauseMode","params":{"isolateId":"isolates/428829441","mode":"Unhandled"}}
```

However, I was able to get the Observatory url and load it in my browser (I removed the `resume` call in the debug protocol so it was paused). It shows the uri as starting with three slashes:

Observatory

So I tweaked the code to send that exact uri when setting the breakpoint:

```
==> {"id":"3","method":"addBreakpointWithScriptUri","params":{"isolateId":"isolates/428829441","scriptUri":"file:///m:/Coding/Applications/NewsWoWBot/bin/danny.dart","line":2}}
```

However that still returns that it's unresolved:

```
<== {"jsonrpc":"2.0","method":"streamNotify","params":{"streamId":"Debug","event":{"type":"Event","kind":"BreakpointAdded","isolate":{"type":"@Isolate","fixedId":true,"id":"isolates\/428829441","name":"danny.dart$main","number":"428829441"},"timestamp":1471775516233,"breakpoint":{"type":"Breakpoint","fixedId":true,"id":"breakpoints\/1","breakpointNumber":1,"resolved":false,"location":{"type":"UnresolvedSourceLocation","script":{"type":"@Script","fixedId":true,"id":"libraries\/15\/scripts\/file%3A%2F%2F%2Fm%3A%2FCoding%2FApplications%2FNewsWoWBot%2Fbin%2Fdanny.dart\/156aca82642","uri":"file:\/\/\/m:\/Coding\/Applications\/NewsWoWBot\/bin\/danny.dart","_kind":"script"},"line":2}}}}}
```

I was making the assumption this wouldn't show as unresolved if it was correct, but that might not be correct. However, I tried it anyway, and it worked, I'm now at a breakpoint!

So, I'm pushing a tweak to fileToUri to produce the correct output on Windows:

```typescript
	// Handle windows paths; slashes must be converted and we need an extra slash prefixed.
	file = file.replace(/\\/g, "/");
	if (!file.startsWith("/"))
		file = "/" + file;
```

Now I think we're going to need a similar change in `uriToFilePath` but I can't figure out from the code how I can test what that's doing in order to confirm it's broken and/or a fix works. Could you explain where I can see the effects of that to test?

DanTup:
This is really awesome! Really chuffed about this; I'm not sure how far I'd have gotten trying to implement this myself! :+1: 

Some comments from a quick testing (I'm putting checkboxes in to aid tracking; we can tick them off as we've either fixed, discussed and deemed nothing to change, or broken off into issues for later).

1. [ ] Pressing `F10` (step over) on the last line of `main` breaks me into `dart:isolate-patch/isolate_patch.dart` at the end of `port.handler`
1. [ ] Probably related, the call stack lets me walk up into the same file; again seems a bit weird, but possibly this is expected (seems a little weird to expose non-user Dart code to them though?)
1. [ ] Pressing `F11` steps into SDK functions; I both like and dislike this. In C# this is an option you can enable (off by default) so that you can easily `F11` into your own code without constantly dipping into framework stuff and have to drop back out; I know Dart is a bit different but presumably the same frustration would exist without an option?
1. [ ] Watch window shows `\n`, not sure if this can be improved (not sure we can fit newlines in there, but this looks weird - screenshot below)
1. [ ] Some values in watch window/repl don't seem to work (maybe async related - see screenshot below)
1. [ ] Hovering over some variables in the editor when at a breakpoint doesn't seem to do anything (maybe async related - see previous)

Newlines in Watch Window

Watch error for awaited value

Other than that, this is awesome! The watch window works great, the locals window, I can walk up and down the stack, it breaks on exceptions as expected in both modes and even steps through async/await code as I would expect! :)

DanTup:
@devoncarew I've merged master into the debugger branch because it was getting behind and had a bunch of conflicts (due to `package.json` being formatted and my SDK path changes on adjacent lines to your changes); it'll be worth you updating.

devoncarew:
> I was making the assumption this wouldn't show as unresolved if it was correct

I think all breakpoints will return unresolved before the isolate starts execution. I believe the VM doesn't know which locations are legal executable locations before it's parsed and compiled a method. We'll get a follow-on event if the breakpoint is later resolved.

devoncarew:
Re: stepping into dart: library code, this can be controlled by sending a call to make that library not debuggable. A step in then would become a stepover for method in those libraries. Possibly something to expose as a user option.

And yeah, expression evaluation does not for for `async` methods. This is a bug, and does have a lot of impact on the usability of debugging async methods. Tooltips, the repl, and watchpoints are all impacted. Even something like trying to eval `1+2` when at a breakpoint in an async method won't work.

devoncarew:
Also, awesome you got the debugger issue fixed on windows! 🎉

DanTup:
> Re: stepping into dart: library code, this can be controlled by sending a call to make that library not debuggable. A step in then would become a stepover for method in those libraries. Possibly something to expose as a user option.

Cool; I was worried you'd say it wasn't possible. Definitely sounds like a good option to add.

> And yeah, expression evaluation does not for for async methods. This is a bug

Is it something likely to be fixed soon, or is it a ongoing complicated issue?

DanTup:
> "Also we need Debugger in decent Editor."
> https://twitter.com/dawopack/status/767350615769055232

I'm not sure what's wrong with the other editors that have debuggers, but maybe there's a good demand for this! :+1:

DanTup:
I've moved my comments out to #125 to avoid them being buried in a comment in the middle of a discussion in a merged PR!

DanTup:
@devoncarew You probably already noticed, but I shipped this last night as 0.10 :-)

Code Actions were disabled, I'm gonna ship them as soon as Code 1.5 goes stable.

devoncarew:
Yup, awesome! I updated to the latest :)

PR #115, opened 2016-08-19
some tweaks to code completion

devoncarew:
A follow-up to https://github.com/Dart-Code/Dart-Code/pull/111:

- show return types for getters
- change how we calc the completion type icons to use the element kind if available
- show the element type in lower case (instead of the default uppercase)

screen shot 2016-08-19 at 1 48 59 pm

### DIFF
--- a/src/providers/dart_completion_item_provider.ts
+++ b/src/providers/dart_completion_item_provider.ts
@@ -1,9 +1,12 @@
 "use strict";
 
-import { TextDocument, Position, CancellationToken, CompletionItemProvider, CompletionList, CompletionItem, CompletionItemKind, TextEdit, Range } from "vscode";
+import {
+	TextDocument, Position, CancellationToken, CompletionItemProvider, CompletionList,
+	CompletionItem, CompletionItemKind, TextEdit, Range
+} from "vscode";
 import { Analyzer } from "../analysis/analyzer";
-import * as as from "../analysis/analysis_server_types";
 import { logError } from "../utils";
+import * as as from "../analysis/analysis_server_types";
 
 export class DartCompletionItemProvider implements CompletionItemProvider {
 	private analyzer: Analyzer;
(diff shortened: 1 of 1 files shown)


DanTup:
Looks great, thanks!

Do you think there's any value in showing the full params instead of `...`?

I'm not at PC to test, but from screenshot in #111 maybe they're shown when selected - not sure if one way might seem better than the other?

I should probably see how it feels at a PC.. I think there are also a bunch of places now where we display similar info we should probably review to see if improvements made in some places are missing in others (for ex. if there are other places still displaying uppercase kinds and maybe duplicate logic for converting to the Code kinds).

devoncarew:
Definitely more room for iteration here. The TS completion just show names when not selected, not `foo(...)` or `foo()`. When selected, both show the full method sig.

This current implementation is pretty close to how the TS completer displays things. Not sure if we want to match that or look for improvements where possible. Other code completion UIs I'm familiar with show the full method+sig for all items, not just the selected one. I'd like to see how it all fits together w/ the param completer; I'm not very familiar w/ what that looks like.

It would be nice if we could complete more of the item. In atom, you can specify some insertion points and placeholders for params. So something like `foo({0})` would complete foo() but leave the cursor between the parans. Or `foo({bar}, {baz})` would complete the method, select the `bar` text, let you change the name, and hitting tab would then select the `baz

DanTup:
> Not sure if we want to match that or look for improvements where possible.

I think the TS one is a good reference to roughly copy, but if we think we can improve it (we've already seen it has "quirks") then I think we should try it out!

> I'd like to see how it all fits together w/ the param completer; I'm not very familiar w/ what that looks like.

I made a start on trying to implement this using the `Hover` data (after the idea in [analyzer-discuss](https://groups.google.com/a/dartlang.org/d/msg/analyzer-discuss/Abh1vi1CwDA/3EmELYOaCAAJ)) but not sure it's going to work well (I'd need to walk backwards from the current cursor position to the correct symbol for the method call; but I can't use parens in case it's `foo(x, (x + z))` so it starts getting complicated). Might just have to park this and hope in the future the AS gets a feature (this is actually the biggest stumbling block when I sat down to try and write some real Dart the other day; it's really hard when you don't know the SDK!).

> It would be nice if we could complete more of the item.

Yeah, I agree. Maybe it's worth raising a feature for Code? (maybe including a screenshot of how Atom handles it will help).

PR #119, opened 2016-08-21
Provide SDK path to debugger as a command argument.

DanTup:
@devoncarew This removes the hack to use temp files to get the SDK path to the debugger. You may need to delete/re-created your `launch.json` or add `sdkPath` to it (it should auto-complete the default value).


devoncarew:
👍

PR #121, opened 2016-08-22
add two files to .gitignore

devoncarew:
- add two files to .gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 out
 node_modules
-.vscode-test/
 analyzer.log
+.vscode/.browse.VC.db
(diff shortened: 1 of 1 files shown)


DanTup:
No idea what this file is, seems to be created by the C++ extension!

https://github.com/Microsoft/vscppsamples/issues/14

devoncarew:
Ah, I thought it was a new file created by VSCode; apparently just cruft from an extension I installed!

DanTup:
I wondered why I hadn't seen it! Adding to ignore is cheap so if it might avoid others accidentally adding stuff, worth having!

Strange that it seems that extension must be activating for Dart Code though!

PR #132, opened 2016-08-27
fix race condition at debugger start

devoncarew:
Fix an issue w/ a race condition at start. For small applications, we could connect after isolates had already been created.


DanTup:
Awesome, thanks! :)

PR #139, opened 2016-08-28
Send two breakpoints; the real path and the package path.

### LINKED ISSUE #134: Breakpoints don't get hit (possibly if imported with relative paths instead of `package:`)
1. Clone https://github.com/Dart-Code/analysis-server-typescript-generator at revision 514fd331c984a3808e3cf9a6cdcea7e1f4a66e21
1. Put a breakpoint on first line of `main` and another on first line of `ApiSpec.buildDefinitions`
1. Press `F5`; first breakpoint is hit
1. Press `F5` again; program runs to completion without hitting second breakpoint

DanTup:
@devoncarew this seems to fix #134, but I don't know if it's the best way to handle it. I now just create an array for the uris, always including the source path regardless of whether we can make a package path.

Removing the breakpoint seems to also do the right thing (I didn't need to do anything here, seems breakpoints are removed by just calling the same method, which removes old and adds again).

If you can think of any complications or edge cases (or if this doesn't work for you), let me know.


DanTup:
Git did a really bad job with this diff... the body of the code is the same, just wrapped in a `forEach` (and thus indented on extra level) but it's just mashes it all up making the diff harder to review! :(

src/debug/debug_impl.ts
```diff
@@ -181,20 +181,33 @@ export class DartDebugSession extends DebugSession {
...
-			let bpResponse = [];
-			for (let verified of result) {
-				bpResponse.push({ verified: verified });
-			}
+		let uris = this.getBreakpointUris(source.path);
 
-			response.body = { breakpoints: bpResponse };
-			this.sendResponse(response);
-		}).catch((error) => this.errorResponse(response, `${error}`));
+		uris.forEach(uri => {
```
devoncarew: You may want to add some quick docs here to say why we're doing this (to support both file: and package: uris at runtime, for the same file).

devoncarew:
Code looks good! The only edge cases I can think of are to verify that removing breakpoints works, and it sounds like you did that :)

src/debug/debug_impl.ts
DanTup: Good point; I will add :)

DanTup:
Yep, removing them worked fine - it seems to just go through the same method again (with an empty breakingpoint collection against the (pair of) uris).

I'll add some comments then merge this into the `debugger` branch tonight or tomorrow.

PR #154, opened 2016-09-01
run the formatter over source files

devoncarew:
- run the formatter over the  source files
- tweak the .gitignore file

### DIFF
--- a/src/editors.ts
+++ b/src/editors.ts
@@ -4,7 +4,7 @@ import * as vs from 'vscode';
 
 export function hasActiveDartEditor(): boolean {
   if (!vs.window.activeTextEditor)
-  	return false;
+    return false;
 
   return vs.window.activeTextEditor.document.languageId === "dart";
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 out
 node_modules
 analyzer.log
-.vscode/.browse.VC.db
-.vscode/.browse.VC.db-wal
+.vscode/.browse.VC.db*
 .vscode-test/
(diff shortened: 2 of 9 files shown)


devoncarew:
@DanTup, looks like this is failing:

- https://travis-ci.org/Dart-Code/Dart-Code/jobs/156691867#L99
- https://github.com/Dart-Code/Dart-Code/blob/master/src/providers/dart_completion_item_provider.ts#L89

devoncarew:
Possibly a VSCode insiders API?

DanTup:
Urrrgggg...

That thing did change in Insiders, but to keep my new branch clean, I made the change back in master because I thought it was using it as an interface not a class ;-(

I'll change it to call the constructor (I did initially, but this was a smaller change) today. If you need to build in the meantime, feel free to just delete those two lines (`additionalTextEdits` and `command`).

src/editors.ts
```diff
@@ -4,7 +4,7 @@ import * as vs from 'vscode';
 
 export function hasActiveDartEditor(): boolean {
   if (!vs.window.activeTextEditor)
-  	return false;
```
DanTup: This file seems to have two-space indenting?

DanTup:
Merged this since the failure isn't related; though worth fixing the 2-space file. Maybe setting `editor.detectIndentation: false` in this project would fix it from sticking with spaces where they already exist?

I really wish Code would get format-on-save already! ;(

DanTup:
I've removed the bad code online, so failure should at least be fixed!

PR #182, opened 2016-09-16
update to the latest vscode libs

devoncarew:
Update to the latest vscode APIs.

### DIFF
--- a/package.json
+++ b/package.json
@@ -230,12 +230,12 @@
 		"postinstall": "node ./node_modules/vscode/bin/install"
 	},
 	"dependencies": {
-		"vscode-debugadapter": "^1.11.0",
-		"vscode-debugprotocol": "^1.11.0",
+		"vscode-debugadapter": "^1.12.0",
+		"vscode-debugprotocol": "^1.12.0",
 		"ws": "^1.1.1"
 	},
 	"devDependencies": {
 		"typescript": "^1.8.5",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -230,12 +230,12 @@
 		"postinstall": "node ./node_modules/vscode/bin/install"
 	},
 	"dependencies": {
-		"vscode-debugadapter": "^1.11.0",
-		"vscode-debugprotocol": "^1.11.0",
+		"vscode-debugadapter": "^1.12.0",
```
DanTup: Do you know whether this requires a certain version of VSCode? In my terminal branch I updated these along with the `engine/vscode` property but I couldn't find any info on what actually happened if people were using old versions of Code.

devoncarew: I don't, but I was able to run w/ these versions on the stable channel. We're not really making use of the newer APIs; I see they were published to npm ~2 weeks ago.

devoncarew: So, not critical to land this...

DanTup: If this all still works on Stable with no issues; sounds safe to me. If we build anything on any new APIs we might need to be careful, but I guess whatever new features we add we'll test on Stable whether we think they use new APIs or not.

I'm not too worried about people on pre-stable versions; if people update Code as frequently as Dart Code it seems unlikely they'll be stuck around on old versions.

PR #183, opened 2016-09-16
support paged lists

devoncarew:
Support paged Lists in the debugger:

screen shot 2016-09-15 at 9 33 05 pm

@DanTup 

(and, rename a class in the debugger protocol wrapper to be the same as the spec name; `VMObjRef` ==> `VMObjectRef`)


DanTup:
LGTM; is this dependent on the other change (version numbers)?

devoncarew:
No - I was looking at the new debugging APIs, but we're not making use of any of them.

DanTup:
Great, thanks!

Issue #162, opened 2016-09-04
Review hovers display


DanTup:
I think the hover info could do with some tweaks; such as:

`Danny String name` seems a strange way to present the `name` property of `Danny` class. Should we ditch the class, or show it as `String Danny.name`?

![Danny String name](https://cloud.githubusercontent.com/assets/1078012/18230537/9005abbe-7296-11e6-9b4e-9e0a0ebc3190.png)

This doesn't look right - we have `Future.Future.blah` and the sig seems to be displayed twice.

![Duplicated info](https://cloud.githubusercontent.com/assets/1078012/18230549/d1b1d948-7296-11e6-9458-494cbac6f362.png)

Also, other languages seem to put documentation at top, definition below:

![Ordering](https://cloud.githubusercontent.com/assets/1078012/18230562/f9b0b306-7296-11e6-9051-e8efa09d98b9.png)

Might also want to consider the changes being made to hovers (https://github.com/Microsoft/vscode/issues/2330#issuecomment-243157831).

DanTup:
Scrollbars are being added - we should detect the VS Code version and only truncate if it's prior to this version:

https://github.com/Microsoft/vscode/issues/2330

DanTup:
These all look great with the above changes and @devoncarew's recent changes; considering this done.

PR #184, opened 2016-09-16
improve the dartdoc display

devoncarew:
Improve the display of dartdoc hovers (esp. for the VSCode insiders release):

screen shot 2016-09-15 at 9 52 31 pm

(this still works well with the current stable version of vscode)

@DanTup


DanTup:
Cool; I had #162 about some more tweaks here, I'll see if they're still required with this.

The Future class is a good one for testing hovers, has a bunch of code blocks etc. (in insiders, the code blocks will now correctly use Dart formatting too!).

PR #190, opened 2016-10-09
implement a type hierarchy command

devoncarew:
- implement a type hierarchy command (some code borrowed from the type hierarchy branch)
- fix https://github.com/Dart-Code/Dart-Code/issues/133

This shows the direct supers and all children in a flat list - supers above and children below. The names if implements and mixins are show in the item descriptions but they are not represented as separate elements in the selection list.

screen shot 2016-10-09 at 11 00 40 am

screen shot 2016-10-09 at 11 01 10 am

@DanTup

### DIFF
--- a/src/commands/type_hierarchy.ts
+++ b/src/commands/type_hierarchy.ts
@@ -1,0 +1,131 @@
+"use strict";
+
+import * as editors from "../editors";
+import * as vs from "vscode";
+import * as as from "../analysis/analysis_server_types";
+import { Analyzer } from "../analysis/analyzer";
+import { toRange } from "../utils";
+
+export class TypeHierarchyCommand implements vs.Disposable {
+	private context: vs.ExtensionContext;
+	private analyzer: Analyzer;
+	private commands: Array<vs.Disposable> = [];
+
+	constructor(context: vs.ExtensionContext, analyzer: Analyzer) {
+		this.context = context;
+		this.analyzer = analyzer;
+
+		this.commands.push(
+			vs.commands.registerTextEditorCommand("dart.showTypeHierarchy", this.showTypeHierarchy, this)
+		);
+	}
+
+	private showTypeHierarchy(editor: vs.TextEditor, editBuilder: vs.TextEditorEdit) {
+		if (!editors.hasActiveDartEditor()) {
+			vs.window.showWarningMessage("No active Dart editor.");
+			return;
+		}
+
+		let document = editor.document;
+
+		this.analyzer.searchGetTypeHierarchy({
+			file: document.fileName,
+			offset: document.offsetAt(editor.selection.active)
+		}).then(response => {
+			let items = response.hierarchyItems;
+			if (!items) {
+				vs.window.showInformationMessage('Type hierarchy not available.');
+				return;
+			}
+
+			let options = { placeHolder: name(items, 0) };
+
+			// TODO: How / where to show implements?
+			// TODO: How / where to show mixins?
+			let tree = [];
+			let startItem = items[0];
+
+			tree.push(startItem);
+			addParents(items, tree, startItem);
+			addChildren(items, tree, startItem);
+
+			vs.window.showQuickPick(tree.map(item => itemToPick(item, items)), options).then(result => {
+				if (result) {
+					let location: as.Location = result['location'];
+					vs.workspace.openTextDocument(location.file).then(document => {
+						vs.window.showTextDocument(document).then(editor => {
+							let range = toRange(location);
+							editor.revealRange(range, vs.TextEditorRevealType.InCenterIfOutsideViewport);
+							editor.selection = new vs.Selection(range.end, range.start);
+						});
+					});
+				}
+			});
+		});
+	}
+
+	dispose(): any {
+		for (let command of this.commands)
+			command.dispose();
+	}
+}
+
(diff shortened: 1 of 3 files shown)


DanTup:
Awesome; this looks good! I'm out this evening, but will try and have a look/test tomorrow. Thanks!

DanTup:
I can't figure out why, but there seems to be quite a delay (over half a second?) between selected an item in the list and the cursor/selection moving. From the code, it doesn't seem like there are any requests to the AS or anything, but it's quite noticeable.

Do you see it too? I don't have time now, but I'll try and do some debugging; I suspect it's a Code issue, whether there's a good reason for the delay I'm not sure.

I'm testing with this file; cursor is on `Fred` and I'm selected `Danny` in the picklist.

```dart
import 'dart:core';

main() {
  var ints = new List<int>();
  for (var i = 0; i < 4000; i++)
    ints.add(i);
  print('test');
}


class Danny {}
















class Fred extends Danny {}


class Fred2 extends Fred {}

```

src/commands/type_hierarchy.ts
```diff
@@ -0,0 +1,131 @@
...
+		if (child.subclasses.length > 0)
+			addChildren(items, tree, child);
+	}
+}
+
+function itemToPick(item: as.TypeHierarchyItem, items: as.TypeHierarchyItem[]): vs.QuickPickItem {
+	let desc = '';
+
+	// extends
+	if (item.superclass !== undefined && name(items, item.superclass) != 'Object')
```
DanTup: Since we're hiding "extends Object" would it be better to completely hide Object from the top of the list? Feels a bit weird to include it in the list but not include it in the detail (My preference would be to drop it from the list, though if you think there's good reason to keep it, I'd prefer to drop this exclusion for the "extends.." text).

src/commands/type_hierarchy.ts
```diff
@@ -0,0 +1,131 @@
...
+		desc += `extends ${name(items, item.superclass)}`;
+
+	// implements
+	if (item.interfaces.length > 0) {
```
DanTup: I don't know if this is behaving as expected (but it may be the AS's fault)... Given this code:

```dart
class Danny {}
class Person {}
class Fred extends Danny implements Person {}
class Fred2 extends Fred {}
```

- If I `F4` on `Fred`, in the list I see `Fred extends Danny, implements Person`. 
- If I `F4` on `Danny`, in the list I see `Fred extends Danny`
- If I `F4` on `Person`, in the list I see `Fred extends Person`

It seems strange to have three different representations of the same name in the list depending on what I executed the search on. What do you think?

src/commands/type_hierarchy.ts
```diff
@@ -0,0 +1,131 @@
...
+
+	// with
+	if (item.mixins.length > 0) {
+		if (desc.length > 0)
+			desc += ', ';
+		desc += `with ${item.mixins.map(i => name(items, i)).join(', ')}`;
+	}
+
+	// TODO: Show the library location in the details (dart:core, ...) (share code
+	// with dart_workspace_symbol_provider.ts).
```
DanTup: Do you think we could add this in a nice way? I think with only two labels (name, detail) it might get a bit noisy if we try to put both the extends/inherits/withs as well as the location. I think current implementation might be best we can do until (if) Code give us a custom tree option?

DanTup:
I've no idea why my comment about the delay shows up twice.. Pretty sure it wasn't user error!

DanTup:
lol; three times... I think it's being duped every time I comment?

src/commands/type_hierarchy.ts
```diff
@@ -0,0 +1,131 @@
...
+			let items = response.hierarchyItems;
+			if (!items) {
+				vs.window.showInformationMessage('Type hierarchy not available.');
+				return;
+			}
+
+			let options = { placeHolder: name(items, 0) };
+
+			// TODO: How / where to show implements?
+			// TODO: How / where to show mixins?
```
devoncarew: Don't know; mixins we can probably ignore. W/ implements, people use that quite a bit in Dart to indicate inheritance. One example I was playing around with is `dart:io`'s File and Directory; they both implement FileSystemEntity, but in this case I really think it's more of a subclass arrangement.

src/commands/type_hierarchy.ts
devoncarew: It's odd; the info returned from the AS really is customized based on the 'pivot' class chosen.

src/commands/type_hierarchy.ts
devoncarew: I can remove the todo:. I'd assumed that the `details` wouldn't show expect for selected entries, but didn't confirm.

devoncarew:
Hey - I can take a look at the comments this evening; I'll make some tweaks based on your feedback. Thanks!

devoncarew:
Updated!

I do see the slowdown. Possibly a result of the implementation? I'm not sure where the time is going, but the impl is:

- vs.workspace.openTextDocument(location.file) (`async`)
- vs.window.showTextDocument(document) (`async`)
- editor.revealRange(...), editor.selection = ...

I didn't see a 'jump to location' API call, though there must be an internal implementation that the symbol nav. and type nav. use...

DanTup:
I've opened #194 about the delay, will investigate when I have some time.

src/commands/type_hierarchy.ts
DanTup: Do you think it's worth raising as an issue? It seems a bit weird to get a different picture of the hierarchy depending on this?

PR #191, opened 2016-10-11
remove code to work around a pre 1.6 issue

devoncarew:
- remove some pre vscode 1.6 code - hovers in 1.6 better support longer hover docs

@DanTup

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
 	"publisher": "DanTup",
 	"preview": true,
 	"engines": {
-		"vscode": "^1.4.0"
+		"vscode": "^1.6.0"
 	},
 	"license": "SEE LICENSE IN LICENSE",
 	"bugs": {
--- a/src/providers/dart_hover_provider.ts
+++ b/src/providers/dart_hover_provider.ts
@@ -73,19 +73,6 @@ export class DartHoverProvider implements HoverProvider {
 		if (index != -1)
 			doc = doc.substring(0, index);
 
-		// Truncate long dartdoc.
-		let lines = doc.split("\n");
-		if (lines.length > 20 && !vsCodeVersion.hasScrollableHovers) {
-			for (let index = 20 - 6; index < lines.length; index++) {
-				if (lines[index].trim().length == 0) {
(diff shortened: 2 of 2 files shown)


DanTup:
Currently the extension still works with 1.5 so I'm gonna hold off merging this for now (I don't have stats on how fast people update Code - probably should have!).

It's likely I'll will bump the requirement to 1.6 soon (it's already set on the TypeScript v2 branch, and may be required for the targetting terminal work) so this will be valid then.

Keep this open though; it probably won't be long - just don't want to prematurely break 1.5 compatability (without this code the tooltips are unusable) if we ship minor updates before TS2/Debugger work is done.

devoncarew:
> so I'm gonna hold off merging this for now

No worries - no hurry to land :)

DanTup:
I've added the VS Code version to analytics so that it's easier to decide when to do things like this in future.

DanTup:
1.4?!

Code versionsode

DanTup:
Since I can edit your branch now (😁), I've added a change that marks this as requiring Code 1.6 so when this gets merged people won't see the update unless they have >= 1.6.

Will merge this either when a) I can see the huge majority of people are using 1.6 or b) I finish the debugger-targetting-terminal work which also requires it.

devoncarew:
Cool! A practical use of that feature :)

Seems like ~10% are on pre 1.6 versions?

DanTup:
Yeah, just under 10% of sessions for the last week. Annoyingly, I can't find a way to make a graph like the "App Versions" on in analytics (coloured stacked chart) so it's harder to see how quickly people are upgrading without re-running the report for different ranges.

DanTup:
Managed to make an equiv report in Google Data Studio. Didn't really show much with this little data though!

In time it might be handy though.. I should probably migrate the dashboard I created in Analytics to something in there; seems like it can do more :-)

Vode versions

DanTup:
Merged this now... There's the occasional 1.5 session but everything else is 1.6 and 1.7. Many branches are already requiring 1.6 anyway, so requiring 1.6 for the next release seems totally reasonable.

PR #201, opened 2016-11-01
Implement "debug just my code"

DanTup:


DanTup:
@devoncarew Interested in your thoughts on this. It seems to work as I'd like, but I don't know if there are edge cases I haven't covered. I'll add comments to the specific changes I think might be risky/fragile..

src/debug/debug_impl.ts
```diff
@@ -146,6 +146,20 @@ export class DartDebugSession extends DebugSession {
 							let thread = this.threadManager.getThreadInfoFromRef(isolateRef);
 							thread.receivedPauseStart();
 						}
+
+						// Helpers to categories libraries as SDK/ExternalLibrary/not.
+						let localPackageName = this.packageMap.getLocalPackageName();
+						let isSdkLibrary = (l: VMLibraryRef) => l.name.startsWith("dart.") || l.name == "builtin";
+						let isExternalLibrary = (l: VMLibraryRef) => !isSdkLibrary(l) && l.uri.startsWith("package:") && !l.uri.startsWith(`package:${localPackageName}/`);
```
DanTup: Is this logic sound? External packages always begin `package:` and won't be for the current packages name? I'm assuming you can only really be "one package" as a project?

src/debug/utils.ts
```diff
@@ -76,10 +77,18 @@ export class PackageMap {
 					this.map[name] = uriToFilePath(rest);
 				else
 					this.map[name] = path.join(path.dirname(file), rest);
+
+				// If we map to "lib/" then this must be the local package so we can stash the name.
+				if (rest == "lib/")
```
DanTup: Is this sound? It seems logical to me that the current package always maps to `lib/` (and it always seems to show up at the end of the `.packages` file) but I don't know if that's solid or the best way (I can't think of any reason this wouldn't all be solid, but, noob etc.!)

package.json
```diff
@@ -186,7 +196,7 @@
 				"program": "./out/src/debug/debug_entry.js",
 				"runtime": "node",
 				"variables": {
-					"sdkPath": "dart.getSdkPath"
+					"debugSettings": "dart.getDebugSettings"
```
devoncarew: Will existing launch configurations need to be updated?

src/debug/debug_impl.ts
devoncarew: Yup, looks good. I think you can remove the initial `!isSdkLibrary(l) && ` check.

src/debug/utils.ts
devoncarew: I think this works, but it is relying on an implementation detail (that could change).

README.md
```diff
@@ -42,6 +42,8 @@ Dart Code can be [installed from the Visual Studio Code Marketplace](https://mar
 
 - `dart.allowAnalytics`: Note: We only send a few very basic events and the platform and extension/Dart version numbers :-)
 - `dart.checkForSdkUpdates`: Whether to check you are using the latest version of the Dart SDK at startup. Defaults to `true`.
+- `dart.debugSdkLibraries`: Whether SDK libraries should be marked as debuggable. Defaults to `false`.
```
devoncarew: My 0.02 would be to default these to `true` (debuggable), as that would be more familiar to how I work. You milage my vary :)

src/debug/utils.ts
DanTup: I thought that at first, but then I figured that while the `.packages` file exists to provide a mapping from `packages:xxx` uris to actual locations, it'll always have to be there?

Unless you mean specifically being `lib/`.. I guess it could reasonably become a full path and then this would break.

Another option is reading it from `pubspec.yaml`; I just went with this as we already had read the data. Do you think that would be better? (though maybe I need to think about/test without a `pubspec.yaml`/`.packages` too, I presume it's totally valid to want to debug just a single .dart file in a folder?

package.json
DanTup: Yes :(

I haven't thoroughly tested this yet, but I think when you hit `F5` Code notices the values mismatch and tell you to fix it up (and open the file), though it's not very obvious. I don't know if there's any good way to handle this (besides calling out in release notes) since we can't hook `F5`.

Possibly at startup we could detect the badness and show an info tip with better information (and maybe a link to the release notes that describe it in detail)?

src/debug/debug_impl.ts
DanTup: I'll have to check. When I first did it, without this some things would appear as both SDK and External Libraries, and that meant they were altered by either option rather than just the corresponding one. It was a few weeks ago though (before I had a good isExternalLibrary implementation) so might be irrelevant now.

DanTup: > I think you can remove the initial `!isSdkLibrary(l) &&` check

I managed to do this - I change the `isSdk` function to check the `uri` for `dart:`. This seems more reliable anyway (I don't know if anything stops people from using `dart.` in their own library names, and I don't need to special-case `builtin` then).

package.json
DanTup: @devoncarew Managed to handle this (in a way that doesn't involve manually messing with that file).. In Code 1.6 we gained the ability to edit `launch.json` programatically. I've added a function for "project upgrades" which will run automatically at activation and fix this (added to this PR).

Unfortunately the formatting sucks (https://github.com/Microsoft/vscode/issues/15062), but even if it gets fixed in 1.8, I'm not sure it's worth waiting till that releases/people upgrade before shipping.

DanTup:
@devoncarew I think this is basically done. I know you're not keen on the default of debugging only users code but otherwise do you think this is now all sound? Do you want to try it out before I merge to `master` (or do you want me to just merge and you can try it out there?).

If you feel strongly, we could possibly prompt the user at startup with an info bar asking if they want to debug everything or just their code and update their settings for them? (I think we can use extension state to track that we already asked and only do it once)? I'm happy to add it if you think it'd be a better experience.

package.json
devoncarew: \o/ awesome! It will be nice that existing launch config keep working.

devoncarew:
> I know you're not keen on the default of debugging only users code but otherwise do you think this is now all sound?

Yup, code look goods! Re: the debugging change - thanks for keeping me in the loop. It's probably a personal preference thing, where you have to come done on one side of the fence or the other. You'll probably just want to give users a heads up, and find some way of gathering feedback. Some equivalent of a dart-code-announce list? Or just the gitter account. And I don't remember if you hooked the debugging pref up to analytics, but that would probably be a useful way to monitor what people are doing with the setting.

DanTup:
Nobody really uses Gitter (it's only mentioned in the contributing file), the Twitter account is probably the best place for announcements. Not sure if the release notes get read much (MS recently asked us to move them from README to CHANGELOG, so maybe they'll get better) but I planned to call it out there too.

Maybe I'll have a play with adding a question at startup to let the user easily pick which they want (an info bar with "debug all code" / "debug just my code" / "more info") and also add to analytics. If most people just "debug all code" then I'll make that the default in the next version. The popup could be removed on next version or left until we want to prompt the user for something else in future! I'll need to ensure we can reliably only ask once (I think it'll be fine with extension state; I used that in the past when we stashed the last known SDK path).

package.json
DanTup: Yeah; the error you get otherwise is really dopey. It says that `sdkPath` is unknown, but since the user never entered that it's not really clear what they should do! It's good this works, but shame about the formatting! Maybe they'll have an idea how to fix it, I'll wait for a response on that Issue before I merge this.

DanTup:
@devoncarew After some more feedback!

I've pushed another change that will prompt you when you open it (`Dart Code now supports debugging just your own code. Would you like to enable this?`). This will automatically update your user settings and set a flag so it doesn't prompt you again.

I've also added analytics that to record your setting on session start (`My code` / `All code` / `My code + SDK` / `My code + Libraries`) so we can see data (in the future I'd prefer to take this question back out and just have the default being the most used option).

I'll call this out in the release notes and maybe add a "More info" button to the prompt once I know the stable url for those release notes.

Let me know what you think (both of the idea, and whether the implementation looks good). If it all seems reasonable I think this is basically ready to ship :)

devoncarew:
> I've pushed another change that will prompt you when you open it

+1!

> I've also added analytics that to record your setting on session start

This will be great data to have -

Cool stuff! I will try the setting out myself, and see if I end up sticking with it :)

PR #211, opened 2016-11-12
offer to report analysis server errors

devoncarew:
When the analysis server reports an exception, offer to report it as a bug on dart-lang/sdk.

The report looks something like this:

```
Please report the following to https://github.com/dart-lang/sdk/issues/new:

Exception from analysis server (running from VSCode)

### what happened

<please describe what you were doing when this exception occurred>

### versions

- SDK 1.21.0-dev.3.0
- Visual Studio Code
- Dart Code 0.14.0-dev

### the exception

stack overflow 

'''
foo:123
bar:435
'''
```

### DIFF
--- a/src/analyzer_status_reporter.ts
+++ b/src/analyzer_status_reporter.ts
@@ -1,10 +1,20 @@
 "use strict";
 
-import { window, StatusBarItem, Disposable } from "vscode";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import { window, workspace, env, commands, extensions, StatusBarItem, Disposable, TextDocument } from "vscode";
 import { Analyzer } from "./analysis/analyzer";
 import { ServerStatusNotification, ServerErrorNotification } from "./analysis/analysis_server_types";
 import { config } from "./config";
-import { isDevelopment, logError } from "./utils";
+import { dartSdkRoot } from "./extension";
+import { getDartSdkVersion } from "./utils";
+
+const maxErrorReportCount = 3;
+
+let errorCount = 0;
+
+// TODO: We should show in the status line when the analysis server's process is dead.
 
 export class AnalyzerStatusReporter extends Disposable {
 	private statusBarItem: StatusBarItem;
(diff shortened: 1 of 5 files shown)


DanTup:
This looks great; I'll give it a test when I'm back home and got a little time.

I also made a note that we should log an analytic event when this happens so we can see if it's happening lots. We've found a couple of crashes lately (crashes on filenames like `con` on Windows and a crash when pasting in Python code to port to Dart) so it'd be reassuring to know these are edge cases and some people aren't just crashing out all the time!

src/analyzer_status_reporter.ts
```diff
@@ -38,8 +48,56 @@ export class AnalyzerStatusReporter extends Disposable {
...
+### the exception
+
+${error.message} ${error.isFatal ? ' (fatal)' : ''}
+
+\`\`\`
+${error.stackTrace}
+\`\`\`
+`;
+
+		let tempPath = path.join(os.tmpdir(), 'bug.md');
```
devoncarew: Yeah, good idea. It looks like `tmp` might help out here: https://www.npmjs.com/package/tmp#synchronous-file-creation-1

DanTup: Heh, this is one of the reasons I hate Node... That package has a dependency on another which claims it should be used over `os.tmpdir()` to work across platforms properly. So two packages to do something that Node should do natively! :/

Personal preference is if the code is trivial I'd rather just add it to our project than take deps (avoids many potential problems, inc. things like left-pad ;)) but if it's not, happy to take this as long as we review the code is sane and the issue trackers aren't loaded with ignored bugs ;(

devoncarew:
(updated to use `tmp`)

DanTup:
@devoncarew Do you know if a way I can reliably crash the AS to test this? I thought I would do it with some known bugs in old SDK versions, but I'm actually struggling to find any (I thought I'd reported some, but can't find them now!)

DanTup:
I managed to cause an error, but it wasn't caught by this code (it was a `SERVER_ERROR` but not an `event`). I've raised #213 about this.

I didn't manage to test this (struggled to generate a suitable error!) and ran out of time; but when I next get a few mins I'll try just throwing a fake error for it to handle to see it working (and verify on Windows). I'm happy to merge this without #213 though would be good to add it before doing a release (I'm happy to tackle the change, but might not be till Friday).

Thanks again!

devoncarew:
re: testing, I did have some code in that would mock an AS exception when you click on the `Analyzing...` status line item. I remove that before putting out this PR :)

I'll look at removing the `tmp` dep - we're not using much from it -

DanTup:
> I did have some code in that would mock an AS exception when you click on the Analyzing... status line item. I removed that before putting out this PR :)

That's a better idea than the hard-coding I was going to do (not for leaving in, but just for better testing), I'll do that!

> I'll look at removing the tmp dep - we're not using much from it -

I saw [this](https://nodejs.org/api/fs.html#fs_fs_mkdtempsync_prefix_options) mentioned somewhere before (it creates a folder, but could always just use `bug.md` appended or something) - seems like those packages are mostly handling a) old node version support (we don't care about this, we know our version from Electron an dit's pretty new) and b) delete the temp file on process end (I'm not sure how important that is).

DanTup:
Oh, I see you already removed it :)

Btw, Linux build is currently broken due to TS2 and our Travis Linux having an old version of node/npm (I think, anyway). Not sure if this is because of `language: dart`, just having a quick play to see if I can fix before I go, else might have to remain broken till next time!

PR #220, opened 2016-11-19
tweaks to the exception report

devoncarew:
Some tweaks to the exception report:

- trim extra whitespace
- make the subject line for the report a bit more terse
- remove an obsolete TODO:

Also, I noticed that the reported value for the dart sdk version was `null`, but did not see an obvious reason for it. The status line contribution for the sdk version was correct (non-null).

@DanTup

### DIFF
--- a/src/analyzer_status_reporter.ts
+++ b/src/analyzer_status_reporter.ts
@@ -97,7 +97,7 @@ Exception from analysis server (running from VSCode / Dart Code)
 - ${env.appName} ${codeVersion}
 - Dart Code ${dartCodeVersion}
 
-### Exception ${error.isFatal ? ' (Fatal)' : ''}
+### Exception${error.isFatal ? ' (fatal)' : ''}
 
 ${error.message}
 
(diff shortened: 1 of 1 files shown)


DanTup:
Hmmm, I'll do some testing of SDK version when next got time to work on this (I'll leave this open till then as a reminder) - I was sure all the version numbers were there when I added the Code version but could've been mistaken.

devoncarew:
Re: the null dart sdk - I'm not sure that there wasn't an issue with my sdk on disk; it may not be a reproducible issue.

DanTup:
I can't see any obvious reason it'd fail unless the `version` file was not readable (we have a `try`/`catch` there that returns `null`; probably it should `console.log` the exception!). I'll try to repro when at PC but suspect I won't be able to. I might tweak it so the version in `extension.ts` is shared and just re-use it rather than re-read from the file (seems pointless to read the SDK path from `extension` then use it to fetch the version that `extension` already knows).

DanTup:
I can't repro the null SDK issue (nor understand it from code), but I noticed you had another issue in the SDK project you thought might've been related to a bad SDK so I'm gonna assume you somehow didn't have the `version` file and ignore unless you can repro again!

PR #221, opened 2016-11-21
add some additional analysis server args

devoncarew:
- add support for `--instrumentation-log-file`
- add support for additional, arbitrary args to the analysis server

@DanTup

### DIFF
--- a/src/analysis/analyzer.ts
+++ b/src/analysis/analyzer.ts
@@ -25,20 +25,26 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 		let args = [];
 
 		// Optionally start Observatory for the analyzer.
-		let observatoryPort = config.analyzerObservatoryPort;
-		if (observatoryPort)
-			args.push(`--observe=${observatoryPort}`);
+		if (config.analyzerObservatoryPort)
+			args.push(`--observe=${config.analyzerObservatoryPort}`);
 
 		args.push(analyzerPath);
 
 		// Optionally start the analyzer's diagnostic web server on the given port.
-		let diagnosticsPort = config.analyzerDiagnosticsPort;
-		if (diagnosticsPort)
-			args.push(`--port=${diagnosticsPort}`);
+		if (config.analyzerDiagnosticsPort)
+			args.push(`--port=${config.analyzerDiagnosticsPort}`);
 
 		// Add info about the extension that will be collected for crash reports etc.
 		args.push(`--client-id=DanTup.dart-code`);
 		args.push(`--client-version=${extensionVersion}`);
+
+		// The analysis server supports a verbose instrumentation log file.
+		if (config.analyzerInstrumentationLogFile)
+			args.push(`--instrumentation-log-file=${config.analyzerInstrumentationLogFile}`);
+
+		// Allow arbitrary args to be passed to the analysis server.
+		if (config.analyzerAdditionalArgs)
+			args.concat(config.analyzerAdditionalArgs);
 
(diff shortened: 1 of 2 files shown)


DanTup:
Out of curiosity, are there any details on this flag (or other flags you might expect to be added)?

src/analysis/analyzer.ts
```diff
@@ -25,21 +25,27 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
...
+		if (config.analyzerDiagnosticsPort)
+			args.push(`--port=${config.analyzerDiagnosticsPort}`);
 
 		// Add info about the extension that will be collected for crash reports etc.
 		args.push(`--client-id=DanTup.dart-code`);
 		args.push(`--client-version=${extensionVersion}`);
 
+		// The analysis server supports a verbose instrumentation log file.
+		if (config.analyzerInstrumentationLogFile)
+			args.push(`--instrumentation-log-file=${config.analyzerInstrumentationLogFile}`);
```
devoncarew: I don't know - do we escape the dart sdk path config value? We should probably be consistent.

devoncarew:
There aren't docs on `--instrumentation-log-file`, but that could be added (hereabouts: https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/spec/spec_input.html#L141).

The other flags to change a bit, and are often used by us for experimental things that then roll into the analysis server by default, w/o a flag. Currently we want to be able to enable `--enable-new-analysis-driver` - used to toggle on some performance work. Once we have confidence in that work, the flag will go away (which is why I didn't add a specific config option in vscode for it).

src/analysis/analyzer.ts
DanTup: The path of the VM gets passed to `child_process.spawn` as its own arg, so it didn't need it (I think we tested it when #106 was done). I don't think any of the other args here have been strings before (at least, not if you enter them correctly) so it hasn't come up.

I can do some testing of this when I get some time at a PC (I really ought to give Linux+Code a go on my Chromebook, would get many more opportunities to spend time on this!); will leave this open as a reminder, but otherwise it looks fine to me!

DanTup:
> There aren't docs on `--instrumentation-log-file`, but that could be added

It might not be worth it; I guess it depends if it's generally useful to IDE integrators or just mainly useful to you as a Dart team dev (I don't know, haven't seen the output yet - but no doubt I'll have a poke around!)

> Currently we want to be able to enable `--enable-new-analysis-driver` - used to toggle on some performance work

I did wonder what that was - Brian mentioned it in an issue I raised!

> Once we have confidence in that work, the flag will go away (which is why I didn't add a specific config option in vscode for it)

I did wonder where we'd draw a line with options vs using that; this makes perfect sense! :)

devoncarew:
Sorry - forgot to say what it does; `--instrumentation-log-file` logs traffic to and from the analysis server to a file, and includes some AS specific logging as well. It's used to dump lots of info that could then be useful later for debugging AS issues.

DanTup:
I had a quick peek - I wondered if it could be a replacement for our logging; but it seems a lot more verbose. Although it'd be good to have that additional info for troubleshooting; the simpler log might make be easier to deal with unless it's needed.

All good; shipping this in a version shortly.

devoncarew:
> shipping this in a version shortly.

👍

src/analysis/analyzer.ts
DanTup: Not sure why, but it actually worked fine with the space (writes to the file with a space in it) :-)

PR #225, opened 2016-11-28
fix a bug with passing additional args to the analysis server

devoncarew:
@DanTup, fix a bug with passing additional args to the analysis server - the args weren't being passed to process creation.

Also, filter some analysis server logging from the json stream.


DanTup:
@devoncarew Haven't had chance to do much on DC lately, so I've pushed out v0.14.5 with the few little tweaks made recently (this fix and the error report tweaks).

PR #227, opened 2016-11-30
record the request we were responding to

devoncarew:
When reporting a server error received from a request, record the server method that we'd called.

```
### Request

While responding to request: `foo.bar`
```

Also, fix an issue where when we report the server error, we didn't complete the handler (with the error).

### DIFF
--- a/src/analysis/analyzer.ts
+++ b/src/analysis/analyzer.ts
@@ -11,7 +11,7 @@ import { log, logError, extensionVersion } from "../utils";
 export class Analyzer extends AnalyzerGen implements vs.Disposable {
 	private analyzerProcess: child_process.ChildProcess;
 	private nextRequestID = 1;
-	private activeRequests: { [key: string]: [(result: any) => void, (error: any) => void] } = {};
+	private activeRequests: { [key: string]: [(result: any) => void, (error: any) => void, string] } = {};
 	private messageBuffer: string[] = [];
 	private logStream: fs.WriteStream;
 
@@ -89,7 +89,7 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 			// This will include things like Observatory output and some analyzer logging code.
 			message = message.trim();
 			if (!message.startsWith('--- ') && !message.startsWith('+++ ')) {
-				logError({ message: `Unable to parse message (${e}): ${message}` });
+				console.error(`Unable to parse message (${e}): ${message}`);
 			}
 			return;
 		}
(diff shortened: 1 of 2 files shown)


src/analysis/analyzer.ts
```diff
@@ -89,7 +89,7 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 			// This will include things like Observatory output and some analyzer logging code.
 			message = message.trim();
 			if (!message.startsWith('--- ') && !message.startsWith('+++ ')) {
-				logError({ message: `Unable to parse message (${e}): ${message}` });
+				console.error(`Unable to parse message (${e}): ${message}`);
```
DanTup: How come you changed this? The `logError` function shows the error in  a toast notification when running in the extension host which I found handy to make sure I definitely notice if it goes bad (normal users wouldn't see it)!

src/analysis/analyzer.ts
```diff
@@ -136,12 +136,18 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 
 	private handleResponse(evt: UnknownResponse) {
 		let handler = this.activeRequests[evt.id];
-		if (evt.error && evt.error.code == "SERVER_ERROR")
+		let method: string = handler[2];
+
+		if (evt.error && evt.error.code == "SERVER_ERROR") {
+			evt.error['method'] = method;
 			this.notify(this.requestErrorSubscriptions, <as.RequestError>evt.error);
-		else if (evt.error)
```
DanTup: Whoops, good catch!

src/analysis/analyzer.ts
DanTup: Ah, ok, I haven't used that. Is there a better way to separate AS responses we care about and that sort of output? Feels a bit hacky to be attempting `JSON.parse` on every messages and assuming those that work are for us and those that don't are not?

devoncarew: The logging output from the driver didn't have any consistent markers we could use. It may be enough to check for .startsWith('{') and .endsWith('}'), and then assume that any parse errors are worth reporting?

DanTup: Checking the end for `}` would miss anything that might've been truncated (or our collecting of chunks gone bad?), but maybe worth doing startswith `{`. That said, I don't know how likely it is we'd get bad JSON back; I suspect the serialisation doesn't change much and that it'd be noticed well before a release if it was broken!

Issue #224, opened 2016-11-26
add additional info about the analysis server to bug reports


devoncarew:
We can add a bit more info to bug reports - the analysis server has a diagnostic call:

https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html#domain_diagnostic

That can return info like the number of analysis context, the size of each context, and some other metadata.

DanTup:
I saw that a few days ago (possibly following one of your commits on sdk!) and made a note to have a look at the data that came back (hasn't happened yet though!).

I presume it'll only work after non-fatal errors though; but would've been useful on fatal ones! ;(

If it's cheap to call, we could possibly call it periodically so if a fatal error does occur, we could at least have some "fairly recent" stats?

devoncarew:
Even on fatal errors the analysis server is often still around to query (we should be defensive about the call not returning however).

devoncarew:
We should probably update the spec docs re the behavior on fatal errors.

PR #228, opened 2016-12-02
Include diagnostics in error report

DanTup:


DanTup:
@devoncarew Can you check if you think this is a good implementation for #224? It does mean we're including paths (in the contexts), though I hope users will scan reports rather than blindly sending.

I've only tested this with a non-fatal error; I'm not sure how I can reliably simulate a fatal error (I was never able to get the analyzer working from my own source so I can't mess inside there to force one!)

DanTup:
Oops, and here's what the extra info looks like:

---

### Analyzer Info

The analyzer was launched with the arguments:

```text
--client-id=DanTup.dart-code
--client-version=0.15.0-dev
```

Diagnostics requested after the error occurred are:

```js
[
    {
        "name": "m:\\Coding\\TestStuff\\FootballGameDart-bad",
        "explicitFileCount": 20,
        "implicitFileCount": 4,
        "workItemQueueLength": 0,
        "cacheEntryExceptions": []
    }
]
```

devoncarew:
Looks great!

> It does mean we're including paths (in the contexts), though I hope users will scan reports rather than blindly sending.

Perhaps we should add a note near the beginning of the bug report asked the user to double-check for sensitive information?

DanTup:
@devoncarew Added a note in https://github.com/Dart-Code/Dart-Code/pull/228/commits/b093a855a4ddf69970b0b04917615606aae1b84d

src/analysis/analyzer.ts
```diff
@@ -61,6 +64,10 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 		this.serverSetSubscriptions({
 			subscriptions: ["STATUS"]
 		});
+
+		// Hook error subscriptions so we can try and get diagnostic info if this happens.
+		this.registerForServerError(e => this.requestDiagnosticsUpdate());
```
devoncarew: Just wanted to say, this is a pretty cool way to collect diagnostics. I'd taken a stab at a PR like this when I filed the related issue, but got hung up on trying to get a `timeout()` method working in typescript, and eventually abandoned the PR.

DanTup: Yeah, I was trying to come up with a good way - this seemed reasonable, and we could also ping every few mins if they tend to not come through after a crash to try and get some "recent" data instead (though worth seeing what the data is like first).

There's been a *lot* of crashes/errors in SDK 1.20, I'm hoping they're mostly all related to the issue I raised that was fixed and 1.21 will be much better!

PR #251, opened 2017-01-26
Update to TypeScript 2.1.5

### LINKED ISSUE #232: Dart Code doesn't compile under TypeScript 2.1
See https://github.com/Microsoft/TypeScript/issues/12794

Adding casts probably fixes this, but it just feels like a workaround for a TypeScript bug so I'd like to see a sensible response from TypeScript first.

sharwell:
Fixes #232

### DIFF
--- a/src/user_config_prompts.ts
+++ b/src/user_config_prompts.ts
@@ -14,7 +14,7 @@ function prompt(context: vs.ExtensionContext, key: string, prompt: () => Thenabl
 	//context.globalState.update(stateKey, undefined);
 
 	// If we've not prompted the user with this question before...
-	if (context.globalState.get(stateKey, false) !== true) {
+	if (context.globalState.get(stateKey) !== true) {
 		// Prompt, but only record if the user responded.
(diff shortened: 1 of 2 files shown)


src/user_config_prompts.ts
```diff
@@ -14,7 +14,7 @@ function prompt(context: vs.ExtensionContext, key: string, prompt: () => Thenabl
 	//context.globalState.update(stateKey, undefined);
 
 	// If we've not prompted the user with this question before...
-	if (context.globalState.get(stateKey, false) !== true) {
+	if (context.globalState.get(stateKey) !== true) {
```
sharwell: :memo: The default value was unnecessary here, because `undefined` and `false` have the same result when compared strictly to `true`.

DanTup: Thanks! I kept meaning to come back to this. It seems TS is sticking with this and casting a literal `false` to boolean felt dirty to me - I'm much happier with this :D

DanTup:
@sharwell Thanks!

It's been a while!

Issue #187, opened 2016-09-23
Enable strict mode for TypeScript


DanTup:
- [x] alwaysStrict
- [x] noImplicitAny
- [ ] noImplicitReturns
- [x] noImplicitThis
- [ ] noUnusedLocals
- [ ] noUnusedParameters
- [x] strictNullChecks
- [x] strictFunctionTypes
- [x] strictPropertyInitialization.
- [x] strict (enables several of the above)

DanTup:
I tried to turn on `strictNullChecks` and it causes a lot of errors. I'm also a little unhappy with how it works, for ex. it lets me have a field that claims not to be null:

```ts
private logStream: fs.WriteStream;
```

Yet the value is null by default! When I try to assign null, it complains. Seems inconsistent - I'd rather it rejected the above declaration ;(

PR #252, opened 2017-01-26
Set noImplicitAny to true and fix all errors

sharwell:
⚠️ I am expecting changes will need to be made. Please review the approach I used in each situation carefully to determine if it's what you really want to go with.

Related to #187

### DIFF
--- a/src/analysis/analyzer.ts
+++ b/src/analysis/analyzer.ts
@@ -144,14 +144,15 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
 	private handleResponse(evt: UnknownResponse) {
 		let handler = this.activeRequests[evt.id];
 		let method: string = handler[2];
+		let error: as.RequestError & { method?: string } = evt.error;
 
-		if (evt.error && evt.error.code == "SERVER_ERROR") {
-			evt.error['method'] = method;
-			this.notify(this.requestErrorSubscriptions, <as.RequestError>evt.error);
+		if (error && error.code == "SERVER_ERROR") {
+			error.method = method;
+			this.notify(this.requestErrorSubscriptions, error);
 		}
 
-		if (evt.error) {
-			handler[1](evt.error);
+		if (error) {
+			handler[1](error);
 		} else {
 			handler[0](evt.result);
 		}
--- a/src/analysis/analyzer_gen.ts
+++ b/src/analysis/analyzer_gen.ts
@@ -8,7 +8,7 @@ import * as as from "./analysis_server_types";
 import * as a from "./analyzer";
 
 export abstract class AnalyzerGen {
-	protected abstract notify<T>(subscriptions: ((notification: T) => void)[], params: T);
+	protected abstract notify<T>(subscriptions: ((notification: T) => void)[], params: T): void;
 	protected abstract subscribe<T>(subscriptions: ((notification: T) => void)[], subscriber: (notification: T) => void): vs.Disposable;
 	protected abstract sendRequest<TReq, TResp>(method: string, params?: TReq): Thenable<TResp>;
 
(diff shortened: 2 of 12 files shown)


src/analysis/analyzer.ts
```diff
@@ -144,14 +144,15 @@ export class Analyzer extends AnalyzerGen implements vs.Disposable {
...
 		let handler = this.activeRequests[evt.id];
 		let method: string = handler[2];
+		let error: as.RequestError & { method?: string } = evt.error;
 
-		if (evt.error && evt.error.code == "SERVER_ERROR") {
-			evt.error['method'] = method;
-			this.notify(this.requestErrorSubscriptions, <as.RequestError>evt.error);
+		if (error && error.code == "SERVER_ERROR") {
+			error.method = method;
+			this.notify(this.requestErrorSubscriptions, error);
```
sharwell: :memo: This cast was unnecessary prior to this pull request.

src/analysis/analyzer_gen.ts
```diff
@@ -8,7 +8,7 @@ import * as as from "./analysis_server_types";
 import * as a from "./analyzer";
 
 export abstract class AnalyzerGen {
-	protected abstract notify<T>(subscriptions: ((notification: T) => void)[], params: T);
+	protected abstract notify<T>(subscriptions: ((notification: T) => void)[], params: T): void;
```
sharwell: :memo: This appears to be in generated code, but I don't know how to update the original source.

src/debug/debug_impl.ts
```diff
@@ -860,7 +860,7 @@ class ThreadManager {
 
 	removeStoredIds(ids: number[]) {
 		for (let id of ids) {
-			delete this.storeData[id];
```
sharwell: :memo: I'm pretty sure this was an actual bug (`storeData` vs. `storedData`).

DanTup:
The file `analyzer_gen.ts` is generated from the API docs (https://github.com/Dart-Code/analysis-server-typescript-generator) so shouldn't be manually edited. I'll see if I can get some to go through it this weekend and update it (as well as review all these). Thanks!

src/debug/debug_impl.ts
DanTup: That does look weird, I suspect you're right. I'm not totally familiar with everything in this file as @devoncarew contributed it and it's rather complicated to debug (the irony!). I think Code has incoming changes that might make that easier (debugger attached to multiple processes - the extension and the debugger).

src/analysis/analyzer_gen.ts
DanTup: heh, yep, I commented about that just after. It's a dart script which takes the [API Doc](https://htmlpreview.github.io/?https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/doc/api.html) and generates all the TS types from it. I don't know if there's enough info for us to do this totally accurately (eg. know what can/can't be null) but certainly changes like this are simple :-)

src/debug/debug_impl.ts
devoncarew: Yup, looks like. I'm guessing this would have manifested mostly as a memory leak?

DanTup: I suspect so; though how much of how often, I don't know. I'll try and get this merged soon anyway; I've wanted to use these TS options for a while but never got round to completely getting it compiling again!

DanTup:
Actually, the project I linked didn't generate that file at all; I never finished porting it (it only does the types file); it's currently an old crufty C# project not shared! I'll fix it up for now and add porting it to the dart project to my list!

DanTup:
All seems to work fine to me; thanks! 👍

PR #288, opened 2017-05-12
Added ability for findDartSdk to use the Flutter cached version.

### LINKED ISSUE #286: Dart SDK should be used from local cache folder for Flutter projects
Flutter bundles its own Dart SDK. When I open a Flutter project, I see this error:

capture

Ideally:

* VS Code should detect I've opened a Flutter project (probably from pubspec.yaml containing "sdk: flutter"), and then find the Dart SDK that is downloaded by Flutter (inside of FLUTTER_HOME\bin\cache\dart-sdk)

Suggesting that the user "Go to Dart downloads" isn't correct for Flutter. A Flutter user wants the Dart SDK that is in Flutter, not the SDK that is downloaded from dartlang.org

BTW thanks for this plugin. Really nice!

Skylled:
Fixes #286
Also fixed a minor annoyance of Dart-Code offering to update the Dart SDK when Flutter rolls their own.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -28,7 +28,7 @@ export function findDartSdk(): string {
 	let paths = (<string>process.env.PATH).split(path.delimiter);
 
 	// Putting Flutter before userDefined, so that UD takes priority
-	if (isFlutterProject)
+	if (isFlutterProject())
 		paths.unshift(path.join(findFlutterSdk(), "bin/cache/dart-sdk/bin"));
 
 	// We don't expect the user to add .\bin in config, but it would be in the PATHs
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -10,14 +10,26 @@ import { config } from "./config";
 const isWin = /^win/.test(process.platform);
 const dartExecutableName = isWin ? "dart.exe" : "dart";
 const pubExecutableName = isWin ? "pub.bat" : "pub";
+const flutterExecutableName = isWin ? "flutter.bat" : "flutter";
 export const dartVMPath = "bin/" + dartExecutableName;
 export const dartPubPath = "bin/" + pubExecutableName;
 export const analyzerPath = "bin/snapshots/analysis_server.dart.snapshot";
 export const extensionVersion = getExtensionVersion();
 export const isDevelopment = checkIsDevelopment();
 
+export function isFlutterProject(): boolean {
+	if (workspace.rootPath)
+		if (fs.existsSync(path.join(workspace.rootPath, "pubspec.yaml")))
+			return fs.readFileSync((path.join(workspace.rootPath, "pubspec.yaml"))).includes("sdk: flutter");
+	return false;
+}
+
 export function findDartSdk(): string {
 	let paths = (<string>process.env.PATH).split(path.delimiter);
+
+	// Putting Flutter before userDefined, so that UD takes priority
+	if (isFlutterProject)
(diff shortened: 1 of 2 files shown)


Skylled:
There's a mistake in this, hang on.

src/utils.ts
```diff
@@ -10,15 +10,27 @@ import { config } from "./config";
...
 export const dartVMPath = "bin/" + dartExecutableName;
 export const dartPubPath = "bin/" + pubExecutableName;
 export const analyzerPath = "bin/snapshots/analysis_server.dart.snapshot";
 export const extensionVersion = getExtensionVersion();
 export const isDevelopment = checkIsDevelopment();
 
+export function isFlutterProject(): boolean {
+	if (workspace.rootPath)
+		if (fs.existsSync(path.join(workspace.rootPath, "pubspec.yaml")))
+			return fs.readFileSync((path.join(workspace.rootPath, "pubspec.yaml"))).includes("sdk: flutter");
```
DanTup: I think it would be slightly better to use a regex here to be flexible about the whitespace between `sdk:` and `flutter` (maybe even allow option whitespace before the colon; not sure how struct YAML is).

Skylled:
Double-checked everything, looks okay. Tripped myself, thinking there was a problem, but it's all working.

src/utils.ts
```diff
@@ -10,15 +10,27 @@ import { config } from "./config";
...
+	return false;
+}
+
 export function findDartSdk(): string {
 	let paths = (<string>process.env.PATH).split(path.delimiter);
 
+	// Putting Flutter before userDefined, so that UD takes priority
+	if (isFlutterProject)
```
DanTup: I'm not sure (I've never used Flutter), but I think it might be invalid/incorrect to use the "normal" Dart SDK for Flutter projects. If so, it might be best to export a new `findSdk` function which internally just calls `findDartSdk` or `findFlutterSdk` accordingly. This would also make it easier in the flutter case if `findFlutterSdk` doesn't find it to alert the user to do whatever is required to pull the flutter-dart-sdk down.

src/utils.ts
```diff
@@ -47,6 +59,28 @@ function hasDartExecutable(pathToTest: string): boolean {
...
+	let flutterPath = paths.find(hasFlutterExecutable);
+	if (!flutterPath)
+		return null;
+	
+	let realFlutterPath = fs.realpathSync(path.join(flutterPath, flutterExecutableName));
+
+	return path.join(path.dirname(realFlutterPath), "..");
+}
+
+function hasFlutterExecutable(pathToTest: string): boolean{
```
DanTup: Very minor but since this whole function is a copy/paste with just the filename changed, I think it'd be better to rename `hasExecutable` and just pass the full path/filename in.

DanTup:
Not sure why the CI builds failed; I'll investigate when I merge/test this. There aren't really any useful tests and they're a bit fragile currently.

src/utils.ts
Skylled: Honestly, my JS/TS experience is minimal, so I don't really know how to do that, while still using `paths.find`.

DanTup: No worries, I'll merge like this and tidy up after testing. Thanks!

DanTup:
I messed up a little merging these in (due to conflicts), but it's all resolved now. Both PRs are merged into master; I'll test/tweak an include in the next release. Thanks again!

Issue #308, opened 2017-05-21
Syntax highlighter should allow $ in identifiers. 


rkirsling:
It appears that the syntax highlighter isn't expecting `$` to occur within identifiers, so for instance, `$return` will have keyword highlighting applied (_while, say, `this$` and `$1$2$` use highlighting for distinguished variables and numeric literals, respectively_).

Stopping this extra highlighting is easy enough by qualifying `\b`s with negative lookaround for `\$`, similar to VSCode's [JavaScript grammar](https://github.com/Microsoft/vscode/blob/master/extensions/javascript/syntaxes/JavaScript.tmLanguage.json#L410).

The harder part is ensuring that `$` doesn't stop desired highlighting from occurring. The problem is the grammar here seems to be 10x shorter than a lot of other languages—all we really have are some [naive definitions for class / function /  constant identifiers](https://github.com/Dart-Code/Dart-Code/blob/master/syntaxes/dart.json#L194-L209), so I'm at a loss for the "right" thing to do.

I can submit a PR with my suggestion: to remove the constant definition (_which is actually bogus_ 😛) and loosen the other two (_which should get us pretty close to correct, anyway_).

Skylled:
Can you show a screenshot of this happening in valid code? I'm curious.

DanTup:
I think we took this file from Atom, so @devoncarew might be interested in this too (though admittedly I haven't checked if there are updates there I haven't taken).

If https://github.com/Microsoft/vscode/issues/1967 ever gets implemented, we could switch to using the highlighting provided by the analyzer, but it's not looking likely.

@rkirsling I'll test out the PR when next working on this (might be a couple of weeks; got lots on next weekend) to see how it behaves. Thanks! :-)

rkirsling:
@Skylled: Here's some of the code that inspired this:
![image](https://cloud.githubusercontent.com/assets/2172590/26287170/4eeea562-3e2a-11e7-92f4-c9fe95ae005a.png)

@DanTup: Having an analyzer like that would be nice indeed!
Either way, no hurry—I debated whether to submit it here or to [dart-atom/dartlang](https://github.com/dart-atom/dartlang/) first, but the maintenance here seemed to be more active. :smile:

DanTup:
I wondered why you'd be naming variables like this; but makes sense now!

Strange that the lowercase/uppercase a/z are different too!

I will test and merge when I get some free time at PC at home. Ta!

DanTup:
[...]

PR #309, opened 2017-05-21
Syntax Highlighting: Allow $ in identifiers.

rkirsling:
Addresses #308.

This is the best solution I could come up with, short of making massive additions to the grammar. :smile:

### DIFF
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -180,67 +180,63 @@
 		"constants-and-special-vars": {
 			"patterns": [
 				{
-					"match": "\\b(true|false|null)\\b",
+					"match": "(?<!\\$)\\b(true|false|null)\\b(?!\\$)",
 					"name": "constant.language.dart"
 				},
 				{
-					"match": "\\b(this|super)\\b",
+					"match": "(?<!\\$)\\b(this|super)\\b(?!\\$)",
 					"name": "variable.language.dart"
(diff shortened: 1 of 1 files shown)


syntaxes/dart.json
```diff
@@ -180,67 +180,63 @@
...
-					"match": "(_?[a-z][a-zA-Z0-9_]+)(\\(|\\s+=>)",
+					"match": "([_$]*[a-z][a-zA-Z0-9_$]*)(\\(|\\s+=>)",
 					"captures": {
 						"1": {
 							"name": "entity.name.function.dart"
 						}
 					}
-				},
-				{
-					"match": "\\b[A-Z_0-9]+\\b",
```
rkirsling: Sorry, I should've made this explicit! This regex intended to detect and highlight constants as a separate sort of entity, but it's wrong for two reasons:

1. It's subsumed by the regex for classes, so it would actually only cause single capital letters to be highlighted differently (_although, as you noticed, depending on your color scheme selection in VSCode this may not have any effect at all_).

2. `SCREAMING_CAPS` isn't an appropriate way to look for constants in Dart anyway, seeing as [Effective Dart](https://www.dartlang.org/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names) explicitly recommends `lowerCamelCase`.

PR #344, opened 2017-07-03
fix a tsc type error

devoncarew:
I think this is a place where the tsc compiler's inference was falling down, but in any case, this should make the bots happier.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -857,7 +857,7 @@ class ThreadManager {
 		if (promise)
 			return promise;
 
-		let completer = new PromiseCompleter();
+		let completer = new PromiseCompleter<boolean[]>();
 		let result = [];
 		for (let i = 0; i < breakpoints.length; i++) {
 			result.push(true);


DanTup:
Actually, I do remember looking at this! I believe it's a new typescript (and I think this behaviour is "better", since it was previously a non-generic type, there was nothing to stop pushing non-bools as results). I was confused about why the tests were suddenly getting new typescript and I wasn't locally but never got to investigate. Maybe cleaning out `node_modules` and running `npm install` locally would've done the same thing.

Thanks!

devoncarew:
Hmm, well travis is happier, but appveyor still fails...

DanTup:
```text
Failed to execute tests: Error: spawn C:\projects\dart-code\.vscode-test\Code ENOENT
Tests exited with code: -4058
Command exited with code 1
```

Very useful ;(

The tests are useless right now anyway - I keep wanting to get back to them but with limited time it's been more useful to work on things like Flutter than try and figure out how to easily test all this. I'll take a look at if I can get better logs sometime (for some reason it's running `npm test --silent` which probably isn't most useful).

DanTup:
Turns out to be an issue with the Code build from MS; it'll fix itself with the next Code release (there's a workaround but since the new release isn't far off I'm avoiding having to undo it again!). Details @ https://github.com/Microsoft/vscode-extension-vscode/issues/64

Issue #348, opened 2017-07-04
show dart packages dependencies in the explorer view


devoncarew:
I think this could be done by parsing the .packages file, listening for changes, and contributing a new explorer view and TreeDataProvider:

```
		"views": {
			"explorer": [
				{
					"id": "dartPackages",
					"name": "Dart Packages",
					"when": "resourceLangId == 'dart'"
				}
			]
		},
```

DanTup:
I've merged this into a branch and changed the context to always when a Dart project is loaded. I agree that opening the packages as editable sucks, so I've raised a request code Code to support readonly files: https://github.com/Microsoft/vscode/issues/31410

devoncarew:
👍 cool!

DanTup:
Looks like I misunderstood, and the text content providers aren't required to be HTML (the sample code I saw was relying on Code being clever and guessing content type based on the first line, which was an HTML tag).

Should be trivial to switch this over, and then maybe it'll be reasonable to ship.

DanTup:
@devoncarew Made some tweaks to this (inc. readonly) and shipped in v2.3.2. Thanks!

PR #365, opened 2017-07-15
add a Dart Packages view

devoncarew:
Most of the implementation for a Dart Packages view (contribution to the Explorer) - work towards https://github.com/Dart-Code/Dart-Code/issues/348.

After using this for a while, I'm not 100% happy with how it works. The contribution shows and hides as you select dart and non-dart files, the expansion state of the view items are not remembered, and the packages files open in a writeable state (which is risky - users could easily edit files in their pub_cache).

I don't recommend landing this, but putting the PR out there for posterity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,41 @@
-# v2.0 - Flutter Run/Debug/Reload Support!
+# v2.3.1
+
+- An issue with searching symbols in a file being unreliable (appearing to hang) has been fixed
+- A new setting `flutterHotReloadOnSave` has been added to force a hot reload upon saving (if there are no errors in the file being saved)
+- A new `Full Restart` command is now available during Flutter debugging sessions for a full restart
+
+Additionally, a new prototype editor feature is available for Flutter projects to improve readability of `build` methods, see #383 for details.
+
+# v2.3.0 - Flutter Debugging Commands
+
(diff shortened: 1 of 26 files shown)


DanTup:
>  The contribution shows and hides as you select dart and non-dart files

Why only show when a Dart file is selected and not just always when a Dart/Flutter project is loaded?

>  the packages files open in a writeable state

Maybe fixable with [TextDocumentContentProvider](https://code.visualstudio.com/docs/extensionAPI/vscode-api#_a-nametextdocumentcontentprovideraspan-classcodeitem-id457textdocumentcontentproviderspan)? Not sure how well it'd work or how much effort it'd be though.

I'll see if I can find any other extensions doing this (I'd expect C# will have/want this) and how they compare.

devoncarew:
> Why only show when a Dart file is selected and not just always when a Dart/Flutter project is loaded?

If you can scope it to that, sounds great.

DanTup:
Unless I've overlooked something; it should be as simple as changing `"when": "resourceLangId == 'dart'"` to `"when": "dart-code:dartProjectLoaded"` which is a custom context we set when loading a dart project (flutter gets this too, though also has its own - if you need dart projects that aren't flutter you can use `"when": "dart-code:dartProjectLoaded && !dart-code:flutterProjectLoaded"`).

These were added during Flutter dev, so we can hide `pub get` etc. in favour of the Flutter equivs :-)

PR #393, opened 2017-07-28
Support angular template analysis

LorenVS:
If you have the angular_analyzer_plugin [1] active, the dart analysis server supports analysis and code completion for AngularDart templates. This PR surfaces this information by registering the same dart providers for html files. This functionality is gated behind a new "analyzeAngularTemplates" option (default: false).

[1] https://github.com/dart-lang/angular_analyzer_plugin

### DIFF
--- a/package.json
+++ b/package.json
@@ -391,11 +391,6 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Whether to automatically send a 'hot reload' request during a Flutter debug session when saving files."
-				},
-				"dart.analyzeAngularTemplates": {
-					"type": "boolean",
-					"default": false,
-					"description": "Whether to analyze HTML files as AngularDart templates"
 				}
 			}
 		},
--- a/package.json
+++ b/package.json
@@ -391,6 +391,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Whether to automatically send a 'hot reload' request during a Flutter debug session when saving files."
+				},
+				"dart.analyzeAngularTemplates": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to analyze HTML files as AngularDart templates"
 				}
 			}
 		},
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -35,6 +35,7 @@ import { FlutterWidgetConstructorDecoratorProvider } from "./providers/flutter_w
 import { DartPackageFileContentProvider } from "./providers/dart_package_file_content_provider";
 
 const DART_MODE: vs.DocumentFilter = { language: "dart", scheme: "file" };
+const HTML_MODE: vs.DocumentFilter = { language: "html", scheme: "file" };
 const DART_DOWNLOAD_URL = "https://www.dartlang.org/install";
 const FLUTTER_DOWNLOAD_URL = "https://flutter.io/setup/";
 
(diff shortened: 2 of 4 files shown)


DanTup:
Awesome; thanks! I'll try and test it out this weekend :-)

package.json
```diff
@@ -391,6 +391,11 @@
 					"type": "boolean",
 					"default": false,
 					"description": "Whether to automatically send a 'hot reload' request during a Flutter debug session when saving files."
+				},
+				"dart.analyzeAngularTemplates": {
```
DanTup: We should probably add this to `getAnalyzerSettings` [here](https://github.com/Dart-Code/Dart-Code/blob/c033ff08ab52cc4da9f67bcb6cb34609db624cd6/src/extension.ts#L368) so that changes to it prompt the user to reload (otherwise the change will appear to not take effect).

package.json
```diff
@@ -391,6 +391,11 @@
...
+					"type": "boolean",
+					"default": false,
```
DanTup: If this functionality works well, maybe it'd be better to default it to on?

package.json
```diff
@@ -391,6 +391,11 @@
...
+					"description": "Whether to analyze HTML files as AngularDart templates"
```
DanTup: Should also add this description to the README which lists all of the options.

src/extension.ts
```diff
@@ -137,17 +138,44 @@ export function activate(context: vs.ExtensionContext) {
...
+	context.subscriptions.push(vs.languages.registerCompletionItemProvider(DART_MODE, completionItemProvider, ".", ":", " ", "=", "("));
+	context.subscriptions.push(vs.languages.registerDefinitionProvider(DART_MODE, definitionProvider));
+	context.subscriptions.push(vs.languages.registerDocumentSymbolProvider(DART_MODE, documentSymbolProvider));
+	context.subscriptions.push(vs.languages.registerReferenceProvider(DART_MODE, referenceProvider));
+	context.subscriptions.push(vs.languages.registerDocumentHighlightProvider(DART_MODE, documentHighlightProvider));
+	context.subscriptions.push(vs.languages.registerCodeActionsProvider(DART_MODE, codeActionProvider));
+	context.subscriptions.push(vs.languages.registerRenameProvider(DART_MODE, renameProvider));
+
+	// Analyze Angular2 templates, requires the angular_analyzer_plugin.
+	if (config.analyzeAngularTemplates) {
```
DanTup: I'd probably put `DART_MODE` and `HTML_MODE` into an array and `.forEach` to avoid duplicating all of these. It seems `registerWorkspaceSymbolProvider` is the only analyzer-backed provided that's not added; is that something the Angular plugin doesn't support (I don't know if it makes sense logically, I've not used Angular seriously).

package.json
DanTup: I just re-read this and realised this plug isn't in the SDK, you need to built it yourself. I'm wondering if for now this should be a hidden option so it doesn't show up in the settings completion - otherwise I worry people will turn it on expecting it to work and it won't! I've started listing "Preview Features" in the [release notes](https://github.com/Dart-Code/Dart-Code/releases/tag/v2.3.3) so maybe it makes sense there for now?

package.json
LorenVS: The comment in getAnalyserSettings confuses me somewhat. This setting change does not require an analyzer restart, but it does require a project load. The first half of the comment makes it seem like I should add the new setting, and the last half makes it seem like I shouldn't

package.json
LorenVS: Yeah, the vast majority of users will not have an analyzer that will support this option, at least for now. I'll move the flag to the release notes for now.

src/extension.ts
LorenVS: The registerWorkspaceSymbolProvider call was the only one that didn't take a file type filter, so I figured it was active for the whole workspace, and calling it again with identical parameters wouldn't be usefull

package.json
DanTup: You're right; the last line is bogus. We include options here that just generally require a reload (because of behaviour at activation time). I've improved the comment in 2a8527608e39d401f56608476a4ccb241ecacf7c.

package.json
DanTup: The release notes won't exist for this yet - you're free to create a placeholder in CHANGELOG but otherwise I can do it when preparing a release (I go through the Issues + Changesets in a release when compiling the notes).

I do think we should remove it from packages.json to hide it for now though, at least until the ability to use it is simpler (I presume "building it yourself" isn't the long-term plan!).

src/extension.ts
DanTup: Oops, didn't spot that. Makes sense :-)

I think it's worth making the array/loop though, to avoid the copy/paste. Makes it less likely someone will add a new provider for only one language accidentally in future.

package.json
LorenVS: I've renamed the flag to previewAnalyzeAngularTemplates, removed it from package.json, added a v2.4.0 section to CHANGELOG.md and added a note about this flag there

package.json
LorenVS: I've renamed the flag to previewAnalyzeAngularTemplates, removed it from package.json, added a v2.4.0 section to CHANGELOG.md and added a note about this flag there

package.json
LorenVS: The other preview flag is not listed in the README, so I've only added a description to the CHANGELOG

src/extension.ts
LorenVS: Changed to using an array of filters and forEach-ing. I left registerWorkspaceSymbolProvider as is.The only one handled manually is registerOnTypeFormattingEditProvider, since this is no corresponding implementation in the angular_analysis_plugin

DanTup:
Thanks! I've merged this in and opened #396 as a place to gather feedback. I also noted that you don't need to overwrite the analyzer snapshot since you can set `dart.analyzerPath` (another hidden setting - maybe some of these don't need to be hidden!).

I had a quick test and was able to get some errors, but I couldn't get completion on the property shown in my screenshot on #396.. Not sure if this is a bug or just unfinished?

@MichaelRFairhurst What's the end goal for distribution of this? Will it come in the SDK or via `pub` with Angular, or will there always be steps for the user? (I hope not, but if there is, maybe we can have a command in Dart Code to do it for them).

DanTup:
@LorenVS I shipped a small update (v2.3.4) with this in, since it'll probably be a while before I get around to implementing more of the things I wanted in v2.4. If you're trying it out, please let me know if you have any issues! :-)

MichaelRFairhurst:
@DanTup our external story should be coming together shortly! Looks like eventually (timeline malleable) users importing angular will just have to add something like

```
plugins:
  angular:
    enabled: true
```

and then the analysis server will run our plugin at no extra hassle. Angular itself will tell the analysis server which version of our plugin to run, for versioning compliance reasons.

DanTup:
Great; that sounds much simpler :-)

PR #399, opened 2017-07-31
Activate Dart-Code on encountering html files

LorenVS:
Activates the Dart-Code extension when html files are encountered. Optimally, this would only be enabled if the `dart.previewAnalyzeAngularTemplates` option is set, but I don't think there is infrastructure for that. This solves issues with auto complete and warnings not showing up in Angular templates unless a dart file is opened first.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
 	"icon": "media/icon.png",
 	"activationEvents": [
 		"onLanguage:dart",
+		"onLanguage:html",
 		"onView:dartPackages",
 		"workspaceContains:**/pubspec.yaml"
 	],


DanTup:
I'm a little nervous about this - it'll activate Dart Code for a large number of unrelated projects. There's an activation even for when the workspace contains a pubspec, which I presume yours does, so I think that might be a bug we can fix?

LorenVS:
Aaah, that would probably work for most people, unfortunately, our directories do not contain pubspecs

LorenVS:
I've created a vscode bug, and I'll see what the response is before pursuing this further

DanTup:
Do you have an example of an Angular project without a pubspec? I assumed you'd need a pubspec to list Angular as a dependency?

LorenVS:
We use bazel [1] for building our Angular2/Dart projects, and manage our dependencies through bazel's deps arguments. Probably not a common environment

[1] https://github.com/dart-lang/bazel

DanTup:
I wonder if we could do workspaceContains:*.dart?

DanTup:
See https://github.com/Microsoft/vscode/issues/27483, apparently we can do this:

```
"activationEvents": [ "workspaceContains:**/*.dart" ]
```

If this works for you, I'd much prefer it to activating on opening all HTML files. WDYT?

DanTup:
I've changed this to activate when the workspace contains any dart file (#409). I believe this solves your problem without activating on any HTML file opening. The change will be in 2.4. Please let me know if you believe/find it doesn't solve your problem. Thanks!

LorenVS:
Thanks @DanTup , and sorry for not responding. This sounds great

DanTup:
Cool; the change was shipped in the last update - let me know if you have any issues with it!

PR #418, opened 2017-08-25
Remove text reference to FLUTTER_HOME

Skylled:
As of v1.4.0, the FLUTTER_HOME variable is no longer used, in favor of FLUTTER_ROOT. This message told the user to set their FLUTTER_HOME instead of FLUTTER_ROOT.


DanTup:
Good spot; thanks!

PR #434, opened 2017-10-20
Update Fuchsia Dart SDK location

yyoon:
The Dart SDK location changed under fuchsia tree, and this commit
reflects that change.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -190,7 +190,7 @@ function findFuchsiaDartSdk(fuchsiaRoot: string): string {
 	else if (process.platform == "darwin")
 		platformName = "mac";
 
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks! How quickly will this change be live for everyone? Do we need to consider searching both locations to avoid breaking people that aren't on the bleeding edge?

DanTup:
Whoops, I confused Flutter and Fuchsia for a minute, so probably everything is bleeding-edge anyway!

Expecting to do a release today, so I'll include this in it. Thanks!

DanTup:
This changes has shipped. Thanks!

yyoon:
Nice. Thanks for shipping it so quickly!

Issue #222, opened 2016-11-22
Allow quick switching of SDK versions


DanTup:
I often find myself swapping between SDKs, eg. testing dev releases that have fixes for bugs I've reported, switching to an older version to try and repro bugs, testing Dart Code against older SDKs that users might be using.

I think this could be made much easier with fairly little effort..

Add a new user setting that points to a folder that is a *collection* of SDKs. When set, the version number shown in the status bar can be clicked on to open a picklist (same as how Code's indent settings work) and let you pick a sub-folder from that location (the list should only include those that look like SDKs). This would set the `sdkPath` setting and prompt to reload (or maybe automatically reload if no dirty buffers).

This should support symlinks too, eg. so if you have checked out versions of the SDK repo you could have a `dev` (or whatever) symlink pointing there for easy switching in and out of release/dev.

If we can detect the architecture, we could also possibly add a new option to the "new SDK" notification for "Download now" which automatically downloads and extracts (this might need specific x-plat code) to that folder and switches to it.

@devoncarew Might this be useful to you? Anything else to add?

devoncarew:
I generally run with the latest dev sdk release (or sometimes, with the sdk version that flutter has vendored in), but am likely an atypical user. I suspect that many users just stay on the latest stable release, or latest dev release.

DanTup:
Maybe this isn't that useful to normal users, though would be pretty useful to me so might still do if it's easy and clean!

Skylled:
Would it be possible to turn `dart.sdkPath` into a map along the lines of:
```
    "dart.sdkPaths": {
        "Stable": "C:\\Program Files\\Dart\\dart-sdk",
        "Dev": "C:\\Dart-Dev\\dart-sdk"
    },
    "dart.currentSdk": "Stable"
```
Then add a command to switch SDKs and reload?

Same could technically be done for Flutter SDK, if someone wanted to have master branch and alpha branch separately installed, or a local branch.

DanTup:
My plan (which was formed pre-flutter) was to keep it as a string and just scan all adjacent folders for the dart binary to get the possible SDKs. I have my SDKs all together like this:

```text
M:\Apps\Dart <-- latest
M:\Apps\Dart-1.20.0
M:\Apps\Dart-1.21.0
M:\Apps\Dart-{...}
```

Whenever a new SDK comes out I just rename the old folder and drop the new one in Dart. Currently I also then change my settings path to switch between them.

[...]

PR #437, opened 2017-11-07
Update the Dart SDK path finding logic for fuchsia

yyoon:
This changes the logic for finding the correct Dart SDK location for
Fuchsia.

Also, now the user-defined "dart.sdkPath" setting takes precedence, in
case the auto-detected Dart SDK is not the desired one to use.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -36,19 +36,28 @@ export function findSdks(): Sdks {
 	const fuchsiaRoot = findFuchsiaRoot();
 
 	const flutterSdk = isFuchsiaProject ? findFuchsiaFlutterSdk(fuchsiaRoot) : isFlutterProject ? findFlutterSdk() : null;
-	const dartSdk = isFuchsiaProject ? findFuchsiaDartSdk(fuchsiaRoot) : isFlutterProject ? findFlutterDartSdk(flutterSdk) : findDartSdk();
+
+	// The user defined Dart SDK should override the auto-detected ones.
+	let dartSdk: string;
+	if (config.userDefinedSdkPath) {
+		// We don't expect the user to add .\bin in config, but it would be in the PATHs
+		dartSdk = findDartSdkInPaths([path.join(config.userDefinedSdkPath, "bin")]);
+	}
+
+	// Failed to find the Dart SDK from the user defined path. Try to auto-detect one.
+	if (dartSdk == null) {
+		dartSdk = isFuchsiaProject ? findFuchsiaDartSdk(fuchsiaRoot) : isFlutterProject ? findFlutterDartSdk(flutterSdk) : findDartSdk();
+	}
 
 	return { dart: dartSdk, flutter: flutterSdk };
 }
 
 function findDartSdk(): string {
 	let paths = (<string>process.env.PATH).split(path.delimiter);
+	return findDartSdkInPaths(paths);
+}
 
-	// We don't expect the user to add .\bin in config, but it would be in the PATHs
-	let userDefinedSdkPath = config.userDefinedSdkPath;
-	if (userDefinedSdkPath)
-		paths.unshift(path.join(userDefinedSdkPath, "bin"));
-
+function findDartSdkInPaths(paths: string[]): string {
 	// Resolve all paths to allow things like ~
 	paths = paths.map(resolveHomePath);
 
@@ -173,7 +182,7 @@ function findFuchsiaFlutterSdk(fuchsiaRoot: string): string {
 	if (!fuchsiaRoot)
 		return null;
 
-	const fuchsiaFlutterPath = path.join(fuchsiaRoot, "lib/flutter");
+	const fuchsiaFlutterPath = path.join(fuchsiaRoot, "third_party/dart-pkg/git/flutter");
 	if (hasFlutterExecutable(path.join(fuchsiaFlutterPath, "bin")))
 		return fuchsiaFlutterPath;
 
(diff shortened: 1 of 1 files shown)


DanTup:
Ok, just let me know when it's good to go (and whether it's best to release immediately after merging).

yyoon:
Thanks for waiting :)

So the Dart location ended up remaining the same as before. I reverted that portion from the PR.
Still, this PR makes the `dart.sdkPath` to take priority, because it would be more natural. Our team members tried to temporarily override the sdk path by setting that property and then surprised by the fact that it's only used after the auto detection is failed.

Also, it updates the flutter location under fuhsia.

Because the sdk location is reverted back, the analysis is all working again and we're unblocked. This doesn't need to be released immediately, but would be nice to be included in the next release.

DanTup:
Is it valid to use a standard Dart SDK for Fuchsia dev? I was under the impression (but this may be incorrect) that the Dart SDK shipping in Flutter was patched in someway, so a standard one wouldn't work. If this is true and the same for Fuchsia, then I wonder if using userDefinedSdkPath might be incorrect?

Eg., I use userDefinedSdkPath to jump between Stable/Dev SDKs when testing (https://github.com/Dart-Code/Dart-Code/issues/222#issuecomment-315556211), would this interfere with Fuchsia/Flutter dev? If so, would I be better having separate overrides for them?

(cc @devoncarew - maybe you know the details about whether Flutter's Dart SDK is actually custom or not)

devoncarew:
Flutter's is using a _mostly_ stock version of the Dart SDK. They pin it at a specific version - it's generally slightly older than the last dev sdk. A few new libraries are added into the `dart:` namespace via a special analyzer mechanism: https://github.com/flutter/engine/blob/6c5315c3a1d72ee2aae520e05c1ebf0f6794ae0e/sky/packages/sky_engine/lib/_embedder.yaml.

For regular Flutter development, you are better off using the analysis server from flutter's version of the dart sdk, but that above mechanism let's most dart sdks work with flutter code. Not sure if that's true for fuchsia development or not -

yyoon:
> Is it valid to use a standard Dart SDK for Fuchsia dev? I was under the impression (but this may be incorrect) that the Dart SDK shipping in Flutter was patched in someway, so a standard one wouldn't work. If this is true and the same for Fuchsia, then I wonder if using userDefinedSdkPath might be incorrect?

In Fuchsia, the situation is similar to what @devoncarew said, but we generally need more recent version of Dart SDK than the one Flutter uses, as we make fuchsia-specific changes to the sdk relatively frequently. Therefore, we now use the latest prebuilt Dart SDK, instead of a pinned version, which gets automatically downloaded when we update our source code.

Occasionally, there could be some gap between when a fuchsia-change is made and when it gets shipped in the new version of SDK. That's when we need to temporarily switch to the locally built Dart SDK instead of the prebuilt one, to make the Dart analysis work correctly. Also, many of us thought the `dart.sdkPath` would override the auto-detection and were surprised when we realized that it's not the case. Because there was no way to force override the dart sdk path with the setting, we needed to build a patched version of Dart-Code to fix it temporarily, which is not ideal.

> Eg., I use userDefinedSdkPath to jump between Stable/Dev SDKs when testing (#222 (comment)), would this interfere with Fuchsia/Flutter dev? If so, would I be better having separate overrides for them?

Do you use set the sdk paths in the user settings or the workspace settings? I would think that's what the workspace settings in VSCode are for, and setting the right set of sdks under the fuchsia workspace should generally work. Even when setting it in the user settings, I would imagine that adding the fuchsia's Dart SDK locations as the options would just work.

DanTup:
> Do you use set the sdk paths in the user settings or the workspace settings? I would think that's what the workspace settings in VSCode are for

I set it in the user settings, since I want it set globally (and to keep out of source control since it contains paths).

I'm slightly conflicted on making the same override work for them all - if I switch to an older Dart SDK for testing Dart stuff, probably doesn't make sense for it to just break my Flutter stuff because it's a standard SDK and Flutter needs a patched one. So I think it'd make sense for flutter to have its own override. But then it feels weird if Fuchsia doesn't have its own?

I'm wondering it it'd make most sense to have three separate overrides (Dart, Flutter, Fuchsia) so the user is in complete control. That way the Dart SDK override can point at a stock SDK, the Flutter one can point at a Flutter-specific one and for consistency, Fuchsia gets its own (and probably there will be a runtime override to support WSL too, but that's for another day!).

I can merge and ship this PR for now (will likely be Thursday - currently I'm not set up for publishing from my MacBook and the PC is disconnected - need to fix one of these soon though!). I think some of this SDK detection code could do with tidying up (it's grown over time, and I think it's become a bit cluttered and hard to follow) so trying to decide on the best way for this to work to avoid re-working it again afterwards. Unless anyone has any strong opinions for why a shared override setting is better, probably I'll keep them separate.

yyoon:
Both approaches sound okay to me. I see the benefit of having separate overrides when you're using the user settings as you mentioned.

DanTup:
This change was pushed out in v2.4.4. I've also raised two new cases:

#440 to support separate Flutter + Fuchsia overrides (since the main override might just point at a normal Dart SDK)
#441 to extend userDefinedSdkPaths (note the S!) and Flutter + Fuchsia equivs to support pointing at a single folder - this will allow you to quickly switch between a "normal" and a single dev SDK without opening the settings file and pasting paths in (I think this will support the use case mentioned above pretty well)

yyoon:
Thanks!

Issue #350, opened 2017-07-05
Move snippets to completion provider to allow conditionals


DanTup:
Since MS [don't want to](https://github.com/Microsoft/vscode/issues/26943) add contexts to snippets, we should build them in the completion provider. This will allow us to add flutter-specific snippets without adding useless noise to the Dart completion list.

DanTup:
Done on a branch, but there's a slight difference from built-in snippets. See https://github.com/Microsoft/vscode/issues/41202

DanTup:
For now, I'm going to hide the previews (to protect against any changes MS may make to the rendering). The previews are nice, but non-critical. We'll put them back when MS have decided how best to handle this.

Actually - we could leave the standard Dart ones using the old mechanism, so at least only the conditional ones lose their previews.

DanTup:
On reflection I've decided to leave this as-is. It means the user may see things like `typedef ${1:Type} ${2:Name}(${3:params});` but that's probably better than seeing just `typedef` and then all these placeholders appearing when they complete.

PR #473, opened 2017-12-15
Add Flutter snippets

xqwzts:
Adds snippets for Flutter StatelessWidget and StatefulWidget classes

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -126,6 +126,38 @@
 				"  $2",
 				"});"
 			]
+		},
+		"Flutter StatelessWidget": {
+			"prefix": "stless",
+			"description": "Insert a StatelessWidget",
+			"body": [
+				"class $1 extends StatelessWidget {",
+					"",
+					"\t@override",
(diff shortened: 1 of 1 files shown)


Skylled:
I love these snippets. I use similar ones in my own setup.

The only concern, as I recall, was being able to have context before offering a snippet. Such as only showing a Flutter snippet when working on a Flutter project. I'm not sure that's yet possible in VS Code.

xqwzts:
@Skylled Looking at this https://github.com/Dart-Code/Dart-Code/issues/350 it seems the way towards context-aware snippets is to use the completion provider.

DanTup:
Yeah; I'd like to move this to the completion provider before adding, so they can be filtered by project type. I want to sort out Angular plugin stuff for v2.7.0, but I'll put this into v2.8.0 and see if I can do it there. It shouldn't be a lot of work.

xqwzts:
Looking forward to it, thanks @DanTup

DanTup:
Flutter snippets will be in the next version (see #474) :)

PR #513, opened 2018-01-29
Added more trigger characters to auto completion

### LINKED ISSUE #512: Automatically open completion when typing import statements
Splitting this from #320. The request is to have completion automatically open when typing an import:

```dart
import '
```

Currently you have to manually invoke it (and several users have expressed that they didn't know the shotrcut to do that, and were mainly relying on auto triggering).

pulyaevskiy:
Fixes #512 .

I tried to tweak the logic as you described but realized that thanks to completion of `import` statements (which inserts `import '|';` I rarely actually type `import '` or `import "`. So it wasn't doing anything useful for me.

I did try different combinations and settled on the one in this PR. I spent about an hour in the editor working on some Dart code with this setting and (to my personal taste) found it to be exactly what I wanted.

One edge case I found was not related to this change:

In annotations, if I type e.g. `@TestOn(` auto complete is triggered (by `(` character) even though there is nothing to complete in this case. This one is tricky since in some cases an annotation can have named arguments in which case it actually could be useful.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -163,10 +163,11 @@ export function activate(context: vs.ExtensionContext) {
 		activeFileFilters.push(HTML_MODE);
 	}
 
+	const triggerCharacters = ".: =(${'\"abcdefghijklmnopqrstuvwxyz".split("");
 	activeFileFilters.forEach((filter) => {
 		context.subscriptions.push(vs.languages.registerHoverProvider(filter, hoverProvider));
 		context.subscriptions.push(vs.languages.registerDocumentFormattingEditProvider(filter, formattingEditProvider));
-		context.subscriptions.push(vs.languages.registerCompletionItemProvider(filter, completionItemProvider, ".", ":", " ", "=", "(", "$", "{"));
+		context.subscriptions.push(vs.languages.registerCompletionItemProvider(filter, completionItemProvider, ...triggerCharacters));
 		context.subscriptions.push(vs.languages.registerDefinitionProvider(filter, definitionProvider));
 		context.subscriptions.push(vs.languages.registerDocumentSymbolProvider(filter, documentSymbolProvider));
 		context.subscriptions.push(vs.languages.registerReferenceProvider(filter, referenceProvider));
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -163,7 +163,7 @@ export function activate(context: vs.ExtensionContext) {
 		activeFileFilters.push(HTML_MODE);
 	}
 
-	const triggerCharacters = ".: =(${'\"abcdefghijklmnopqrstuvwxyz".split("");
+	const triggerCharacters = ".: =(${'\"".split("");
 	activeFileFilters.forEach((filter) => {
 		context.subscriptions.push(vs.languages.registerHoverProvider(filter, hoverProvider));
 		context.subscriptions.push(vs.languages.registerDocumentFormattingEditProvider(filter, formattingEditProvider));
--- a/src/providers/dart_completion_item_provider.ts
+++ b/src/providers/dart_completion_item_provider.ts
@@ -40,6 +40,10 @@ export class DartCompletionItemProvider implements CompletionItemProvider {
 			switch (context.triggerCharacter) {
 				case "{":
 					return line.endsWith("${");
+				case "'":
+					return line.endsWith("import '") || line.endsWith("export '");
+				case "\"":
+					return line.endsWith("import \"") || line.endsWith("export \"");
 			}
 		}
 


src/extension.ts
```diff
@@ -163,10 +163,11 @@ export function activate(context: vs.ExtensionContext) {
 		activeFileFilters.push(HTML_MODE);
 	}
 
+	const triggerCharacters = ".: =(${'\"abcdefghijklmnopqrstuvwxyz".split("");
```
DanTup: I don't like the idea of adding the whole alphabet - what was this to try and do?

My concern is that if a user is in the middle of typing a word (for ex) and has dismissed completion, this will cause it to re-open on the next keystroke. Based on the previous bug on `{` I think we need to be really careful about where we pop completion open automatically, as if the next character pressed is one that commits completion (enter/tab/etc) then it'll cause random text to appear.

DanTup:
> I rarely actually type `import '` or `import "`. So it wasn't doing anything useful for me.

Good point! I think it'd make sense for us to be able to open completion after a completion event. I'll take a look at whether this is possible and if not raise a request against Code. We shouldn't have to decide between being able to complete the `import "";` command and getting automatic triggering of completion on the filename.

Maybe @mjbvz is familiar with working around this area? We have a snippet/completion item that inserts `import "";` and puts the cursor inside the quotes - can we trigger completion automatically? (We know how to make it open by putting a trigger on the quote and then scanning the line to see if it's an import, but it sounds like that doesn't trigger if the whole line was inserted via completion and not explicitly typed)

src/extension.ts
pulyaevskiy: I understand your concern, I was actually hesitant to introduce this for the same reasons.

After spending some time testing with and without full alphabet in trigger characters I found that:

> My concern is that if a user is in the middle of typing a word (for ex) and has dismissed completion, this will cause it to re-open on the next keystroke.

This is actually how it works today in 2.7.3, according to my testing. Feel free to test it yourself though.

Adding full alphabet did not change this behavior for me at all. Including if I'm inside strings other than in import (and export) statements.

I suspect if Analyzer doesn't have any suggestions (which is the case inside strings) there is nothing to show, so we don't see it.

> idea of adding the whole alphabet - what was this to try and do?

It is what you said - invoke completion popup when user types a character in `a-z` range. This solves the imports (and exports) issue including when user tries to modify it. And, as far as I can tell, this doesn't change anything else comparing to how it works today.

I will have to spend more time with this to get more confidence, but this is my observation so far.

I'm not sure if it's the best approach in this case, but I thought I'd create this PR as a suggestion to get more opinions from people who know more than me.

mjbvz:
@DanTup Yes, this is actually what our css extension does when you complete a css property name. Just [set `editor.action.trigger` suggest as the command](https://github.com/Microsoft/vscode-css-languageservice/blob/9e8e4dceaeb8721456e0e160a94b2ee64e922948/src/services/cssCompletion.ts#L157) on the completion item

src/extension.ts
DanTup: > This is actually how it works today in 2.7.3, according to my testing. Feel free to test it yourself though.

You're right! However, there is a slight difference. What's happening today is that when you press the next character, Code is re-displaying the previous results but filtered (client-side) and not actually going to the completion provider. With your change, it always goes to the completion provider which means the changes results in a lot more over-the-wire communication (and I guess possibly increased CPU in the analyzer).

I think a better fix would be to trigger on the quotes and review the line to see if it's an import, and then use the mechanism @mjbvz suggested (thanks!) to trigger it for completions. I think this will provide the behaviour we want, without sending a request to the analyzer for a new completion list on every key press.

WDYT? I should have time tomorrow evening to play around with it.

pulyaevskiy: > which means the changes results in a lot more over-the-wire communication

Yeah, this is bad.
I think what @mjbvz suggested is a nice improvement and we should definitely do that.

I will update this PR to handle quotes as originally suggested.
Tried to find where "suggest a command" fix can be applied, but didn't find anything at the first look.

DanTup: > Tried to find where "suggest a command" fix can be applied, but didn't find anything at the first look.

I assume you mean where to force completion when the `import "";` snippet is entered? This actually depends a little on whether it's coming from the analysis server or a snippet in Dart Code (I can't remember - however, there's a branch that converts snippets to a completion provider which I want to merge soon...).

For the completion provider, we need to add the command when we build the completion item:

https://github.com/pulyaevskiy/Dart-Code/blob/e0a25bff2dd77faafdbdb5a33b2715f23af460c4/src/providers/dart_completion_item_provider.ts#L119-L134

We'd probably want to build a whitelist so we can choose exactly which ones trigger the completion (so in the first instance, just the one with `import "";` or w/e as the completion text.

pulyaevskiy:
PR is updated to only trigger on quotes.

DanTup:
Awesome, thank you!

I can sort the other part (triggering when using the snippet) later in the week if you're not sure about it (or if you want some more info to have a crack, lmk).

pulyaevskiy:
Thanks @DanTup !

If you have a quick hint which file/API I need to use I can definitely try and submit a PR for the suggested trigger part.

DanTup:
It's basically adding the code @mjbvz linked to the completion item:

```js
command: {
    title: 'Suggest',
    command: 'editor.action.triggerSuggest'
}
```

but only if the completion item is the import one. The changes will go somewhere around https://github.com/Dart-Code/Dart-Code/blob/b6af7b57e4f32872a0e0c25ce3d80b6ee011d2b7/src/providers/dart_completion_item_provider.ts#L119-L134

We just need a condition for when to add it - we may add more in future, so making it easy to extend with either a switch statement or an array of which completionText's we want to trigger for may be worthwhile.

pulyaevskiy:
Cool, sounds pretty straightforward. I'll give it a shot, thanks!

PR #517, opened 2018-01-30
Trigger suggest command for import statement completions

### LINKED ISSUE #512: Automatically open completion when typing import statements

pulyaevskiy:
Fixes #512 .

This should be a nice improvement on top of what we have.

The only edge case not covered would be when editing text inside existing import statement, but it probably happens not as often as writing a new import.


DanTup:
Awesome, thanks! 👍

PR #617, opened 2018-02-25
Update package-lock to match a fresh install

xqwzts:
This PR commits the updated package-lock.json generated by running `npm install` on a fresh checkout of the master branch [which results in a modification of the `"@types/ws"` entry].


DanTup:
Thanks! Not sure why this doesn't get updated on my machine (I just tested again), though all the others are in that form so it makes sense.

Issue #583, opened 2018-02-16
Incorrect block folds on multi-line String


xster:

DanTup:
Currently the folding is done by Coding, and seems to be indent-based, which is why it's failing where it is.

Related:
https://code.visualstudio.com/docs/editor/codebasics#_folding
https://github.com/Microsoft/vscode/issues/3422

DanTup:
Please 👍 https://github.com/Microsoft/vscode/issues/3422 to support the feature request!

DanTup:
v1.21 of Code has a proposed API for code folding:

https://code.visualstudio.com/updates/v1_21#_folding

Issue #578, opened 2018-02-15
Allow imports section to be collapsed


lukef:
It would be convenient for code reviews and general tidiness if the imports section could be collapsed similar to how IntelliJ does it.

Example:
 
<img width="193" alt="screen shot 2018-02-15 at 3 00 40 pm" src="https://user-images.githubusercontent.com/37941/36285610-1222a6d6-1261-11e8-8af9-37676a48a540.png">

Low priority as its basically a convenience.

DanTup:
Like #583, this requires support from Code so that we can define the folding regions (currently it guesses them based on indents!). Please 👍 https://github.com/Microsoft/vscode/issues/3422 to support the feature request!

DanTup:
v1.21 of Code has a proposed API for code folding:

https://code.visualstudio.com/updates/v1_21#_folding

PR #666, opened 2018-03-08
Add region folding

Mrigank11:


DanTup:
Thanks for this - though I'm a little confused.. I can't find any info on regions in Dart - where did this come from?

VS Code has a proposed API for programatic folding (see [here](https://code.visualstudio.com/updates/v1_21#_folding)) which I expect to ship a month from now, so my plan was going to be to move over to that. I don't know whether this will play nice with it, so we might want to try that out first.

Mrigank11:
Oh, I was not aware of the new proposed API. I agree with you; we should try it out first. I did it the way its currently implemented.

DanTup:
Ok, cool. The issues worth subscribing to for implementation once the API is stable are #583 and #578.

PR #679, opened 2018-03-10
Add keybinding ctrl+f5 to flutter.fullRestart command

### LINKED ISSUE #678: Add a default keybind for Full Restart
Maybe Ctrl+F5? We could use a context to only apply while debugging, so we don't overwrite the Run-without-debugging key.

Phrohdoh:
ctrl+f5 is used on mac too (not cmd+f5)

screen shot 2018-03-10 at 3 22 43 pm

---

If merged this will close #678.

### DIFF
--- a/package.json
+++ b/package.json
@@ -204,6 +204,11 @@
 				"key": "f4",
 				"mac": "f4",
 				"when": "editorLangId == dart"
+			},
+			{
+				"command": "flutter.fullRestart",
+				"when": "dart-code:flutterProjectLoaded && inDebugMode",
+				"key": "ctrl+f5"
 			}
 		],
 		"menus": {


DanTup:
I think I described there this goes badly - it should be in a [`keybindings` section in `contributes`](https://code.visualstudio.com/docs/extensionAPI/extension-points#_contributeskeybindings); it doesn't seem to work for me here.

Also, I think you don't need the `mac` part when it's same as `key`.

You should be able to test by launching in extension dev host (pressing F5) then open the keybinds page and check it's there. If you've already set it in your own keybinding file, you'll need to remove it to ensure you're not just seeing that one.

Phrohdoh:
@DanTup Sorry about that! I have updated this PR (and added a screenshot taken in the debug host to the OP).

Phrohdoh:
Please don't merge this yet, I am fixing the committer and author details to be my personal info (not my work account).

Phrohdoh:
Committer and author credentials have been corrected.

DanTup:
Awesome, thanks! 👍

PR #691, opened 2018-03-14
Remove superfluous 'the' in get-packages workspace selection dialog

Phrohdoh:


DanTup:
Er, whoops!

The second line is actually totally wrong, it's upgrades!

PR #707, opened 2018-03-19
Dont report errors in stable sdk

DanTup:
@devoncarew Please review and let me know what you think... some questions:

1. Is it correct to use the Flutter SDK version when making this decision for a flutter project? It seems strange since the errors are only reported for the analyzer, but if not, all analyzers inside Flutter will be considered dev, since they always have "pre-release" versions (containing hyphens).
2. Is the problem of stable reports being out of date just a consequence of there being no stable releases for so long? Will this change when Dart 2 ships?
3. When Dart 2 ships, won't we see a huge drop in people on dev SDKs and want to revert this?

### DIFF
--- a/src/analysis/analyzer_status_reporter.ts
+++ b/src/analysis/analyzer_status_reporter.ts
@@ -100,8 +100,6 @@ export class AnalyzerStatusReporter {
 	}
 
 	private shouldReportErrors(): boolean {
-		// TODO: Confirm if we should be using Flutter SDK version in flutter
-		// (if we don't, won't all servers always be dev?)
 		if (this.sdks.projectType === ProjectType.Flutter && this.sdks.flutter)
 			return !isStableSdk(getSdkVersion(this.sdks.flutter));
 		else
--- a/src/analysis/analyzer_status_reporter.ts
+++ b/src/analysis/analyzer_status_reporter.ts
@@ -7,9 +7,10 @@ import { Analyzer } from "./analyzer";
 import { Analytics } from "../analytics";
 import { config } from "../config";
 import { PromiseCompleter } from "../debug/utils";
-import { extensionVersion, getSdkVersion, Sdks } from "../utils";
+import { extensionVersion, getSdkVersion, Sdks, ProjectType, isStableSdk } from "../utils";
 
 const maxErrorReportCount = 3;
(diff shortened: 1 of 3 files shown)


src/analysis/analyzer_status_reporter.ts
```diff
@@ -85,6 +99,15 @@ export class AnalyzerStatusReporter {
 		}
 	}
 
+	private shouldReportErrors(): boolean {
+		// TODO: Confirm if we should be using Flutter SDK version in flutter
```
devoncarew: 👍 , this looks good to me

devoncarew:
> Is it correct to use the Flutter SDK version when making this decision for a flutter project? It seems strange since the errors are only reported for the analyzer, but if not, all analyzers inside Flutter will be considered dev, since they always have "pre-release" versions (containing hyphens).

Yes, this seems correct - like you said, without this, we won't get any reports from flutter users.

> Is the problem of stable reports being out of date just a consequence of there being no stable releases for so long? Will this change when Dart 2 ships?

I think a lot of users have moved to using the dev releases, given how long it's been since the last stable.

> When Dart 2 ships, won't we see a huge drop in people on dev SDKs and want to revert this?

We'll still have a percentage of our users on dev releases, and they'll be a canary for issues.

This is really a tradeoff; right now we get reports from users on stable, but we often can't act on them. They either clutter up our issue tracker, or we close them right away (and that sends the wrong message to users about reporting issues).

DanTup:
> Yes, this seems correct - like you said, without this, we won't get any reports from flutter users.

Isn't this the opposite? If we use the Dart SDK version, all flutter users would always be treated as unstable (they all have a `-flutter` pre-release tag) so we'd get lots of reports. If we use the Flutter version (as this code currently does) then when Flutter goes stable, we'll stop getting reports from people using it (even though their Dart SDK may be unstable/recent).

I'm gonna merge as-is for now (using the Flutter version - eg. `0.2.2-pre.55`) and can change it over if/when you think we should.

devoncarew:
> Isn't this the opposite?

Oops, yup! What you have is correct. We want report from dart-dev (and flutter non-stable), but not from dart stable and flutter stable. Thanks!

DanTup:
Great, thanks for confirming!

Issue #739, opened 2018-03-27
Show Memory usage during Flutter debug sessions


DanTup:
Similar to https://github.com/flutter/flutter-intellij/pull/2006 though we'll be limited to text in the status bar for now. To avoid polluting the status bar too much, we may wish to consider adding this into the device-picker status bar label.

- [x] Show memory usage
- [x] Maybe only show in profile mode to avoid misleading info in debug mode?
- [ ] Subscribe to GC events

DanTup:
Notes from reviewing IntelliJ source (looking at just memory for now)

- Wait for a performance listener to be added (I think this is happening when the panel is rendered - for us, this will just be always since we're writing to status bar)
- Set a timer that runs every 1 second that does:
  - Call `getVM()`
  - Enumerate Isolates in response
    - Call `getIsolate()`
    - Add `heap.used` and `heap.external` to `current`
    - Add `heap.total` and `heap.external` to `total`
  - Render `current of total`

Additionally it seems to subscribe to GC events but I can't figure out exactly what it's doing there (seems like it's updating the graph, but I can't see how that differs to what's being done in the polling). @devoncarew Do I need this bit?

DanTup:
@devoncarew Is it correct that I'm reading `_heaps` (with an underscore!) to get this info?

In the doc [here](https://github.com/dart-lang/sdk/blob/master/runtime/vm/service/service.md#types) it shows `heaps` without an underscore (however there's no other information in the doc about this).

devoncarew:
Yes, the underscore convention in the service protocol means undocumented, and not public API - `_heaps` is correct.

Re: the GC events - we get those per isolate when a GC happens. We add that into the data from the polling in order to have a slightly high frequency view of the data. We don't have new info for the heaps for the other isolates, but do for that isolate.

DanTup:
Split FPS into #804 to avoid tying these together.

DanTup:
@devoncarew 

> Re: the GC events - we get those per isolate when a GC happens. We add that into the data from the polling in order to have a slightly high frequency view of the data. 

We're currently already polling every second, so is there much value in this?

devoncarew:
Some value - you'll get slightly more granular data - but not in any way a deal breaker. I have a bug open to get the before data in the GC event. That would let us graph the two data points for each GC event and know exactly what the high and low were.

DanTup:
[...]

PR #751, opened 2018-03-29
Add memory usage to status bar

DanTup:
See #739.

@devoncarew Please take a look. I'm a little skeptical about the numbers - this reads like "6MB of 9MB" when I run the stocks example app (iOS simulator)... Does this seem in the right ballpark? I'm surprised anything would consume that little memory these days when my editor consumes GBs!

- [x] Add option to disable (maybe separate for debug/profile, maybe with diff defaults?)
- [x] Add in old heap values

### DIFF
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -75,7 +75,7 @@ export class DebugCommands {
 				const memory = e.body.memory;
 				const message = `${Math.ceil(memory.current / 1024 / 1024)}MB of ${Math.ceil(memory.total / 1024 / 1024)}MB`;
 				this.debugMetrics.text = message;
-				this.debugMetrics.tooltip = "This is the amount of memory being consumed by your application out of what has been allocated.";
+				this.debugMetrics.tooltip = "This is the amount of memory being consumed by your applications heaps (out of what has been allocated).\n\nNote: memory usage shown in debug builds may not be indicative of usage in release builds. Use profile builds for more accurate figures when testing memory usage.";
 				this.debugMetrics.show();
 			}
 		}));
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -23,11 +23,12 @@ export class DebugCommands {
 	private currentDebugSession: vs.DebugSession;
 	private progressPromise: PromiseCompleter<void>;
 	private reloadStatus = vs.window.createStatusBarItem(vs.StatusBarAlignment.Left);
+	private debugMetrics = vs.window.createStatusBarItem(vs.StatusBarAlignment.Right, 0);
 	private observatoryUri: string = null;
 
 	constructor(context: vs.ExtensionContext, analytics: Analytics) {
 		this.analytics = analytics;
-		context.subscriptions.push(this.reloadStatus);
+		context.subscriptions.push(this.reloadStatus, this.debugMetrics);
 		context.subscriptions.push(vs.debug.onDidReceiveDebugSessionCustomEvent((e) => {
 			if (e.event === "dart.progress") {
 				if (e.body.message) {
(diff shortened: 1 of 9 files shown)


src/commands/debug.ts
```diff
@@ -70,6 +71,12 @@ export class DebugCommands {
 			} else if (e.event === "dart.flutter.firstFrame") {
 				// Send the current value to ensure it persists for the user.
 				this.sendAllServiceSettings();
+			} else if (e.event === "dart.debugMetrics") {
+				const memory = e.body.memory;
+				const message = `${Math.ceil(memory.current / 1024 / 1024)}MB of ${Math.ceil(memory.total / 1024 / 1024)}MB`;
```
devoncarew: If we're very tight on space, we could just show the current heap usage (`78MB`).

src/debug/dart_debug_impl.ts
```diff
@@ -805,6 +810,32 @@ export class DartDebugSession extends DebugSession {
...
+
+		const isolatePromises = vm.isolates.map((isolateRef) => this.observatory.getIsolate(isolateRef.id));
+		const isolatesResponses = await Promise.all(isolatePromises);
+		const isolates = isolatesResponses.map((response) => response.result as VMIsolate);
+
+		let current = 0;
+		let total = 0;
+
+		for (const isolate of isolates) {
+			current += isolate._heaps.new.used + isolate._heaps.new.external;
```
devoncarew: We also need to add in the data from the `old` heap - that's the new generation (young) objects, and the longer lived objects, that have been promoted to the old heap.

DanTup:
> I believe the slightly odd number is from not counting in the old heap (n isolates, 2 heaps per).

Aha, ofcourse! You did explain this but I guess I didn't write it down! Should be easy to fix.

> We probably want to decide whether we want to have an option to show in debug mode (, and a separate option to show in profile mode?)

How useful/useless is the value in debug mode? Maybe we should *only* show it in profile?

DanTup:
I added in the old heap too, and I'm still seeing similar numbers (eg. 6 of 10MB). I have no point of reference to know whether this is in the right ballpark or not - what sort of numbers would you expect to see running the Stocks app on an Android device?

src/commands/debug.ts
DanTup: I don't think it looks too bad like this for now (though admittedly my screen is quite wide..)

screen shot 2018-04-03 at 08 44 55

src/debug/dart_debug_impl.ts
DanTup: Done in bc4a42d though it didn't seem to change the numbers much.

devoncarew:
Hmm, those numbers look too small; for debug mode, by an order of magnitude, and for profile mode, still 2-3x too optimistic.

devoncarew:
Perhaps opening the same app in the observatory, to get a second opinion?

DanTup:
@devoncarew The old/new heap in Observatory do seem to add up to about the same:

screen shot 2018-04-16 at 12 34 52

Though VM total is much higher:

screen shot 2018-04-16 at 12 35 03

So, is what I'm showing correct? Is it confusing to the user if we say they're using 10MB when the VM is using 100? Are there other values I should add on?

(This was a profiling build).

devoncarew:
OK, for a profiling build those heap totals make sense. 

We will need to be careful in docs / tooltips / messaging about the difference between the process size and the dart heap sizes (there's some overhead from the VM), but still useful to show users the dart heap sizes, as that's something they have direct control over.

DanTup:
I updated the tooltip to this:

> This is the amount of memory being consumed by your applications heaps (out of what has been allocated).
> 
> Note: memory usage shown in debug builds may not be indicative of usage in release builds. Use profile builds for more accurate figures when testing memory usage.

Currently you'll only see memory in profile builds, or if you manually set `showMemoryUsage` in your launch config. I'm gonna merge as-is for v2.12 (so it'll be in a beta build this week); let me know if you want me to tweak anything.

PR #808, opened 2018-04-16
add selectDeviceOnConnect setting

343max:
Every few minutes flutter would find my iPhone or iPad on wifi and select it as the current device which would drive me crazy when all I wanted to do was to develop on the simulator.

So I added an option that won't switch the device one `dart.selectDeviceOnConnect` is set to `false`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -590,6 +590,12 @@
 					"items": {
 						"type": "string"
 					}
+				},
+				"dart.selectDeviceOnConnect": {
+					"type": "boolean",
+					"default": true,
+					"description": "Select newly connected devices as current device.",
+					"scope": "window"
 				}
 			}
 		},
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -590,6 +590,12 @@
 					"items": {
 						"type": "string"
 					}
+				},
+				"dart.selectDeviceOnConnect": {
```
DanTup: I think maybe we should prefix it with `flutter` to match the other Flutter settings. This makes it convenient to find all the flutter settings in the options (in hind-sight, maybe they should be `flutter.` instead of `dart.`... maybe something to change in future!).

package.json
```diff
@@ -590,6 +590,12 @@
...
+					"type": "boolean",
+					"default": true,
+					"description": "Select newly connected devices as current device.",
```
DanTup: Minor nit: I think prefixing with "Whether to..." would be more consistent with the descriptions of other options.

DanTup:
This setting is available in the beta of v2.12:

https://github.com/Dart-Code/Dart-Code/releases/tag/v2.12.0-beta.1
https://dartcode.org/releases/v2-12/

PR #809, opened 2018-04-17
Prevent syntax from matching closure syntax (=>) as the greater than operator (>)

aemino:


DanTup:
This seems to result in the `=` and `>` being coloured differently which looks undesireable:

screen shot 2018-04-17 at 07 39 23

I'm not sure if the existing behaviour is bad, what is it you're seeing/trying to change?

The best fix here is for us to let the analysis server provide all the colouring, but it requires an API from Code which hasn't made progress https://github.com/Microsoft/vscode/issues/585 :(

aemino:
I personally use the One Dark theme for VSCode.  For me, sometimes, the closure operator is monochromatic.
e.g.
image

However, in the case of function notation, the `>` symbol is colored, which messes up the ligatures I'm using.
e.g.
image

I thought that this was due to this syntax highlighting match, but it seems that this occurs differently for different users.  I'm not sure what a better solution for this would be.

DanTup:
> I thought that this was due to this syntax highlighting match, but it seems that this occurs differently for different users. I'm not sure what a better solution for this would be.

I'm not sure, I don't like the ligatures so never really used them. It's possible your fix is on the right lines, but the `=` and `>` are being set as separate tokens. Maybe if you explicitly add `=>` as a match to that group it may fix this (so they're just one token of the same type, rather than two adjacent?).

> On another note, I sometimes feel that the full color palette of my syntax theme isn't being utilized—is this something wrong with the theme or perhaps the result of how the syntax highlighter works?

It could be a combination of the theme or the definitions in the syntax file. Honestly there are a lot of highlighting issues with the current code (it's not possible to accurately parse Dart with regex, which is basically what this file is doing) but it's something I've been kinda tolerating in the hope that one day https://github.com/Microsoft/vscode/issues/585 will be done and we can get accurate information from the analysis server. I think all languages in Code are similarly constrained by this restriction :(

DanTup:
LGTM, does it solve your ligature issue this way?

aemino:
Yep, it seems to do the trick!
image

DanTup:
Excellent, thanks!

PR #830, opened 2018-04-23
Stash evaluateName to allow building watch expressions for child variables

### LINKED ISSUE #829: "Add to Watch" on child values doesn't work
On `p.name` the `Add to Watch` action just adds `name` so it doesn't work.

screen shot 2018-04-23 at 10 59 18 am

DanTup:
Fixes #829.

@devoncarew Could you take a look if this makes sense? The issue was that if you clicked `Add to Watch` on a field within an instance, you'd just get its name (since MS removed their heuristic in https://github.com/Microsoft/vscode/commit/00cbee4bbf66290b81eb737367016586bfbdba25) in the Watch window:

screen shot 2018-04-23 at 11 34 32 am

As far as I can tell, when we have an instance from the VM we don't have any info on how to build a string that would evaluate to it. So, when we call `storeData` to stash the InstanceRef, I'm supplementing it with its `evaluateName` (which by default it just its name), and then when we build child-variables, we prefix it with the parents `evaluateName` (formatted based on whether it's a field, association, item):

screen shot 2018-04-23 at 11 37 07 am

Does this seem like a sane thing to do? Can you think of any edge cases this code might not handle?


devoncarew:
> when we build child-variables, we prefix it with the parents evaluateName

Yes, this looks good, and makes sense - the user would want the textual path to the symbol (`foo[0].bar`), rather than something like the captured `foo[0]` object, and watching the `bar` field.

PR #856, opened 2018-04-30
Add preview of hot reload coverage markers

### LINKED ISSUE #611: Hot Reload "coverage" markers
Some things left over from notes in the PR (which has merged for preview):

- [ ] Ensure we don't send requests for coverage if another is already in progress
- [ ] Handle errors better
- [ ] Add automated tests

DanTup:
Work in progress...

- [x] Remove markers when code is executed
- [x] Review logic for when to request coverage
- [x] Move decorations to gutter w/ tooltip
- [ ] ~~Add gutter tooltips~~
- [ ] Ensure we don't send requests for coverage if another is already in progress
- [ ] Handle errors better
- [x] Address all TODOs
- [ ] Add automated tests


src/debug/dart_debug_impl.ts
```diff
@@ -738,6 +744,65 @@ export class DartDebugSession extends DebugSession {
...
+			const scriptRefs = flatMap(libraries, (library) => library.scripts);
+
+			// Filter scripts to the ones we care about.
+			// TODO: Do we need to getPossibleSourceUris() here?
+			const scripts = scriptRefs.filter((s) => scriptPaths.indexOf(uriToFilePath(s.uri)) !== -1);
+
+			for (const scriptRef of scripts) {
+				const script = (await this.observatory.getObject(isolate.id, scriptRef.id)).result as VMScript;
+				try {
+					const report = await this.observatory.getSourceReport(isolate, [SourceReportKind.Coverage], scriptRef);
```
DanTup: @devoncarew Currently I'm sending a LOT of requests to the VM for this:

- getVM
- for each isolate, getIsolate
- for each library in each isolate, getLibrary
- for each script in each library in each isolate, getScript
- for each script in each library in each isolate, getSourceReport

I'm not sure how safe it would be to cache any of these (and we don't really have any caching in the debug adapter right now, except for the storeData/getData lookup for passing back "references" to Code), since presumably new isolates can be created, new libraries can be loaded, etc.

Looking at your code, I think you're doing the same (`getScriptRefFor` seems to wrap some of it up). I'm not sure if that's just because it's work-in-progress or it's the correct thing to do though.

devoncarew: So:

- the VM calls are fast :) (modulo the getSourceReport() call)
- I re-request everything each time I query for coverage, but cache during that call (for things like script token positions)
- I only query for the current active file, given the cost of getSourceReport

DanTup: Ok, sounds similar :) I'm not explicitly caching anything, but if there's only one isolate and a script can only exist in one library, then I don't think there will be duplicated requests for anything? (I do reuse the same `Script` instance for mapping all tokens in the response since it's in scope with the results).

I'm currently doing "all visible files" so that if a user has a split view, it'll update each (rather than them appearing/disappearing as they move focus between the two panes, which would also be increased calls) but once I've got everything working I can check timings and see how things are.

Thanks!

devoncarew: Makes sense re: caching - my impl was written before we changed to a very targeted view of the markers (I think you're correct in that the 'cache' would only be hit once).

> I'm currently doing "all visible files" so that if a user has a split view

This will likely be performant - requesting for a small number of files will be fast. The general call - getSourcereport w/o qualifying the script - requests for all 600-700; that is slow.

DanTup:
Gonna merge this for easier previewing. The outstanding items from the list above have been moved into #611 which is the "make it production ready" issue and will remain open.

PR #875, opened 2018-05-04
Handle initial thread states other than PauseStart

stuartmorgan:
Passes initial thread pause states through the same handling as when
they are received as Debug events. This is should have no effect on
'launch' mode, where these states would not be encountered, but
will improve future 'attach' handling.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -178,9 +178,8 @@ export class DartDebugSession extends DebugSession {
 							isolate.runnable ? "IsolateRunnable" : "IsolateStart",
 						);
 
-						if (isolate.pauseEvent.kind === "PauseStart") {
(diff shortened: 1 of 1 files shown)


stuartmorgan:
This is to support https://github.com/Dart-Code/Dart-Code/pull/867 but since it stands alone I thought it would be better to pull out into a separate patch to avoid that patch becoming progressively larger over time. If you prefer I fold it in over there, just let me know!

This change makes attaching to a previously-paused process show the correct pause state, so it gets the stack and provides the continue button, rather than being stuck with at pause button that doesn't work.

(I didn't put this on top of dev-2-13 per your comment about requiring the Insider version of VS Code, but presumably it'll be easy to move as necessary.)

DanTup:
LGTM, thanks!

PR #876, opened 2018-05-04
Fix static error - Dart 2 semantics

kevmoo:


DanTup:
This code is used in a test (you can run the tests from the Side Bar and selecting the "Dart" tests launch config):

```
  2) dart_hover_provider returns expected information for a method taking a function:
      AssertionError: 'methodTakingFunction((String) → void myFunc) → void' == 'methodTakingFunction((String) → int myFunc) → void'
      + expected - actual
      -methodTakingFunction((String) → void myFunc) → void
      +methodTakingFunction((String) → int myFunc) → void
      
      at Object.<anonymous> (C:\projects\dart-code\test\dart_only\providers\dart_hover_provider.test.ts:136:10)
      at Generator.next (<anonymous>)
      at fulfilled (C:\projects\dart-code\out\test\dart_only\providers\dart_hover_provider.test.js:4:58)
      at <anonymous>
```

I've been meaning to fix it for a while, but always been in the middle of something and knowing it needed the test tweaking, put it off =)

Issue #879, opened 2018-05-06
Debugger doesn't properly show content of map


tfriem:
When in a breakpoint, it's not possible to see the content of Map variables.

Versions:
VS Code 1.23
Dart Code 2.12.1

This is how it looks like in VS Code:

![vscode](https://user-images.githubusercontent.com/4568372/39674683-bd74efe4-514f-11e8-840a-53e12a400e7c.PNG)

![vscode_open](https://user-images.githubusercontent.com/4568372/39674712-535829e0-5150-11e8-80d4-ff66038f3238.PNG)

There are two problems:
1. Not all entries are shown. There should be 4 entries, but only one is shown.
2. It's not possible to see the value of the, key as only the type is shown.

The first point is related to VS Code only showing entries where the keys differ. As it's always the type of the key, only one entry will be shown. This could be quickly fixed by adding an index (see PR).

For the second point, it seems necessary to add a new level for key and value of each entry. That is more than I would do with my limited code insight.

This is how it looks like in Android Studio:

![android_studio](https://user-images.githubusercontent.com/4568372/39674687-c171821a-514f-11e8-9651-162705670861.PNG)

DanTup:
I've split the complex-keys part into #882 so this issue can just be about not showing all of the keys. I'm going to merge the PR as a fix for this and will work on the other issue this week (in addition to adding tests). If you'd like to have a go at it, let me know and I can hold off.

PR #880, opened 2018-05-06
Add an index to map entries for debugger display

tfriem:
Related to #879

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -501,16 +501,18 @@ export class DartDebugSession extends DebugSession {
 									variables.push(this.instanceRefToVariable(thread, canEvaluate, `${instanceRef.evaluateName}[${i + start}]`, `[${i + start}]`, element));
 								}
 							} else if (instance.associations) {
+								let index = 0;
 								for (const association of instance.associations) {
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks!

I've actually made some tweaks to this for the next version already (v2.13) so this will need rebasing on top of that (the changes were to improve the display and also added some automated tests).

You can rebase this on the `dev-v2-13` branch or wait for a few days when I'll likely merge that into master. We should also extend the tests to cover this (for example see the `provides local variables when stopped at a breakpoint` test at https://github.com/Dart-Code/Dart-Code/blob/dev-v2-13/test/dart_only/debug/dart_cli.test.ts).

DanTup:
Also; I'm not sure how this solves the issue - isn't this just prefixing the name with the index? Don't we need to add an additional node for the key so it can be examined (like in the IntelliJ screenshot)?

DanTup:
Oh, I see now, this was related to:

> The first point is related to VS Code only showing entries where the keys differ. As it's always the type of the key, only one entry will be shown. This could be quickly fixed by adding an index (see PR).

I'd actually prefer we fix this properly for the next version. Let me know if you'd like to take a stab at it and/or require help, otherwise I can take a look later in the week.

Issue #877, opened 2018-05-04
Include `PUB_ENVIRONMENT` when invoking pub command


kevmoo:
https://github.com/Dart-Code/Dart-Code/blob/5b97e7a66251917d121f3cde591f737b0248a602/src/commands/sdk.ts#L142

When invoking pub, please set the `PUB_ENVIRONMENT` environment variable to `vscode.dart-code`.

This helps us track how pub is being invoked.

See an example here: https://github.com/dart-lang/sdk/blob/332dc7f3e90babd370fe4342603f5d74d8518a23/pkg/analysis_server/lib/src/plugin/plugin_manager.dart#L846-L861

A couple of notes:

- [ ] If `PUB_ENVIRONMENT` is already set, please amend your value w/ a `:`. So `existing.value:vscode.dart-code`

- [ ] When running tests, pass in a environment variable so we can know any hits to the pub site or via a test. Our convention is to use `bot` in the value. So for you `dart-code.test.bot`

Yes, I'd love the same model for invoking pub via flutter – so any call to `flutter packages`. I'll file a separate bug at https://github.com/Dart-Code/Flutter

DanTup:
Makes sense; we already do something similar for flutter commands (`FLUTTER_HOST = "VSCode";`).

For tests; I presume you mean Dart Code's tests running on Travis/AppVeyor and not the end users tests in the IDE?

PR #885, opened 2018-05-07
Set environment variables for Pub

DanTup:
This implements #877 as requested by @kevmoo.

We already have an env var we pass for flutter (which sets `FLUTTER_HOST`). Rather than adding more and having to make sure they were used int he correct places, I've made a single env variable which sets both `FLUTTER_HOST` and `PUB_ENVIRONMENT` and then is used inside `safeSpawn` (which is the function we always use for spawning processes to handle Windows path escaping issues).

In theory, this should solve all the requirements (including that when Pub is invoked by Flutter, assuming it passed the env through).

Does this sound reasonable? Are there any problems with sending both of these all the time?

@kevmoo I assumed the note about test runs was for Dart Code's tests (and not end users tests) but if this is incorrect, let me know. @devoncarew Also let me know if you want me to change the FLUTTER_HOST for test runs (we do run `flutter config --no-analytics` in the Travis/AppVeyor scripts but that doesn't cover people (like me) running the tests locally during dev).

### DIFF
--- a/src/analysis/analyzer.ts
+++ b/src/analysis/analyzer.ts
@@ -69,7 +69,7 @@ export class Analyzer extends AnalyzerGen {
 		// Register for version.
 		this.registerForServerConnected((e) => { this.version = e.version; this.capabilities.version = this.version; });
 
-		this.createProcess(undefined, dartVMPath, args, undefined);
+		this.createProcess(undefined, dartVMPath, args);
 
 		this.serverSetSubscriptions({
 			subscriptions: ["STATUS"],
--- a/src/debug/flutter_run.ts
+++ b/src/debug/flutter_run.ts
@@ -1,13 +1,12 @@
 import { Disposable } from "vscode";
 import * as f from "../flutter/flutter_types";
 import { StdIOService, UnknownNotification, UnknownResponse } from "../services/stdio_service";
-import { flutterEnv } from "./utils";
 
 export class FlutterRun extends StdIOService<UnknownNotification> {
 	constructor(flutterBinPath: string, projectFolder: string, args: string[], logFile: string) {
 		super(() => logFile, true, true);
 
-		this.createProcess(projectFolder, flutterBinPath, ["run", "--machine"].concat(args), flutterEnv);
+		this.createProcess(projectFolder, flutterBinPath, ["run", "--machine"].concat(args));
 	}
 
 	protected shouldHandleMessage(message: string): boolean {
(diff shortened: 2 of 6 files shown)


DanTup:
Merging this; but @devoncarew if you think we should do something similar with `FLUTTER_HOST` for Dart Code's test runs (or that we should make the values between these two more consistent), lmk and I'll update.

devoncarew:
> Also let me know if you want me to change the FLUTTER_HOST for test runs (we do run flutter config --no-analytics in the Travis/AppVeyor scripts but that doesn't cover people (like me) running the tests locally during dev).

I think the CI systems are our biggest concern. If you like to pass in `FLUTTER_HOST=VSCode.test_run` or similar for tests, sounds good (though, we could just change to not passing in for test runs at all...).

DanTup:
@devoncarew It was CI test runs we were talking about here (specifically, Dart Code's CI runs). We are turning off analytics (I presume this is your preference) but if you want me to append a value so if they do come through anywhere they can be identified, let met know.

Currently, we're not doing any different for end users test runs. If there's any reason to, also let me know and I can tweak.

devoncarew:
> It was CI test runs we were talking about here (specifically, Dart Code's CI runs). We are turning off analytics 

Ah, gotcha. Yeah, turning off analytics complete there sounds good.

> Currently, we're not doing any different for end users test runs.

We just may want individuals who run tests to opt-out of analytics locally, so that those runs don't skew the data too much.

DanTup:
> We just may want individuals who run tests to opt-out of analytics locally,

Now I'm more confused, I was talking about users running their own tests for their on Flutter apps there. I don't think you would want them to opt out (it's also opt them out for non-test runs)?

Did you think I meant people running Dart Code's tests locally? Is there a way to opt-out of Flutter analytics only for a session?

devoncarew:
You can run `flutter --suppress-analytics foo` to suppress analytics for a single flutter command run, and `flutter config --no-analytics` to disable for a machine. Above, I was referring to disabling analytics for a developer machine that frequently runs the VS Code tests, in order to reduce the noise in the analytics stream. Sorry for any confusion :)

DanTup:
Ok, I've added a `globalFlutterArgs` array that gets prefixed to flutter args everywhere and we put --`suppress-analytics` into it based on the env var we set in all test runs (893f788f228a804f568589c3bd81da6efdacd41a).

PR #891, opened 2018-05-10
Add support for attaching to process

DanTup:
Some notes from initial test:

- [x] Currently unpauses upon attaching; is this the best option for attach, or should we leave it paused for the user to resume?
- [x] We should default the packages file in code to avoid it needing to be in the config
- [x] We should add a `Dart: Attach to Process` command to eliminate the need for the launch.json config
- [x] If you type junk into the observatory box it gives a bad error (seems to get passed to the WebSocket library)
- [x] The prompt for the observatory uri looks a bit weird (large bold text), should check if we can tweak this
- [x] ~~Can we get the packages info/paths from the VM (@stuartmorgan did you already look at this? I won't bother if you concluded it wasn't possible, but it'd be cool if the open project didn't need to match the exact folder that's open)~~ #896
- [x] Url is handled incorrectly if it already ends with `/ws/` (eg. `http://127.0.0.1:8181/ws/`)
- [x] Support relative path for .packages

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -865,7 +865,7 @@ export class DartDebugSession extends DebugSession {
 
 			if (shouldRemainedStoppedOnBreakpoint) {
 				thread.handlePaused(event.atAsyncSuspension, event.exception);
-				this.sendEvent(new StoppedEvent(reason, thread.number, exceptionText));
+				this.sendEvent(new StoppedEvent(reason, thread.num, exceptionText));
 			} else {
 				this.observatory.resume(thread.ref.id);
 			}
@@ -1048,13 +1048,11 @@ class ThreadManager {
 	public nextThreadId: number = 0;
 
 	public threads: ThreadInfo[] = [];
-	public debugSession: DartDebugSession;
 	public bps: { [uri: string]: DebugProtocol.SourceBreakpoint[] } = {};
 	private hasConfigurationDone = false;
 	private exceptionMode = "Unhandled";
 
-	constructor(debugSession: DartDebugSession) {
-		this.debugSession = debugSession;
+	constructor(public readonly debugSession: DartDebugSession) {
 	}
 
 	public registerThread(ref: VMIsolateRef, eventKind: string) {
@@ -1066,7 +1064,7 @@ class ThreadManager {
 			this.threads.push(thread);
 
 			// If this is the first time we've seen it, fire an event
-			this.debugSession.sendEvent(new ThreadEvent("started", thread.number));
+			this.debugSession.sendEvent(new ThreadEvent("started", thread.num));
 
 			if (this.hasConfigurationDone)
 				thread.receivedConfigurationDone();
@@ -1099,14 +1097,14 @@ class ThreadManager {
 
 	public getThreadInfoFromNumber(num: number): ThreadInfo {
 		for (const thread of this.threads) {
-			if (thread.number === num)
+			if (thread.num === num)
 				return thread;
 		}
 		return null;
 	}
 
 	public getThreads(): Thread[] {
-		return this.threads.map((thread: ThreadInfo) => new Thread(thread.number, thread.ref.name));
+		return this.threads.map((thread: ThreadInfo) => new Thread(thread.num, thread.ref.name));
 	}
 
 	public setExceptionPauseMode(mode: string) {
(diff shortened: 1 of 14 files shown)


src/debug/dart_debug_impl.ts
```diff
@@ -221,14 +264,42 @@ export class DartDebugSession extends DebugSession {
 					this.sendEvent(new TerminatedEvent());
 			}, 100);
 		});
+
+		this.observatory.onError((error) => {
+			if (completer.promise) {
+				completer.resolve(error);
```
DanTup: Probably should `reject` here.

src/debug/dart_debug_impl.ts
```diff
@@ -221,14 +264,42 @@ export class DartDebugSession extends DebugSession {
...
+			const removeBreakpointPromises = [];
+			for (const thread of this.threadManager.threads) {
+				removeBreakpointPromises.push(thread.removeAllBreakpoints());
+			}
+			Promise.all(removeBreakpointPromises).then((_) => {
+				// Restart any paused threads.
+				const resumePromises = [];
+				for (const thread of this.threadManager.threads) {
+					if (thread.paused) {
+						resumePromises.push(this.observatory.resume(thread.ref.id));
```
DanTup: This doesn't seem to be working for me; when I disconnect it remains paused.

src/debug/dart_debug_impl.ts
```diff
@@ -721,23 +792,32 @@ export class DartDebugSession extends DebugSession {
...
-
-				// Output any logpoint messages.
-				for (const logPoint of logPoints) {
-					// TODO: Escape triple quotes?
-					const logMessage = logPoint.logMessage
-						.replace(/(^|[^\\\$]){/g, "$1\${") // Prefix any {tokens} with $ if they don't have
-						.replace(/\\({)/g, "$1"); // Remove slashes
-					const printCommand = `print("""${logMessage}""")`;
-					await this.evaluateAndSendErrors(thread, printCommand);
+				// When attaching to an already-stopped process, this event can be handled before the
```
DanTup: @stuartmorgan I'm not quite sure what's going on here. Is this that we're hitting breakpoints in the VM before the debug adapter knows about them?

src/debug/dart_debug_impl.ts
DanTup: Hmm, can't repro now :)

DanTup:
Soooo, it turns out I had a bug in my tests and they were all *launching* new apps rather than attaching.

Whoops! Fixing...

src/debug/dart_debug_impl.ts
stuartmorgan: Right, specifically if you attach to a process that is already stopped on a breakpoint (you can test by either using the Observatory web UI, or by disabling the cleanup-on-disconnect code, attaching, detaching while stopped on a breakpoint, and then re-attaching). In that case, this code runs before your recently-added code that builds the breakpoint map.

src/debug/dart_debug_impl.ts
stuartmorgan: It could be that the thread.paused logic is incorrect in some cases. It worked in simple cases, but I didn't test it extensively.

src/providers/debug_config_provider.ts
```diff
@@ -129,6 +107,35 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
...
+			}
+		}
+	}
+
+	private async getObservatoryUri(): Promise<string> {
+		let userInput = await vs.window.showInputBox({ prompt: "Enter Observatory address. This can be a full URL, or just a port for localhost." });
+
+		// If the input is just a number, treat is as a localhost port.
+		if (userInput && /^\s*[0-9]+\s*$/.exec(userInput)) {
+			userInput = `http://127.0.0.1:${userInput}`;
```
stuartmorgan: This lost the trim(), so if there's whitespace at the start it'll create an invalid URL.

DanTup:
> I was surprised by this as well. I left it alone since it was the existing behavior, but changing it to stay paused certainly seems reasonable to me.

Cool, I'll do that :)

> I didn't end up investigating that once I discovered that it was easy to configure. I figured it would be a good follow-up for the future but not on the critical path.

Cool; I didn't want to spend time if you'd discovered it wouldn't work. I'll make sure it's on a list somewhere if I don't look at it soon.

> Whoops, forgot to tolerate a trailing slash.

No worries, I'll add some better tests and fix it up :-)

> Right, specifically if you attach to a process that is already stopped on a breakpoint (you can test by either using the Observatory web UI, or by disabling the cleanup-on-disconnect code, attaching, detaching while stopped on a breakpoint, and then re-attaching). In that case, this code runs before your recently-added code that builds the breakpoint map.

Gotcha :)

> It could be that the thread.paused logic is incorrect in some cases. It worked in simple cases, but I didn't test it extensively.

After not changing anything deliberately, it seems to work. I'll investigate more if I see it again; it's possible I was getting confused (especially as I had a bug in the tests that meant it was launching new processes instead of attaching!).

> This lost the trim(), so if there's whitespace at the start it'll create an invalid URL.

Whoops, good catch - I think I planned to move the trim further up or something, but must've got distracted.

I'm doing some tidying up and trying to flesh the tests out a little more, and I'll tackle these issues. Once I think it's all stable I'll give you a ping if you want to test that I haven't broken anything for how you were using it.

PR #896, opened 2018-05-14
Attach to dart without declaring a path to the .packages file

DanTup:
@devoncarew @stuartmorgan I made some changes to the debugger to support finding the `.packages` file from the `rootLib` (to aid attaching) but it's based on the assumption that we'll always get an isolate with a `rootLib` (and that the first one that does, if there are many, is the one we're interested in). Does this sound like a valid assumption?

Another option is we always use the `.packages` from the open folder, but this means if you attach a project you don't have open, weird things might happen.

Assuming this is ok, it might also be a step towards being able to open VS Code and attach to a Dart project (and from there, debug the source) without first opening the folder (however, that'll obviously come with limitations, as you won't have the folder open in explorer, and I don't think we could open it without restarting the debug session).

### DIFF
--- a/package.json
+++ b/package.json
@@ -691,10 +691,6 @@
 								"type": "string",
 								"description": "Workspace root."
 							},
-							"packages": {
-								"type": "string",
-								"description": "Path to the packages file, if it's not at the workspace root."
-							},
 							"observatoryUri": {
 								"type": "string",
 								"description": "URI (or port) of the Observatory instance to attach to."
--- a/package.json
+++ b/package.json
@@ -691,6 +691,10 @@
 								"type": "string",
 								"description": "Workspace root."
 							},
+							"packages": {
+								"type": "string",
+								"description": "Path to the packages file (only required if cannot be discovered from the running process automatically)."
+							},
 							"observatoryUri": {
 								"type": "string",
 								"description": "URI (or port) of the Observatory instance to attach to."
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -663,10 +663,6 @@
 								"type": "string",
 								"description": "Workspace root."
 							},
-							"packages": {
```
stuartmorgan: This should still be an option, since a Dart process can be launched with an arbitrary packages file rather than whatever would be found with the default logic relative to the entry point. (That's where I hit a wall on trying to get this working with the observatory: the root library path works for guessing, but I couldn't find anywhere that explicitly had either the packages file or the full paths for each library)

DanTup: Ah ok, that makes sense :)

If we put this back and the user doesn't specify it, should we prefer the one in their current workspace, or from the VM? My concern with the one from the workspace is that if they have the wrong project open it'll just be wrong.

stuartmorgan: Guessing based on the VM sounds reasonable to me (assuming the assumptions about isolate ordering are valid); I can't think of any reason the one in the folder would be more likely to be right if a non-default packages file was used to launch the process.

DanTup:
@stuartmorgan I've put `packages` back into the launch config, so I think this supports what you want but is optional. Unless anyone objects, I'm hoping to merge both this and the main attach work into master this week (probably tomorrow) and create a beta release.

PR #924, opened 2018-05-21
Show getters in debugger watch/variables/hovers

### LINKED ISSUE #401: Support showing properties in debugger tooltips/watch
There are loads of properties shown in this completion list; yet the debug tooltip only shows a few. Not sure if this is expected (before I investigate this, do you know if this is a bug @devoncarew?) but it seems confusing to me.

screen shot 2017-08-01 at 18 35 48
screen shot 2017-08-01 at 18 36 18

DanTup:
Fixes #401.

@devoncarew Could you have a scan over this to see if it all makes sense? There's a TODO about a few things I'm not totally happy with, but it does seem to work. I don't know what the perf impact would be like on a big object, but someone told me the VM is fast :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -453,6 +453,12 @@
 					"description": "Whether external libraries should be marked as debuggable.",
 					"scope": "resource"
 				},
+				"dart.evaluateGettersInDebugViews": {
+					"type": "boolean",
+					"default": true,
+					"description": "Whether to evaluate getters to display them in debug views (such as Variables, Watch and Hovers).",
+					"scope": "resource"
+				},
 				"dart.sdkPath": {
 					"type": "string",
 					"description": "The location of the Dart SDK to use for analyzing and executing code. If blank, Dart Code will attempt to find it from the PATH environment variable.",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -435,6 +435,12 @@
 					"description": "Whether external libraries should be marked as debuggable.",
 					"scope": "resource"
 				},
+				"dart.evaluateGettersInDebugViews": {
+					"type": "boolean",
+					"default": true,
+					"description": "Whether to evaluate getters to display them in debug views (such as Variables, Watch and Hovers).",
```
devoncarew: `Whether to evaluate getters in order to display them in debug views (such as the Variables, Watch and Hovers views).`?

DanTup:
> Is there any way to call out getters slightly differently in the debugger views?

I've opened #928 to look into this. I can see there's some "presentation hint" in the objects, so I'll see what that does. If it's no good, we can at least tweak the text somehow.

package.json
DanTup: Done!

jacob314: +1 to shipping this by default with all flags turned on. We turned this on by default for dart4web users and never heard a single complaint about getters being evaluated.
We should consider adding a VM service protocol option to evaluate all getters to make this more efficient for cases where lots of objects are displayed.

PR #942, opened 2018-05-24
Support VS Code's Outline View (Requires Code v2.14)

DanTup:


DanTup:
Turns out we don't need to do anything here; it's using conatinerName. There's a better API coming, but it's not ready (#34968).

PR #957, opened 2018-05-29
Simplify breakpoint uris to always send package: where possible

### LINKED ISSUE #868: setBreakPointsRequest can send multiple responses
If getPossibleSourceUris in setBreakPointsRequest (in dart_debug_impl.ts) returns multiple URIs, then the loop that calls threadManager.setBreakpoints multiple times will result in sendResponse being called more than once.

While debugging the extension, this results in:
"attempt to send more than one response for command setBreakpoints"
being logged as an error to the console each time a breakpoint is set (at least while debugging the extension).

### LINKED ISSUE #941: Remove device-path mapping code from Flutter debugger
First we need to get the reload tests all working (with checking paths), but we should be able to remove the code that maps device-local paths back to the host paths.

See https://github.com/Dart-Code/Dart-Code/issues/905#issuecomment-391142245

DanTup:
This fixes #941, however there are a load of tests that are being skipped due to bugs. As such, I don't have total confidence that this change is good.

1. Removes the mapping of paths from host to device
1. Only sends paths for packages as `package:` and not the normal fs path

Note: `1` is not being tested until the `reload-tests` branch is merged, and that's also held up by some bugs.

Also fixes #868.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -432,7 +432,9 @@ export class DartDebugSession extends DebugSession {
 			breakpoints = [];
 
 		// Get the correct format for the path depending on whether it's a package.
-		const uri = this.packageMap.convertFileToPackageUri(source.path) || formatPathForVm(source.path);
+		const uri = this.packageMap
+			? this.packageMap.convertFileToPackageUri(source.path) || formatPathForVm(source.path)
+			: formatPathForVm(source.path);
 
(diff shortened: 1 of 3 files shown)


DanTup:
This still seems broken, seems like we don't hit breakpoints sent in the main app using a `package:` uri:

```
[14:07:08 GMT+0100 (BST)] [Observatory] [Info] ==> {"id":"7","method":"addBreakpointWithScriptUri","params":{"isolateId":"isolates/1056266028","scriptUri":"package:hello_world/main.dart","line":22}}
```

Issue #605, opened 2018-02-21
Flutter Outline


DanTup:
This case is to implement something similar to the Flutter Outline view from IntelliJ (note: this is still heavily in dev, using an experimental API and subject to change!)

![flutter outline in intellij](https://user-images.githubusercontent.com/1078012/37085186-4a0f81be-21ec-11e8-97d9-f7f9f20a1db6.png)

- [x] Implement basic tree
- [x] Navigate to code on selection
- [x] Stop outline refactor commands appearing in the palette
- [ ] Refactors on context menu (https://github.com/dart-lang/sdk/issues/32462)
- [x] Widget (Flutter) icons
- [x] Tree nodes collapse when performing refactor
- [x] Tooltips
- [x] Add a preview flag
- [x] Sorting of context menu is bad
- [x] Require 1.20 analysis server
- [x] Update to new ranges provided by server 1.20
- [x] Stale data renders while waiting for updated data when switching file
- [ ] Highlight code for selected item (https://github.com/Microsoft/vscode/issues/45059)
- [x] Position code on screen better (https://github.com/Microsoft/vscode/issues/45742)
- [x] Reveal tree node as navigating code
- [x] Clicking out of editor (eg into debug output) shouldn't hide Outline (https://github.com/Microsoft/vscode/issues/45188)
- [ ] Set unique IDs per node (globally unique) so updating code/running refactor keeps correct item selected
- [x] Don't collapse nodes on clicking (https://github.com/Microsoft/vscode/issues/34130)
- [x] Show outline above deps (https://github.com/Microsoft/vscode/issues/45053)
- [x] Tree loses focus on click (so can't navigate with keyboard)
- [x] Method/field icons? (https://github.com/Microsoft/vscode/issues/45057)
- [x] Move to its own sidebar icon
- [x] Remove `enableProposedApi`
- [ ] Doesn't unselect node when you move out of all elements (https://github.com/microsoft/vscode/issues/48754)

chances:
What does this issue cover?

DanTup:
Apologies, I thought I'd replied to this! This case is to add something like the Flutter Outline view that was added to IntelliJ recently:

https://groups.google.com/forum/#!topic/flutter-dev/lKtTQ-45kc4

sionicion:
What's the current status on this? Is there a way to test the current state of it?

DanTup:
[...]

PR #961, opened 2018-05-29
Flutter Outline preview

DanTup:
~~Gonna merge this, but behind a flag with an EXPERIMENTAL note because I think it's kinda unusable without a couple of issues (like https://github.com/Microsoft/vscode/issues/48198)~~.

Fixes https://github.com/Dart-Code/Dart-Code/issues/605.


DanTup:
Scratch that - this still feels really unfinished to use. And with VS Code now having a nice Outline preview, it just makes things worse!

There are a few issues I've opened in VS Code that might make it possible for us to use the built-in Outline view for this:

https://github.com/Microsoft/vscode/issues/51332
https://github.com/Microsoft/vscode/issues/51334

winsonet:
but why I still can't use this new featuer (Preview: Flutter Outline) in my VS code?

DanTup:
@winsonet can you be more specific - what do you mean you can't use it? Did you enable it in the user settings, and check you have version v3.3 of the Dart and Flutter extensions? See https://dartcode.org/releases/v3-3/#preview-flutter-outline for notes on this.

If you're still having issues, please open an issue with explicit details on what you're seeing and I can take a look.

winsonet:
Yes, I will open an issue for that, thanks!

PR #978, opened 2018-06-04
Reduce memory usage

DanTup:
@devoncarew I don't expect you to remember this, but hopefully you can sanity check...

In the original debug impl there was some commented out code for removing stored data from the thread manager on `resume`. It passed a list of IDs for data that was stored for that thread, but this list was never populated. It had a comment that said `// TODO: I don"t think we want to do this...`!

This results in the `storedData` building up over a long debug session. I think we *did* want to do what was in the commented code, so I uncommented it and tweaked it so we just loop through the collection of stored data looking for anything that's for that given thread and deleting that.

Tests are passing and everything seems good in manual testing so I think all is good, but I'm not sure what the reason for the original comment/commenting out was. Ring any bells?

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -1328,9 +1328,10 @@ class ThreadManager {
 		return this.storedData[id];
 	}
 
-	public removeStoredIds(ids: number[]) {
-		for (const id of ids) {
-			delete this.storedData[id];
+	public removeStoredData(thread: ThreadInfo) {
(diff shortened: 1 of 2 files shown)


devoncarew:
Hmm, unfortunately I don't have context on this anymore :( Would love to help, but I'd likely need a few hours to get back into the code.

DanTup:
No worries, was a long shot! You don't need to try and find time to dig into it, I'm reasonable sure it's good and I'm gonna try and get an early v2.14 into a few hands this week so if there are issues, we should pick them up quick.

Issue #657, opened 2018-03-07
Request: Implement a `Flutter: Screenshot` command


Phrohdoh:
This should match the `flutter run` CLI `s` command.

DanTup:
@devoncarew I can't see this in the `daemon` docs - is it exposed to editors?

DanTup:
I just noticed that you can run `flutter screenshot` anywhere and it'll take a screenshot and dump it to that folder. So how about:

- Add a new `Flutter: Save Screenshot` command
- Read the path from settings
- If not path, prompt the user for a Folder to save to (and save it into workspace settings - as a relative path if possible - need to check how this will work from a workspace file)
- If folder doesn't exist (eg. was read from settings and has since been deleted), create it
- Invoke `flutter screenshot` in that folder

timsneath:
Taking a look at this one, as a way to get more familiar with the Dart-Code extension. Started hacking something up in the flutter-screenshot branch of my fork.

DanTup:
Cool, let me know if you need any assistance or when you want feedback 👍

Ramalus:
> I just noticed that you can run flutter screenshot anywhere and it'll take a screenshot and dump it to that folder. So how about:
> 
> Add a new Flutter: Save Screenshot command
> Read the path from settings
> If no path, prompt the user for a Folder to save to (and save it into a variable so we don't prompt again for this session)
> If folder doesn't exist (eg. was read from settings and has since been deleted), create it
> Invoke flutter screenshot in that folder
 

@DanTup and @timsneath maybe this could be something more platform specific. At least for the Mac. On macs the default saving location for screenshots is the `~/Desktop` folder.

Then it should be possible to add a path config if one wants to change the default behaviour. (something like we used to do with dart code's logs, for instance).

I'm not sure what the default behaviour for screenshots is on windows and I'm pretty sure different Linux desktop environments use different default locations, but at least for mac (and probably windows) it could follow the system's default.

Just a suggestion, of course. :)

DanTup:
@Ramalus The plan is to include a setting for you to set a folder, and prompt the first-time-per-session if it's not set. I don't know about shipping defaults without the user choosing them though. On Windows, PrtScreen copies to clipboard but I don't think there's a built-in way that saves anywhere by default.

@timsneath FYI, I might merge and finish this off and include in v2.15; trying to clear out branches and PRs to reduce effort resolving merges later.

timsneath:
[...]

PR #1039, opened 2018-06-26
WIP - Flutter screenshot

Ramalus:
This is another approach to the screenshot issue in #657. Its a work in progress, but what's working so far is:
- On command activation the folder selection pops up (only the first time);
-  If a selection has been triggered before, the folder selection does not popup and it saves to the latest location selected.

TODO:
- Change `screenshotUri` to be a setting and configurable by the user;
- Guarantee that if the setting is filled, the folder selection popup does not appear and the screenshot file goes to the settings path.

Any ideas or suggestions please let me know!

### DIFF
--- a/src/commands/sdk.ts
+++ b/src/commands/sdk.ts
@@ -20,6 +20,7 @@ const flutterNameRegex = new RegExp("^[a-z][a-z0-9_]*$");
 export class SdkCommands {
 	private sdks: Sdks;
 	private analytics: Analytics;
+	private screenshotUri: string;
 	// A map of any in-progress commands so we can terminate them if we want to run another.
 	private runningCommands: { [workspaceUriAndCommand: string]: child_process.ChildProcess; } = {};
 	constructor(context: vs.ExtensionContext, sdks: Sdks, analytics: Analytics) {
@@ -85,18 +86,21 @@ export class SdkCommands {
 		}));
 		context.subscriptions.push(vs.commands.registerCommand("flutter.screenshot", async (uri) => {
 			if (!uri || !(uri instanceof Uri)) {
-				var selectedFolder =
-					await window.showOpenDialog({ canSelectFolders: true, canSelectMany: false, openLabel: "Select folder to save screenshot" });
 
-				if (selectedFolder && selectedFolder.length > 0) {
-					uri = selectedFolder[0].path
-				} else {
-					// Do nothing if the user canceled the folder selection
-					return
+				// Only prompt the user if variable has never been set
+				if (!this.screenshotUri) {
+					var selectedFolder =
+						await window.showOpenDialog({ canSelectFolders: true, canSelectMany: false, openLabel: "Select folder to save screenshot" });
+					if (selectedFolder && selectedFolder.length > 0) {
+						this.screenshotUri = selectedFolder[0].path
+					} else {
+						// Do nothing if the user canceled the folder selection
+						return
+					}
 				}
 			}
 
-			return this.runFlutterInFolder(uri, ["screenshot"], "screenshot");
+			return this.runFlutterInFolder(this.screenshotUri, ["screenshot"], "screenshot");
 		}));
 		context.subscriptions.push(vs.commands.registerCommand("flutter.packages.upgrade", (selection) => {
 			return vs.commands.executeCommand("dart.upgradePackages", selection);
--- a/src/commands/sdk.ts
+++ b/src/commands/sdk.ts
@@ -3,7 +3,7 @@ import * as fs from "fs";
 import * as os from "os";
 import * as path from "path";
 import * as vs from "vscode";
-import { ProgressLocation, Uri } from "vscode";
+import { ProgressLocation, Uri, window } from "vscode";
 import { Analytics } from "../analytics";
 import { config } from "../config";
 import { globalFlutterArgs, safeSpawn } from "../debug/utils";
@@ -12,7 +12,7 @@ import { DartHoverProvider } from "../providers/dart_hover_provider";
 import { DartSdkManager, FlutterSdkManager } from "../sdk/sdk_manager";
 import { dartPubPath, flutterPath, showFlutterActivationFailure } from "../sdk/utils";
 import * as util from "../utils";
-import { ProjectType, Sdks, fsPath, isFlutterWorkspaceFolder } from "../utils";
+import { fsPath, isFlutterWorkspaceFolder, ProjectType, Sdks } from "../utils";
 import * as channels from "./channels";
 
 const flutterNameRegex = new RegExp("^[a-z][a-z0-9_]*$");
@@ -82,6 +82,21 @@ export class SdkCommands {
 				// TODO: Move this to a reusable event.
 				DartHoverProvider.clearPackageMapCaches();
 			}
+		}));
+		context.subscriptions.push(vs.commands.registerCommand("flutter.screenshot", async (uri) => {
+			if (!uri || !(uri instanceof Uri)) {
+				var selectedFolder =
+					await window.showOpenDialog({ canSelectFolders: true, canSelectMany: false, openLabel: "Select folder to save screenshot" });
+
+				if (selectedFolder && selectedFolder.length > 0) {
+					uri = selectedFolder[0].path
+				} else {
+					// Do nothing if the user canceled the folder selection
+					return
+				}
+			}
+
+			return this.runFlutterInFolder(uri, ["screenshot"], "screenshot");
 		}));
 		context.subscriptions.push(vs.commands.registerCommand("flutter.packages.upgrade", (selection) => {
 			return vs.commands.executeCommand("dart.upgradePackages", selection);
--- a/package.json
+++ b/package.json
@@ -256,6 +256,11 @@
 				"command": "flutter.openInXcode",
 				"title": "Open in Xcode",
 				"category": "Flutter"
+			},
+			{
+				"command": "flutter.screenshot",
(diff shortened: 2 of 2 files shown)


Ramalus:
This PR now follows the flow described by @DanTup.

> I just noticed that you can run flutter screenshot anywhere and it'll take a screenshot and dump it to that folder. So how about:
> Add a new Flutter: Save Screenshot command
> Read the path from settings
> If no path, prompt the user for a Folder to save to (and save it into a variable so we don't prompt again for this session)
> If folder doesn't exist (eg. was read from settings and has since been deleted), create it **
> Invoke flutter screenshot in that folder

** In this case, if the folder doesn't exist, it is not being created. But the rest of the flow is working as expected.

DanTup:
@Ramalus Thanks! As discussed; there was already an open PR from Tim with this work so it was only fair I took his PR. I've merged it into a branch since I don't think it's quite ready; with some notes on what I plan to change here:

https://github.com/Dart-Code/Dart-Code/pull/1042

If you're interested in looking at some of those things, you could branch from the `flutter-screenshot` branch and target your PR there and build on top of that work.

Thanks for the effort either way! 👍 🙂

Ramalus:
@DanTup Sure, no worries.

I've just updated the PR to target `Dart-Code:flutter-screenshot` and I'll take a look at the points you mentioned in #1042.

👍 :)

DanTup:
Cool; give me a shout if you need any assistance!

(you may have already noticed, looks like the merge resulted in a compile error: `Duplicate identifier 'flutterScreenshotPath'.`)

Ramalus:
Yes, I had used the same identifier as Tim. I will be traveling this next
few days but I'll try to take a look at this as soon as possible.

Most of what you suggested is already working, I just need to polish a few
things.

DanTup:
@Ramalus Let me know if you think you'll have time to get back to this soon; if not I'm happy to finish it off. It'd be good to get into v2.17 which is likely to ship around the end of the month.

Ramalus:
Hi Danny,

I believe I will be able to tackle it this weekend.
I moved to another country this past week, so it has been quite turbulent.

I’ll let you know if that’s not the case.

Thank you for your patience.

DanTup:
No worries; hope the move went well. Let me know if you can't get to it and I'll sort it out. Thanks!

Ramalus:
@DanTup Thank you! The move did go very well.

So I did find some time to work on this.

I believe its almost ready, the one thing I wasn't really able to do is make the **"Open Folder"** button work since I'm not sure how to open a file explorer window on a specific folder from within vscode.

But as for the rest of the feature, it works as such:

### On first run
`flutterScreenshotPath` is `null` and as for most cases, so is the settings config.
- The user is prompted with a folder selection popup;
- After selection, the screenshot is saved in the specified location;
- A notification is shown letting the user know the screenshot process finished and where to find the SS.

### On second run
`flutterScreenshotPath` is **not** `null` since it has been set in the first run
- The user is **not** prompted with a folder selection popup;
- No notification is thrown since he has already ran the command for this session and theoretically already knows there the SS was stored.

### On third run
User changes the config setting to a path different than the one previously set for this session
- User it **not** prompted with a folder selection popup;
- User is notified since the directory the SS was saved to has changed since the previous run.
**If the user did not change anything, it defaults to the *second run* behaviour.**

To finish this up, I just need to know how to open a file explorer from within vscode.

Let me know if there's anything else you would like to add or anything you think should be implemented differently.

src/commands/sdk.ts
```diff
@@ -83,6 +84,54 @@ export class SdkCommands {
...
+						this.flutterScreenshotPath = selectedFolder[0].path;
+						shouldNotify = true;
+					} else {
+						// Do nothing if the user cancelled the folder selection.
+						return;
+					}
+				}
+
+				// 	// if folder doesn't exist, create it
+				if (!fs.existsSync(this.flutterScreenshotPath)) {
```
DanTup: I think this might not be recursive? If you rebase on the screenshot branch, you should be able to call `mkDirRecursive`.

src/commands/sdk.ts
```diff
@@ -83,6 +84,54 @@ export class SdkCommands {
...
+						// Set variable to selected path. This allows prompting the user only once.
+						this.flutterScreenshotPath = selectedFolder[0].path;
+						shouldNotify = true;
+					} else {
+						// Do nothing if the user cancelled the folder selection.
+						return;
+					}
+				}
+
+				// 	// if folder doesn't exist, create it
```
DanTup: nit: Doubled comment characters

src/commands/sdk.ts
```diff
@@ -252,28 +301,28 @@ export class SdkCommands {
 		vs.commands.executeCommand("vscode.openFolder", projectFolderUri, openInNewWindow);
 	}
 
-	private async createFlutterScreenshot(): Promise<void> {
-		// get screenshot path
-		let screenshotPath = config.flutterScreenshotPath;
+	// private async createFlutterScreenshot(): Promise<void> {
```
DanTup: If this code isn't used anymore, it should be deleted rather than commented out.

src/commands/sdk.ts
```diff
@@ -83,6 +84,54 @@ export class SdkCommands {
...
+				if (!fs.existsSync(this.flutterScreenshotPath)) {
+					fs.mkdirSync(this.flutterScreenshotPath);
+				}
+			}
+
+			const screenshotResult = this.runFlutterInFolder(this.flutterScreenshotPath, ["screenshot"], "screenshot");
+
+			screenshotResult.then(async (_) => {
+				if (shouldNotify) {
+					const res = await vs.window.showInformationMessage("Screenshot saved at " + this.flutterScreenshotPath.toString(), "Open folder");
```
DanTup: Nit: Use template string: ``` `Screenshot saved at ${this.flutterScreenshotPath}` ```

DanTup: Also `Open Folder` instead of `Open folder` to match other buttons.

src/commands/sdk.ts
```diff
@@ -83,6 +84,54 @@ export class SdkCommands {
...
+					if (res) {
+						const uri = vs.Uri.parse(this.flutterScreenshotPath);
+						// TODO: open folder in file explorer.
```
DanTup: We should either implement this, or remove the button.

The only way I can see to do this is maybe with the `revealFileInOS` command but I don't know if that's intended to be a public API (though it' used by TypeScript). @mjbvz is it possible/OK for us to use this?

https://github.com/Microsoft/vscode/blob/9fb32293774874f2c93da51cd83395c02d230e5c/extensions/typescript-language-features/src/typescriptServiceClient.ts#L521

DanTup:
Oops, I overlooked your comment here when reviewing! 

> I believe its almost ready, the one thing I wasn't really able to do is make the "Open Folder" button work since I'm not sure how to open a file explorer window on a specific folder from within vscode.

I added a comment about this in the review, see if the mentioned command works, and if not, let's just remove the `Open Folder` button for now.,

> But as for the rest of the feature, it works as such:

This behaviour all sounds reasonable to me 👍

DanTup:
My rebase has caused conflicts - if they're not trivial to resolve and you want me to look at them, let me know.

src/commands/sdk.ts
mjbvz: Not an official extension API but I think it should be ok to use since we expose it for keybindings

Ramalus:
I'll take another look at this until Thursday. Thank you for the review!

src/commands/sdk.ts
```diff
@@ -83,6 +84,55 @@ export class SdkCommands {
...
+				}
+			}
+
+			const screenshotResult = this.runFlutterInFolder(this.flutterScreenshotPath, ["screenshot"], "screenshot");
+
+			screenshotResult.then(async (_) => {
+				if (shouldNotify) {
+					const res = await vs.window.showInformationMessage(`Screenshot saved at ${this.flutterScreenshotPath}`, "Open Folder");
+					if (res) {
+						const uri = vs.Uri.parse(this.flutterScreenshotPath);
```
DanTup: I think this may need to be `Uri.file`?

src/commands/sdk.ts
```diff
@@ -83,6 +84,55 @@ export class SdkCommands {
...
+						await vs.commands.executeCommand("revealFileInOS", Uri.parse(this.flutterScreenshotPath));
```
DanTup: I think this should use the variable from the line above?

Ramalus:
For me the merge seems to have worked. 
screen shot 2018-07-26 at 08 39 34

GH says there are no conflicts. 🤔 

Also, I tested `revealFileInOS` and it I don't really need to define that const. I've just pushed a commit removing it.

DanTup:
Hmmm, weird..

screen shot 2018-07-26 at 3 07 53 pm

I'll look at it later; and probably just merge later if I can't figure out what's up here.

DanTup:
Btw, the comment about Uri.parse() vs Uri.file() - I think Uri.parse may fail with Windows file paths, so I think that still may need changing. I'm happy to sort later when I merge though.

Ramalus:
Oh I see. I can't really test if that's the case though. No windows computers around 😅 

If you need anything else, just let me know. 👍

Ramalus:
@DanTup Is this still planned for v2.17?

DanTup:
Yep! I ran out of time because I wanted to make a beta before I finished yesterday and the conflict confused me (it merges cleanly, but rebase conflicts). I meant to come back to it. I will merge early next week and it will ship in vNext :)

PR #1049, opened 2018-07-03
Support launching the analysis server on a remote machine over SSH

LorenVS:
In certain environments, it is preferable to modify files remotely over
SSHFS, rather than copying dart files locally. In my case, corporate
policies prevent me from copying source files to a laptop. However, I am
allowed to mount directories from my workstation using SSHFS. Dart-Code
works out of the box this way, but its performance is severely limited
by the analysis server needing to crawl the entire source tree over
SSHFS.

Because the analysis server communicates over STDIN/STDOUT, it is
trivial to instead launch the analysis server on a remote machine using
SSHFS. Other SDK integrations such as flutter and fuschia will not work
properly in these environments, but just launching the analysis server
for code intelligence works wonderfully. The only additional limitation
for such an environment is that the mounted SSHFS file system is present
at the exact same directory path on both the local and remote machines.

VSCode's typescript support is automatically re-arranging some of the import orders. If you would prefer me to undo these changes I will.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -199,12 +199,6 @@
 			"integrity": "sha512-41qEJgBH/TWgo5NFSvBCJ1qkoi3Q6ONSF2avrHq1LVEZfYpdHmj0y9SuTK+u9ZhG1sYQKBL1AWXKyLWP4RaUoQ==",
 			"dev": true
 		},
-		"@types/shell-escape": {
-			"version": "0.2.0",
-			"resolved": "https://registry.npmjs.org/@types/shell-escape/-/shell-escape-0.2.0.tgz",
-			"integrity": "sha512-7kUdtJtUylvyISJbe9FMcvMTjRdP0EvNDO1WbT0lT22k/IPBiPRTpmWaKu5HTWLCGLQRWVHrzVHZktTDvvR23g==",
-			"dev": true
-		},
 		"@types/sinon": {
 			"version": "4.3.1",
 			"resolved": "https://registry.npmjs.org/@types/sinon/-/sinon-4.3.1.tgz",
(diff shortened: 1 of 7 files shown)


DanTup:
This looks good! Few minor comments:

1. Few lint failures on the cI:

```
ERROR: (no-var-keyword) src/analysis/analyzer.ts[86, 3]: Forbidden 'var' keyword, use 'let' or 'const' instead
ERROR: (no-var-keyword) src/analysis/analyzer.ts[87, 3]: Forbidden 'var' keyword, use 'let' or 'const' instead
ERROR: (quotemark) src/analysis/analyzer.ts[94, 17]: ' should be "
ERROR: (trailing-comma) src/analysis/analyzer.ts[101, 29]: Missing trailing comma
ERROR: (semicolon) src/sdk/utils.ts[140, 4]: Missing semicolon
```

2. I don't think I can easily test this so I'm somewhat taking this blind. It doesn't seem like it can affect anyone not opted-in to it and it's fairly simple, so I'm not worried about this; but it might affect how easily/quickly I can resolve issues if they come up!

3. I think it might be better to just put the escape function into a utils file somewhere rather than a package - that package hasn't been touched in 3 years and it's only a single function. I'm somewhat averse to lots of packages - I think they bring much risk and responsibility (people being sloppy with credentials, package constraint conflicts, sometimes huge amounts of dependencies for small functions) - and this one looks really trivial with little advantage to being able to easily update. I know we already have a fairly big dep tree; but that's not all through choice (VS Code's packages bring a lot) 😄

LorenVS:
2. Fair enough. I agree this is a pretty exotic setup. I normally wouldn't even bother trying to upstream this, but I've had other people on my team ask to use my setup, and I don't want to point them to a forked repository if possible.

3. Done. I had a feeling this would come up.

LorenVS:
I think the failing CI build is transient?

DanTup:
And now it's Azure! I think Travis may be having connectivity issues so I'll just land on the basis of Win/Linux all running and passing.

Thanks! 👍

DanTup:
v2.16.0 has shipped with this; VS Code will probably auto-update next time you open it. Let me know if you hit any issues, and thanks for the contribution!

PR #1260, opened 2018-10-18
Don't break in pub when using "All Exceptions" pause mode

### LINKED ISSUE #1117: Do not pause PubTest debugger on exceptions in the pub isolate
Continuing on https://github.com/Dart-Code/Dart-Code/issues/1105, I think it would be useful if the PubTest debugger does not pause on uncaught exceptions that occur in the pub isolate when setting 'All exceptions'. I think that most of the time a developer is interested in all exceptions inside the project.

DanTup:
Fixes #1117.

@devoncarew Does this seem reasonable? The issue is that if a user sets breaking on "All Exceptions" they start hitting exceptions in the `pub` isolate code, for example:

43105109-04302856-8ed4-11e8-820a-54a2917ccb30

This seems undesirable since it's unrelated to user code (which runs in its own isolate). Note this change does not stop it from breaking on caught exceptions in framework code that runs in the users isolates, it's purely infrastructure code that I think they essentially consider a bug-free black box.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -1491,8 +1491,17 @@ class ThreadManager {
 		this.exceptionMode = mode;
 
 		for (const thread of this.threads) {
-			if (thread.runnable)
-				this.debugSession.observatory.setExceptionPauseMode(thread.ref.id, mode);
+			if (thread.runnable) {
+				let threadMode = mode;
+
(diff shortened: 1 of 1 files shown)


DanTup:
(@devoncarew Gonna merge this since I think it's reasonable and trying to close off branches/PRs, but do shout if you think this is a bad idea and I'll revert/amend)

PR #1267, opened 2018-10-24
Add the ability to add "// ignore: xxx" quick-fixes

### LINKED ISSUE #890: Add quick-fixes to insert lint-ignore comments
IN some cases users want to ignore warnings (For example a function used only in testing that isn't called).  

You can do so using:
```
//ignore: unused_element
void _testingOnlyFunc() {}
```

But if you don't already know what "key" to ignore, you have to use flutter analyze to see what the key is. 

Can we make either the "Problems" tab specify that key, or add a right click in problems or option in the show fixes light-bulb to choose "suppress warning" which adds the //ignore for you?

I think users of the dart code plugin would benefit from any way to ignore without having to use the command line if they're trying to ignore some obscure warning they don't know the "key" for.  I'm not oppinionated how that's done.

### LINKED ISSUE #1086: Support ignoring lints (etc.) from the problems view
It'd be handy to have a quick way to apply ignores for lints etc. from the problems list:

https://github.com/Microsoft/vscode/issues/36881

Probably needs analysis server support though (unless we're happy just sticking comments at the end of the range).

DanTup:
Fixes #890.
Fixes #1086.

### DIFF
--- a/package.json
+++ b/package.json
@@ -899,6 +899,12 @@
 					"description": "Whether to show a warning when modifying files outside of the workspace.",
 					"scope": "window"
 				},
+				"dart.showIgnoreQuickFixes": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to show quick fixes for ignore hints and lints.",
+					"scope": "window"
+				},
 				"dart.previewToStringInDebugViews": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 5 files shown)


escamoteur:
Hi Dan, I don't see that quickfix

DanTup:
@escamoteur what sort of message are you trying to ignore? It only shows up for hints/lints. Do you see it here?

```dart
void foo() {
  final a = "test";
}
```

Untitled

If not, please open an issue and include an analyzer log (comment the offending code out, start log, then uncomment - so the log includes the notification of the error).

Thanks!

escamoteur:
Are you sure it's still working? I just tried with your above example and I don't get a quickfix for the lint check.

DanTup:
Yeah, definitely still works here. Though I forgot to mention in my message above that you need to enable it (`"dart.showIgnoreQuickFixes": true,`) since it can be quite spammy. If you already have it enabled, please open an issue and include an analyzer log (comment the offending code out, start log, then uncomment - so the log includes the notification of the error) and I'll take a look. Thanks!

escamoteur:
Bingo!!!

PR #1391, opened 2019-01-15
Allow creation of Flutter samples from the command palette

DanTup:
Adds a `Flutter: Create Sample Project`:

screenshot 2019-01-15 at 2 57 10 pm

screenshot 2019-01-15 at 2 57 16 pm

Selecting one will write the trigger file for a sample then open that folder (in a new window if you already have a project open, reusing the current window if not). It's written to a random temp folder to avoid prompting the user (though not sure whether we'd prefer the other?).

Outstanding:

- [x] Automated tests
- [x] Manual testing of hard-to-test failure cases (eg. missing Flutter SDK gives a good prompt, recovers if you point it at an SDK)
- ~~Switch to master docs path~~ https://github.com/Dart-Code/Dart-Code/issues/1392
- ~~Should we pre-fetch samples and hide this if we can't get them, or just fail at the time we fetch?~~ Failing at fetch time, to make it clearer when something is wrong

### DIFF
--- a/package.json
+++ b/package.json
@@ -330,7 +330,8 @@
 					"command": "flutter.createProject"
 				},
 				{
-					"command": "flutter.createSampleProject"
+					"command": "flutter.createSampleProject",
+					"when": "config.dart.flutterDocsHost"
 				},
 				{
 					"command": "pub.get",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
 			},
 			{
 				"command": "flutter.createSampleProject",
-				"title": "Create Sample Project",
+				"title": "New Project From Docs",
 				"category": "Flutter"
 			},
 			{
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
 		"workspaceContains:**/pubspec.yaml",
 		"workspaceContains:**/*.dart",
 		"onCommand:flutter.createProject",
+		"onCommand:flutter.createSampleProject",
 		"onCommand:flutter.doctor",
 		"onUri"
 	],
@@ -75,6 +76,11 @@
 			{
 				"command": "flutter.createProject",
 				"title": "New Project",
+				"category": "Flutter"
+			},
+			{
+				"command": "flutter.createSampleProject",
+				"title": "Create Sample Project",
 				"category": "Flutter"
 			},
 			{
@@ -322,6 +328,9 @@
 			"commandPalette": [
 				{
 					"command": "flutter.createProject"
+				},
+				{
+					"command": "flutter.createSampleProject"
 				},
 				{
 					"command": "pub.get",
--- a/package.json
+++ b/package.json
@@ -670,6 +670,12 @@
 					"description": "Whether to pass --track-widget-creation to Flutter apps (required to support 'Inspect Widget'). If this setting is not set, it will default to true for Flutter >= 10.2 and false for any previous versions of Flutter.",
 					"scope": "resource"
 				},
+				"dart.flutterDocsHost": {
+					"type": "string",
+					"default": "docs.flutter.io",
+					"description": "The hostname to use for Flutter Docs, for example when fetching available samples.",
+					"scope": "window"
+				},
 				"dart.evaluateGettersInDebugViews": {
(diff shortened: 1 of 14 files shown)


DanTup:
@pq @devoncarew FYI - this is how this looks so far. I haven't been able to add links in the tooltips because of a VS Code bug, and since we can't show banners, this may be the only entry point (besides potentially external buttons on the website if we choose to do that) we can ship for now.

I'm not sure how we want to gate this - right now it uses the master-docs hostname (the live docs won't have the index file until the next stable release?) and assumes your local Flutter is able to create the samples (I think v1 was able too, so I'm not too concerned about that). Are we happy to just ship using master-docs and then updating to docs after the next stable release?

devoncarew:
I think new users may have trouble determining which option to select (in your first screenshot) - `Flutter: Create Sample Project`, or `Flutter: New Project`. To simplify the choice, I think we'd want `Flutter: New Project` to sort first, and to modify the text for the other to help describe what it does. Perhaps:

`Flutter: New Project` and `Flutter: New Project From Documentation Sample` (or `Flutter: New Project From Doc Sample`)?

Thats a very long action name - not sure how to trim it down and still convey what we want -

DanTup:
What about calling them:

- Flutter: Create (New/Blank/Empty?) Project
- Flutter: Create Sample Project

I don't think we need to call out they're from the docs as they work equally well if the user just thinks they're samples shipped in the Flutter SDK.

devoncarew:
I assume that `Flutter: New Project` creates the `flutter create` `app` template? If so, I think it's useful to call out where the samples are from. Otherwise I'd think that `Flutter Create` would be an empty project and `Create Sample Project` one of the templates. `Flutter: New Project From Doc Sample` has less ambiguity.

DanTup:
> I assume that `Flutter: New Project` creates the flutter create app template? 

It doesn't pass any args, so whatever is default (which I suspect is app!).

> Otherwise I'd think that Flutter Create would be an empty project

I was considering the app sample to essentially be an empty project, but maybe that was misguided.

So shall we go with:

- `Flutter: New Project`
- `Flutter: New Project From Doc Sample` (or `From Docs`?)

?

As for shipping - I'm wondering if I should just fetch the snippets at startup, and then only enable the command if we got a list from the server. That would make it magically appear when the next stable Flutter rolls out (and we can add a setting to override the host to mater-docs for testing/previewing).

devoncarew:
> So shall we go with:
> 
> * `Flutter: New Project`
> * `Flutter: New Project From Doc Sample` (or `From Docs`?)

Sounds good!

> As for shipping - I'm wondering if I should just fetch the snippets at startup, and then only enable the command if we got a list from the server. That would make it magically appear when the next stable Flutter rolls out (and we can add a setting to override the host to mater-docs for testing/previewing).

This is great - it would automatically start working once that file is available.

DanTup:
> This is great - it would automatically start working once that file is available.

Only thing that just occurred to me is that this way won't allow the command to show when the user doesn't have a Dart project open - we can't run code until the extension activates. So we need to decide whether we want the command to appear when the user doesn't have a Dart project open (which is what we do for `New Project` so they can just open VS Code and use it) but it means if they invoke it and the snippets aren't live, we'll pop up an error message.

Maybe the best thing is to ship it so that it only appears if the doc host is set (overridden) and with no default (so it won't show for anyone) and then I'll just push out an update that sets the default to the live docs site as close to the next stable release as possible.

This way we get the best of both worlds - it "appears" when it's ready (albeit because of a manual update) and then after that point it's always visible for everyone.

DanTup:
Reverted the changes and the build passed... I'll work through restoring them half-at-a-time to try and isolate which one(s) are breaking things.

DanTup:
Landing this work, but setting the milestones to vNext as this functionality won't be visible until it's enabled with the hostname.

DanTup:
> but setting the milestones to vNext as this functionality won't be visible until it's enabled with the hostname.

Actually, I'm going to copy what @pq is doing and embed a fallback list that matches the current samples (used if the response is a 404) so that this can be used when we ship this version.

PR #1466, opened 2019-02-16
Avoid activation failures due to link file errors

tsangint:
test

Reproduce step:
```
// In the project folder's primary directory
touch ln_origin.txt 
ln -s ln_origin.txt ln_linked.txt
rm ln_origin.txt
// open VSCode and activate Dart-Code(Flutter plugin)
```

If there is a soft link file, but the original file is deleted, `fs.statSync`will throw an error and cause the activation to fail.


DanTup:
Good spot! Thanks for the fix! 👍

PR #1499, opened 2019-03-02
Updates to debugger

mezoni:
Please accept new implementation of breakpoints service.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -443,7 +443,7 @@ export class DartDebugSession extends DebugSession {
 			: formatPathForVm(source.path);
 
 		try {
-			const result = await this.threadManager.setBreakpoints(uri, breakpoints);
+			const result = await this.threadManager.setBreakpoints(source.path, breakpoints);
 			const bpResponse = [];
 			for (const bpRes of result) {
 				bpResponse.push({ verified: !!bpRes });
--- a/src/debug/dart_debug_protocol.ts
+++ b/src/debug/dart_debug_protocol.ts
@@ -385,6 +385,19 @@ export class ObservatoryConnection {
 		this.callMethod("streamListen", { streamId });
 	}
 
+	public addBreakpoint(isolateId: string, scriptId: string, line: number, column?: number): Promise<DebuggerResult> {
+		let data: {
+			isolateId: string,
+			scriptId: string,
+			line: number,
+			column?: number,
+		};
+		data = { isolateId, scriptId, line };
+		if (column)
+			data.column = column;
+		return this.callMethod("addBreakpoint", data);
+	}
+
 	public addBreakpointWithScriptUri(isolateId: string, scriptUri: string, line: number, column?: number): Promise<DebuggerResult> {
 		let data: {
 			isolateId: string,
--- a/src/debug/threads.ts
+++ b/src/debug/threads.ts
@@ -2,8 +2,8 @@ import { Thread, ThreadEvent } from "vscode-debugadapter";
 import { DebugProtocol } from "vscode-debugprotocol";
 import { logError } from "../utils/log";
 import { DartDebugSession, InstanceWithEvaluateName } from "./dart_debug_impl";
-import { DebuggerResult, VMBreakpoint, VMInstanceRef, VMIsolate, VMIsolateRef, VMResponse, VMScript, VMScriptRef } from "./dart_debug_protocol";
-import { LogCategory, PromiseCompleter } from "./utils";
+import { DebuggerResult, VMBreakpoint, VMInstanceRef, VMIsolate, VMIsolateRef, VMLibrary, VMResponse, VMScript, VMScriptRef } from "./dart_debug_protocol";
+import { formatPathForVm, LogCategory, PromiseCompleter } from "./utils";
 
 export class ThreadManager {
 	public nextThreadId: number = 0;
@@ -124,7 +124,7 @@ export class ThreadManager {
 
 		for (const thread of this.threads) {
 			if (thread.runnable) {
-				const result = thread.setBreakpoints(uri, breakpoints);
+				const result = thread.setBreakpointsEx(uri, breakpoints);
 				if (!promise)
 					promise = result;
 			}
@@ -230,6 +230,107 @@ export class ThreadInfo {
 		return Promise.all(removeBreakpointPromises).then((results) => {
 			return [].concat.apply([], results);
 		});
+	}
+
+	public async setBreakpointsEx(uri: string, breakpoints: DebugProtocol.SourceBreakpoint[]): Promise<VMBreakpoint[]> {
+		// Remove all current bps.
+		await this.removeBreakpointsAtUri(uri);
+		this.vmBps[uri] = [];
+
+		let fileUri: string;
+		if (uri.indexOf("\\") === -1) {
+			fileUri = "file://" + uri;
+		} else {
+			fileUri = "file:///" + uri.split("\\").join("/");
+		}
+
+		const debugSession = this.manager.debugSession;
+		const packageUri = debugSession.packageMap
+			? debugSession.packageMap.convertFileToPackageUri(uri) || formatPathForVm(uri)
+			: formatPathForVm(uri);
+
+		const observatory = this.manager.debugSession.observatory;
+
+		// Warm up isolate.
+		// Not a good place but it guarantees 100% warm-up of the isolate and
+		// this procedure are not so expensive.
+		// Preload the root library script guarantees that the root of the
+		// script tree will be initialized.
+		const isolateId = this.ref.id;
+		const isolate = (await observatory.getIsolate(isolateId)).result as VMIsolate;
+		const rootLib = (await observatory.getObject(isolateId, isolate.rootLib.id)).result as VMLibrary;
+		const rootScript = (await observatory.getObject(isolateId, rootLib.scripts[0].id)).result as VMScript;
+
+		const libraries = await Promise.all(isolate.libraries.map(async (libraryRef) => {
+			try {
+				const library = (await observatory.getObject(isolateId, libraryRef.id)).result as VMLibrary;
+				return library;
+			} catch (e) {
+				logError(e, LogCategory.Observatory);
+				return undefined;
+			}
+		}));
+
+		const scriptRefs: VMScriptRef[] = [];
+		for (const library of libraries) {
+			if (library !== undefined) {
+				for (const scriptRef of library.scripts) {
+					scriptRefs.push(scriptRef);
+				}
+			}
+		}
+
+		const scripts: VMScript[] = await Promise.all(scriptRefs.map(async (scriptRef) => {
+			try {
+				if (scriptRef !== undefined) {
+					const script = (await observatory.getObject(isolateId, scriptRef.id)).result as VMScript;
+					return script;
+				}
+
+				return undefined;
+			} catch (e) {
+				logError(e, LogCategory.Observatory);
+				return undefined;
+			}
+		}));
+
+		const foundScripts = scripts.filter((script) => {
+			if (script !== undefined) {
+				return script.uri === fileUri || script.uri === packageUri;
+			}
+
+			return false;
+		});
+
+		// Possibly Dart VM contains the bug because Dart debugger cannot add
+		// breakpoint even if the Dart VM creates more then one script
+		// reference for differently referenced (imported) the same source
+		// code.
+		// This caused because above mentioned script references will have more
+		// then a single canonicalized url and it is considered as ambiguity,
+		// although in reality this is not quite so.
+		// Most likely this can be described as “by the design,” but, in
+		// principle, it has a solution that is slightly more complicated than
+		// the very simple.
+		return Promise.all(
+			breakpoints.map(async (bp) => {
+				try {
+					if (foundScripts.length !== 0) {
+						const result = await this.manager.debugSession.observatory.addBreakpoint(this.ref.id, foundScripts[0].id, bp.line, bp.column);
+						const vmBp: VMBreakpoint = (result.result as VMBreakpoint);
+						this.vmBps[uri].push(vmBp);
+						this.breakpoints[vmBp.id] = bp;
+						return vmBp;
+					} else {
+						return undefined;
+					}
+
+				} catch (e) {
+					logError(e, LogCategory.Observatory);
+					return undefined;
+				}
+			}),
+		);
 	}
 
 	public async setBreakpoints(uri: string, breakpoints: DebugProtocol.SourceBreakpoint[]): Promise<VMBreakpoint[]> {


mezoni:
I am sorry for my source code style because I never programmed in TypeScript.
If you need the explanations about this algorithm and whyit is so but not otherwise then I am ready to answer on your questions.
This implementation based on Dart VM protocol funtionality which are based on the Dart VM Debugger features and specifications.

mezoni:
Hello, Danny.
I deleted most of my comments.
I am sorry.
And I found that the algorithm can be simplified (reduced the overhead) and it will work (but need test it on deffered loaded libraries).
In principle, it can be optimized (by using cashing) to reduce the overhead much more but this will a bit complicate it..
But new simplified version already has no overhead.

mezoni:
New version.

```dart
public async setBreakpointsEx(uri: string, breakpoints: DebugProtocol.SourceBreakpoint[]): Promise<VMBreakpoint[]> {
		// Remove all current bps.
		await this.removeBreakpointsAtUri(uri);
		this.vmBps[uri] = [];

		let fileUri: string;
		if (uri.indexOf("\\") === -1) {
			fileUri = "file://" + uri;
		} else {
			fileUri = "file:///" + uri.split("\\").join("/");
		}

		const debugSession = this.manager.debugSession;
		const packageUri = debugSession.packageMap
			? debugSession.packageMap.convertFileToPackageUri(uri) || formatPathForVm(uri)
			: formatPathForVm(uri);

		const observatory = this.manager.debugSession.observatory;
		const isolateId = this.ref.id;
		const isolate = (await observatory.getIsolate(isolateId)).result as VMIsolate;
		const rootLib = (await observatory.getObject(isolateId, isolate.rootLib.id)).result as VMLibrary;
		const rootScript = (await observatory.getObject(isolateId, rootLib.scripts[0].id)).result as VMScript;

		const libraries = await Promise.all(isolate.libraries.map(async (libraryRef) => {
			try {
				const library = (await observatory.getObject(isolateId, libraryRef.id)).result as VMLibrary;
				return library;
			} catch (e) {
				logError(e, LogCategory.Observatory);
				return undefined;
			}
		}));

		const foundScripts: VMScriptRef[] = [];
		const scriptRefs: VMScriptRef[] = [];
		for (const library of libraries) {
			if (library !== undefined) {
				for (const scriptRef of library.scripts) {
					if (scriptRef.uri === fileUri || scriptRef.uri === packageUri) {
						foundScripts.push(scriptRef);
					}
				}
			}
		}

		return Promise.all(
			breakpoints.map(async (bp) => {
				try {
					if (foundScripts.length !== 0) {
						const result = await this.manager.debugSession.observatory.addBreakpoint(this.ref.id, foundScripts[0].id, bp.line, bp.column);
						const vmBp: VMBreakpoint = (result.result as VMBreakpoint);
						this.vmBps[uri].push(vmBp);
						this.breakpoints[vmBp.id] = bp;
						return vmBp;
					} else {
						return undefined;
					}

				} catch (e) {
					logError(e, LogCategory.Observatory);
					return undefined;
				}
			}),
		);
	}
```

P.S.
I borrowed ideas from a such libraries as the `vm service client` and `dart observatory`.
They work the same way.

DanTup:
@mezoni Sorry for the delay, I've been slowly catching up with things but hadn't had chance to go through your comments or this PR yet.

At a glance though, this change seems to come down to identifying scripts by *either* file URI or package URI:

```ts
if (scriptRef.uri === fileUri || scriptRef.uri === packageUri) {
	foundScripts.push(scriptRef);
```

This is what https://github.com/dart-lang/sdk/issues/35859 is about - I'd really like the VM to handle this rather than the debugger. If the VM *always* handles package-URIs then we wouldn't need to do anything like this. This change looks like it will introduce many additional round trips - or require caching that may be complicated to keep in-sync (as I understand it, the loaded scripts may differ per-isolate, and change during runtime).

The SDK issue is blocked on another issue (https://github.com/dart-lang/sdk/issues/36023) which appears to be important for some other work too, so I expect that will be resolved. I would like to give it some time before considering putting a workaround in to Dart Code.

That said, I definitely want to have a fix for this, so if the SDK does not fix it, we'll definitely do something. I will try to add some tests for things that aren't covered (like deferred-loaded scripts) in the meantime, so we can be confident that we don't introduce more missed-breaking issues while fixing (or in future).

mezoni:
>> This change looks like it will introduce many additional round trips - or require caching that may be complicated to keep in-sync

Oh, no. No roundtrips or caching required.
No rountrips because this is algorithm from Dart observatory.
They also lookup the script in all libraries and after that add the the breakpoint to this script by calling `foundScript.addBreakpoint()`.
This works in 100% cases.
Please look at this source from Dart runtime observatory debbuger `debugger/debugger_location.dart`.

They, as and my algorithm or they algorithm :-), always, before to add the breakpoint, perform the same operations. Yes, iterate over all libraries to ensure that script exists and this script can be used for adding breakpoint to it.

This is a part of the procedure of adding breakpoint in web based debugger.
I embedded this procedure into your `addBreapoint` procedure.

```dart
static Future<List<Script>> _lookupScript(Isolate isolate, String name,
      {bool allowPrefix: false}) {
    var pending = <Future>[];
    for (var lib in isolate.libraries) {
      if (!lib.loaded) {
        pending.add(lib.load());
      }
    }
    return Future.wait(pending).then((_) {
      var matches = <Script>[];
      for (var lib in isolate.libraries) {
        for (var script in lib.scripts) {
          if (allowPrefix) {
            if (script.name.startsWith(name)) {
              matches.add(script);
            }
          } else {
            if (name == script.name) {
              matches.add(script);
            }
          }
        }
      }
      return matches;
    });
  }
```

>> as I understand it, the loaded scripts may differ per-isolate

No This is impossible. Need still to lookup the script each time (what, in principle, does the Dart observarory debugger) before to add each breakpoint (please, see above code).
The caching of (isolate based) data does not required because this operation (lookup of the script) is very fast. I tested this but you can perform your own perfomace tests if need this.

>> and change during runtime

They cannot be changed during runtime. But they will remains not in the loaded state when isolate not in the running state.
The tree of srcipts will be constructed after the isolate swithes to the running state but not when it only created.
It think this happens because Dart VM compiles the code lazy.
This requires to switch isolate in running state.

I found this behavior (or even I can say an algorithm) in tests in `vm_sercie_lib`. They never use (as you use) `addBreakpointWithScriptUri` in their test. They always use (as and my implementation) `addBreakpoint` because this call will be always adressed to the `real script` which was be already loaded by Dart VM and the script `uri` already specified and well known at this moment.

Also Dart VM Service Protocol 3.14 desribed this.

```dart
// The root library for this isolate.
  //
  // Guaranteed to be initialized when the IsolateRunnable event fires.
  @Library rootLib [optional];
```
When the root lib is loaded then all intermadiate code (or even still a token strems) of the script will be included into the scripts tree. As I already mentioned that does not need to tell the debigger to load the each script (call `script.load()`) because the `ScriptRef` already contains required information (the uri of scripts).

>> The SDK issue is blocked on another issue (dart-lang/sdk#36023) which appears to be important for some other work too, so I expect that will be resolved. I would like to give it some time before considering putting a workaround in to Dart Code.

This because they already implemented the debugger which based on the above mentioned approach. I mean the requirement to lookup (and ensure) the script before to add breakpiont to this script.
This is not an issue at all as you can understand. At least for the people whom was delevoped Dart Observatory Debugger because they know how Dart VM works in runtime and how to interact with Dart VM correctly.


>> That said, I definitely want to have a fix for this, so if the SDK does not fix it, we'll definitely do something. I will try to add some tests for things that aren't covered (like deferred-loaded scripts) in the meantime, so we can be confident that we don't introduce more missed-breaking issues while fixing (or in future).

Yes. Try to test, look at the Dart VM Obseratory Debugger implementation. Compare my code and Goole code.
And ask me the questions if you cannot understand some points. :-)

P.S.
Swithching the isolate into the running state does means to tell the isolate to run immediatelly.
This tell the Dart VM (at least) to prepare the source code representations (kernel dill?, roken stream?) to be ready for compilation. After when Dart VM already in running mode (please remember that all Dart test of the debbuger recommends to `loadRunnable` isolate before to add breakpoints) then recall this method again (`loadRunnable`) still returns the appropriate instance of the VM isolate.

mezoni:
>> The tree of srcipts will be constructed after the isolate swithes to the running state but not when it only created.

Inaccuracy.
The tree of srcipts will be constructed after the script of the root library will be loaded.
This requires to load the isolate, the load root library and it first script (main).
I don't know why this so but currently it only works this way.
Otherwise, some scripts can remains as `not loaded` and they will be absent in this tree (it will impossible to lookup them).
Please, ask the Dart developers if want to know the details.

mezoni:
Everything you need to understand:
- This is based on my research of the current Dart SDK source code and tools for debugging.
- This is based on my idea to use the same approach in the process of implementing debugger service on the remote client.

mezoni:
I can also guarantee that my implementation is operational, although it is slightly different from the original.
This is because different environments and programming languages are being used.

mezoni:
>>  At a glance though, this change seems to come down to identifying scripts by either file URI or package URI:

Not come down but uses different approach which you can imagine itself.
My implementation is  the original implementation from Google.
But they lookup by name when iterate over the libraries but I recomment to check the library uri.
This is the same. Difference only in that how to identify the script.
Of course, possible to use script name but this has drawback - need always to load the script. This is a 101% fall free approach but it requires to perform a bit more work.
Maybe it is required (ask Dart delevopers) why they do not use the `ScripRef` but always requests the `Script` instance.
I think that they more correct and my original algorithm uses the same approach but I found that load the `main` script enogh.
Maybe  I am wrong and need load all scripts because some script can remains in `not loaded` state?

I recommend try to use a more lightweight version but never forget about the full version with `script.loadByRef()`.

mezoni:
If you want I can reimplement it to be more close to original implementation.
Instead of use:

```dart
if (scriptRef.uri === fileUri || scriptRef.uri === packageUri) {
	foundScripts.push(scriptRef);
```

I will use the original:
```dart
if (script.name.startsWith(name)) {
              matches.add(script);
            }
```

But it will be a slightly more complicated implementation. Use the `name` instead of `uri`.
And this will not be my original idea (which currently based on your idea and orginal Google implementation) but based only on the Google implementation.

mezoni:
>> At a glance though, this change seems to come down to identifying scripts by either file URI or package URI:

Oh. Remember the big difference:
First algorithm (common mistake)
- Call `addBreakpointWithScriptUri(fileUri)`
- Also call `addBreakpointWithScriptUri(packageUri)` for a chance

Second algorithm
- Lookup the script in the libraries
- If script was found then
- call `script.addBreakpoint(fileUr)` only if script uri sheme is `file://`
- Or call `script.addBreakpoint(fileUr)` only if script uri sheme is `package://`

mezoni:
Also can be a situation when more than one script will be found.
Yes. This is possible but Dart SDK does not supports this.

Dart observatory implementation.

```dart
else if (scripts.length == 1) {
        return new DebuggerLocation.file(scripts[0], line, col);
      } else {
        // TODO(turnidge): Allow the user to disambiguate.
        return new DebuggerLocation.error(
            "Script '${scriptName}' is ambiguous");
      }
```
I do not agree that this is always is ambiguous and I have a proof of that but the current Dart VM does not support this and this can be considered ambiguous.

DanTup:
Apologies for not responding to this for a while. I'd really like to see how the VM work progresses before making a decision on how we fix this. 

> Oh, no. No roundtrips or caching required.

This implementation adds a lot of round trips between the debug adapter and the the VM service. For every URI that has a breakpoint, it fetches every library, and then for every library, every script. It does this repeatedly, even though the results will be the same for breakpoints in the same thread.

If we do decide to implement a workaround, I think restoring the original behaviour (multiple URIs) is preferable to this (see f80d63360fc49f836687364ec158a09f768eb2d0). However, I'd still like to give the VM team time to look at/respond to the issues raised to avoid implementing a sub-optimal fix if it's not required.

mezoni:
>> This implementation adds a lot of round trips between the debug adapter and the the VM service. For every URI that has a breakpoint, it fetches every library, and then for every library, every script. It does this repeatedly, even though the results will be the same for breakpoints in the same thread.

This implementation is the same implementation as from Google Dart developers.
It is correct that you declare the following?
`As for me, this is bad implementation and it doesn't matter that this is how the debugger included in the Dart SDK (the so-called Observatory) works.`

>> If we do decide to implement a workaround

Why need a workaround? Why not implement everything the way it was done by Google developers?
Why do you always transfer the conversation to workaround?
Why original Dart Debugger works as expected without a workaround? You know that I mean. I mean the algorithm which used in Observatory (the same I implemented for your software).
That is, in your opinion it turns out that the developers from Google developed the wrong debugger, which is based on the wrong algorithms?
I am sorry but I cannot understand you and the course of your thoughts.

mezoni:
>> This implementation adds a lot of round trips between the debug adapter and the the VM service. For every URI that has a breakpoint, it fetches every library, and then for every library, every script. It does this repeatedly, even though the results will be the same for breakpoints in the same thread.

Before making such statements could you answer a few questions?
- Did you measure the performance of this algorithm?
- What are the measurement results? How long does it take to add 10, 100, 100 breakpoints on a project with 100, 1000, 10000 libraries with 10, 100 running isolates?
- What decision did you make based on the tests performed

mezoni:
>> This implementation adds a lot of round trips between the debug adapter and the the VM service. For every URI that has a breakpoint, it fetches every library, and then for every library, every script. It does this repeatedly, even though the results will be the same for breakpoints in the same thread.

The same work is done by the VM debugger (с++)  when adding a breakpoint (fetches every library, and then for every library, every script).
https://github.com/dart-lang/sdk/blob/970390a36e03df8fdce9864796bc1753b7b0d18e/runtime/vm/debugger.cc#L3003

```c++
for (intptr_t i = 0; i < libs.Length(); i++) {
    lib ^= libs.At(i);
    script = lib.LookupScript(script_url);
    if (!script.IsNull()) {
      scripts.Add(script);
    }
  }
```
I know another way. More lightweight. I'll check it out later.
It is similar to your old way, but it is more correct because it is based on Dart VM service protocal API and it are fail free.

DanTup:
I understand your frustration, but I've posted my thoughts both in the original issue and above. My preference is to have a simple clean fix in the VM. If that can't happen in a timely fashion, I will implement a workaround - most likely restoring the original code that sent each breakpoint twice, because this solution does not perform well enough.

Here's my answers to the final few questions above, but I'm not going to keep going back and forth on this. 

> This implementation is the same implementation as from Google Dart developers.
> It is correct that you declare the following?
> > As for me, this is bad implementation and it doesn't matter that this is how the debugger included in the Dart SDK (the so-called Observatory) works.

You are not comparing like-for-like. In Observatory, the code is executed when the user types something like "break file.dart 123" into an input box. It's taking user input and therefore has no control over the format. It's doing it for a single breakpoint, therefore performance is not critical - it does not get repeated over and over in a loop.

> Why need a workaround? Why not implement everything the way it was done by Google developers?
> Why do you always transfer the conversation to workaround?

As mentioned above, Observatory is not doing the same thing.
The VM developers are fixing this in the VM, therefore adding code to avoid the issue here is a workaround.

> Did you measure the performance of this algorithm?

Yes. I ran your code using `flutter_gallery` and it resulted in tens of thousands of additional lines logged just during app startup (each round trip logs two lines, the request and response). It took in the region of 10-15 seconds IIRC to run the loops when adding/modifying breakpoints. The log file was also so large that VS Code refused to open it from the extension folder directly.

I'm sorry it doesn't get you a fix faster, but with the possibility of getting a fix for-free with no performance implications (even the original workaround of sending the paths multiple times does that) it doesn't make sense to do something like this.

DanTup:
> The same work is done by the VM debugger (с++) when adding a breakpoint (fetches every library, and then for every library, every script).

Enumerating these structures inside the VM is a *very* different prospect to doing it over a JSON-RPC API.

Please read my comments above. There are definitely ways in which this could be improved (like not fetching the same libraries/scripts for every breakpoint, for every isolate) however as I've said before - I'm holding out for a VM fix first, and if one doesn't come it makes more sense to restore the original behaviour of just sending each breakpoint twice. It's unlikely I would merge a PR that adds all these round trips unless there is no other fix.

mezoni:
Yes. I understand. Thank you for answers.

mezoni:
>> Enumerating these structures inside the VM is a very different prospect to doing it over a JSON-RPC API.

Yes. 

>> of just sending each breakpoint twice

You can send the request to add breakpoint only once.
But need only to know the correct uri of source file.
I know how to do that. I think that you also know.
But I need to check whether this will work or not.
Hint:
Fast lookup. So called, cold lookup using VM service.

mezoni:
It's done.
Now I modified the algorithm and reduced the implementation to the following:
- Warm up the isolate only once at the first request to add breakpoints to isolate
- Single request to VM service per source file instead of hundreds or even a thousands request when adding the breakpoint in source file (`DebugProtocol.SourceBreakpoint[]`).
- Used only single resolved uri as required

All works fine.

mezoni:
Yeah. The work took about 5-7 minutes.
Now the load when adding breakpoints is minimal.

>> My preference is to have a simple clean fix in the VM. 

You still prefer to wait for fix?

mezoni:
The command `Dart: Capture logs` tracks that only single additional remote method invocation has been added to the log records. So now we can assume that new algorithm does not create overhead.

DanTup:
> The command `Dart: Capture logs` tracks that only single additional remote method invocation has been added to the log records. 

Did you push some new changes? I didn't see any new changesets here so wasn't able to review.

> You still prefer to wait for fix?

My preference hasn't changed - I still think this is best fixed in the VM. If it can't be fixed there, then I think going back to sending breakpoints twice is a simple workaround (though admittedly I haven't seen your new implementation).

Anyway, hopefully this will not matter, as https://github.com/dart-lang/sdk/issues/35859 has been closed as fixed! :) I will aim to do some testing of the fix today.

DanTup:
I have tested the VM changes, and breakpoints in `part` files are now hit with no changes to Dart Code :-)

If you need the fix now, you could use a [nightly build](https://gsdview.appspot.com/dart-archive/channels/be/raw/latest/sdk/) until the change reaches a dev/stable builds.

If you find any other issues (including missed breakpoints), please do let me know.

PR #1508, opened 2019-03-06
Run downloads/setups of Dart/Flutter in parallel during CI builds

DanTup:
A lot of time is spent setting things up, so running this in parallel may speed things up (this will need to be compared with some builds).


DanTup:
Ditching this in favour of using CI caches, which seem faster and much less error prone (and don't result in significantly out-of-order text in the CI online log view).

PR #1515, opened 2019-03-07
Add a notification about DevTools

### LINKED ISSUE #1503: Show a notification about DevTools when a Flutter debug session launches
Some people don't read the release notes so may not know about DevTools. It'd be nice to show a message with a button to launch them when you launch a Flutter debug session.

We should limit how often it appears though, and have a Don't Show Again button (maybe show only once per session, for the first 10(?) sessions, but with Don't Show button always there).

DanTup:
@devoncarew @jacob314 WDYT about this to help people find DevTools? It'll only appear once per 20 hours, and only up to 10 times maximum (and obvious, never again if you click Don't Show Again).

screenshot 2019-03-07 at 4 54 04 pm

Fixes #1503.

### DIFF
--- a/src/commands/debug.ts
+++ b/src/commands/debug.ts
@@ -6,7 +6,8 @@ import { CoverageData, PromiseCompleter } from "../debug/utils";
 import { FlutterServiceExtension, FlutterServiceExtensionArgs, FlutterVmServiceExtensions, timeDilationNormal, timeDilationSlow } from "../flutter/vm_service_extensions";
 import { PubGlobal } from "../pub/global";
 import { DevTools } from "../sdk/dev_tools";
-import { fsPath, getDartWorkspaceFolders, openInBrowser, Sdks } from "../utils";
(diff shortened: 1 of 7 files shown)


jacob314:
Looks good to me.
I'd also personally like an option to launch DevTools every single time I run an app or at least a button I can click to launch it from VSCode instead of having to remember to open the command pallet.

DanTup:
> I'd also personally like an option to launch DevTools every single time I run an app or at least a button I can click to launch it from VSCode instead of having to remember to open the command pallet.

VS Code doesn't have many buttons, and the command palette is pretty much the entry point for *everything* so there isn't really anywhere convenient to put a button (I don't think showing a notification on every debug session launch would be nice).

We do have an additional button for it on the touchbar:

Touch Bar Shot 2019-03-11 at 9 59 51 am 3

We can always come back and change it (for ex. it might make more sense to push it more when it's not considered preview) or add more ways to launch it in future versions.

PR #1531, opened 2019-03-14
Add MaterialApp and CupertinoApp

rubensdemelo:
Some times we just want to play around. Write some code, test, change, debug, re-write, whatever... and the easiest way to do that is starting from scratch. These snippets can help.


DanTup:
Thanks for the PR! Though I'm a little unsure how these are likely to be used. For creating projects, there's a `Flutter: New Project` command which includes a `pubspec.yaml`, runs `pub get` etc. for you.

If you already have a project, then it seems like you'd probably already have things like a `main` method and project scaffold set up - in which case it seems like it may be easier just to replace your main widget (which may be in its own file, so you don't need to re-create what's here).

What does your workflow look like for you to use these frequently?

Thanks!

rubensdemelo:
Hi Dan,

When a project already exists, this isnt helpful. My point is when we would like to "play" with flutter.
Start from scratch, write some code, test/debug anything, copy and paste. 
Just would like to delete my whole code and start again (ctrl-A + delete). 
Create a new project every time isnt a good option. 
But if you think that isnt fit to extension, no worries. 
Im using this in my setup and help me a lot. Improve my dev experience.

DanTup:
Thanks for the info - I guess I wondered how likely people would have all that code in a single file and want to replace it the whole lot (versus just deleting a single widget created inside the main `App`).

I haven't concluded this isn't a good fit, but I would like to be more sure that it's generally useful to many. End users can't hide the snippets we add here so I don't want to end up with a huge collection of snippets that have very limited use. I don't really have a strong opinion on it because my Flutter usage is relatively limited. I'll see if I can find some other opinions though, and if this seems like it might be a common thing to do, I'll merge it in.

Thanks!

PR #1537, opened 2019-03-18
Read snippets from Flutter if supported (to reduce mismatched versions between master/stable)

### LINKED ISSUE #1529: Switch to flutter create --list-samples (to reduce mismatched versions between master/stable)

DanTup:
Fixes #1529.

### DIFF
--- a/src/commands/sdk.ts
+++ b/src/commands/sdk.ts
@@ -460,12 +460,11 @@ export class SdkCommands {
 		}
 
 		// Fetch the JSON for the available samples.
-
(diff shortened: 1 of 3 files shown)


DanTup:
@pq FYI - this is the change to read snippets from `flutter create --list-samples foo`. I ditched the embedded array and left the existing web code as a fallback, but may remove the web code after a stable Flutter release includes this command (in which case I'll just show the error that you need a newer Flutter SDK when `supportsFlutterCreateListSamples == false`).

PR #1686, opened 2019-05-06
Add a link to the Flutter Q2 survey

DanTup:
@pq

### DIFF
--- a/src/user_prompts.ts
+++ b/src/user_prompts.ts
@@ -73,7 +73,6 @@ export function showFlutter2019Q2SurveyNotificationIfAppropriate(context: Contex
 	if (lastShown && now - lastShown < fortyHoursInMs)
 		return false;
 
-
 	// Work out the URL and prompt to show.
 	let clientID: string | undefined;
 	try {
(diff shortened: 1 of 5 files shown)


DanTup:
(don't worry about the failed builds, I landed some tests that will fail until a few components release over the coming days, they'll be green before this merges :-))

src/user_prompts.ts
```diff
@@ -41,7 +41,77 @@ export function showUserPrompts(context: Context, workspaceContext: WorkspaceCon
...
+
+	// Don't show this notification if user previously said not to.
+	if (doNotShow)
+		return false;
+
+	// Don't show this notification if we've shown it in the last 40 hours.
+	if (lastShown && now - lastShown < fortyHoursInMs)
+		return false;
+
+
```
pq: I think this extra space is problematic for `tslint`?

DanTup: Ha, good spot! One day I'll figure out why tslint issues seem to be unreliable showing in the editor... I'll fix this tomorrow when the other stuff is done, so it should all go green. Thanks!

DanTup:
Builds all green except a flake (timeout) in the AppVeyor master-merge one, so merging this.

PR #1693, opened 2019-05-08
Survey fixes

DanTup:
- Fix location of settings on Windows (see https://github.com/dart-lang/usage/blob/5ad6959611dd2f2c504c7a7050a5f770a42203f9/lib/src/usage_impl_io.dart#L63).
- Also don't show if there are no Flutter projects (Dart-only users aren't interested in Flutter surveys)

@pq FYI. I don't think you have this bug (your PR wasn't reading the file directly), but apparently I did (and apparently my subconscious mind can find bugs while I'm making tea that my conscious mind can't while I'm writing the code 🙈).

### DIFF
--- a/src/user_prompts.ts
+++ b/src/user_prompts.ts
@@ -45,8 +45,10 @@ export function showUserPrompts(context: Context, workspaceContext: WorkspaceCon
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


DanTup:
Not sure about 🍵, I think I need some 🍺  for Travis :)

(I'm merging this because it's all broken on master too, so makes sense to just fix it in one branch given how long the runs take :-))

PR #1720, opened 2019-05-17
Add support for target-platform to Flutter launch config.

ened:
`flutter --target-platform` allows developers to choose the target architecture.

It's auto-selected but sometimes there is a definitive need to specify the field sometimes, e.g. App testing for different phones / architectures.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1277,6 +1277,17 @@
 								],
 								"default": "debug"
 							},
+							"flutterPlatform": {
+								"description": "Passes the --target-platform option to the flutter run command. Ignored on iOS.",
+								"enum": [
+									"default",
+									"android-arm",
+									"android-arm64",
+									"android-x86",
+									"android-x64"
+								],
+								"default": "default"
+							},
 							"args": {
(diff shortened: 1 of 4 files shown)


DanTup:
Thanks for the PR! Though is this not something you can just add directly to `args` in your launch configs?

ened:
Yes, it can be added to the `args`. A dropdown field is a lot easer (especially when it offers auto-completion).

DanTup:
That's true. I'm a little worried about starting down a path where we end up adding all of Flutter's command line switches in here and having to keep them up to date (and potentially deal with multiple versions). We already have some like debug/profile mode, but for those there's an intention to expose them as first class commands ("run in profile mode") to avoid having to maintain multiple configs of keep editing the file which I think makes them more valuable.

I'll think more about this when back in the office.

DanTup:
I've merged this one, though I think we might want to tweak it a little when other platforms like Web are fully supported in the main `flutter` tool depending on how they're supported.

Thanks!

PR #1774, opened 2019-06-03
add String List and Map support for highlighting

shvilek:
added String List and Map support for highlighting

### DIFF
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -309,7 +309,7 @@
 					"name": "storage.modifier.dart"
 				},
 				{
-					"match": "(?<!\\$)\\b(?:void|bool|num|int|double|dynamic|var)\\b(?!\\$)",
+					"match": "(?<!\\$)\\b(?:void|bool|num|int|double|dynamic|var|String|List|Map)\\b(?!\\$)",
 					"name": "storage.type.primitive.dart"
 				}
 			]


DanTup:
Thanks for the PR! However, I'm not sure if this change makes sense. String, List and Map are all just type names, and currently they're coloured the same as other types - with the exception of some special ones (`bool`, `num`, `int`, `double`):

Screenshot 2019-06-04 at 8 33 15 am

I think it would be strange to colour `Map` differently to other classes in the SDK like `Set`, for example.

This does raise the question of why int/double/etc. are coloured differently and what's special about them, and I don't really have an answer to that (they are special - in that they start with lowercase letters - though [it doesn't seem like](https://github.com/dart-lang/sdk/issues/1410) there's a solid reason for this). I think if we wanted to make these consistent, removing int/double/etc. would  make more sense (though I'm not sure how people would feel about it - though as far as I can tell, Android Studio does colour them all the same). What do you think?

shvilek:
oops. i forget about Set. Anyway for me it's strange that built-in types have different colors. But let's close this PR. I just looked on DartPad and there same color for int and String but different color for Map and Set .

DanTup:
Yeah, DartPad seems even more arbitrary (since it includes one PascalCased type too).

I'm sort of tempted to remote int/double/bool to make them consistent with others, it seems a bit weird (although I appreciate it's familiar so if we undo it, that'll probably also feel weird for a while).

@devoncarew @pq any opinions/thoughts on this? Android Studio seems to colour int/double/bool the same as all other types. VS Code has them treated specially and coloured with keywords. I think having them different is strange, but I think it makes more sense to un-keyword-ify those few than add more to the special set.

pq:
> I'm sort of tempted to remove (sic) int/double/bool to make them consistent with others,

That'd be my inclination too.

DanTup:
Ok, I'll do that! How does this look? I'm not too sure about `dynamic` and `void`. I think `dynamic` is definitely a type and looks weird if I colour it specially in a list of args (even though Android Studio does that) but I'm less sure about `void`? 🤔

Before:

Screenshot 2019-06-05 at 5 58 29 pm

After:

Screenshot 2019-06-05 at 5 58 43 pm

PR #1788, opened 2019-06-11
Add ide=VSCode to DevTools querystring

DanTup:
@jacob314 @terrylucas any preference on the name/value of this? Although not technically an IDE, `ide` seemed shorter than `editor`. Using a real `name=value` rather than just `&vscode=` makes it easier to just write as-is to a custom dimension if the other editors did the same.


DanTup:
@jacob314 @terrylucas let me know if you have any preferences for this value, otherwise I'll land like this for the next version.

PR #1844, opened 2019-07-09
Use "Service" instead of "_Service" for VM service protocol >= 3.22

DanTup:
@devoncarew would you mind scanning these changes? (Select **Hide whitespace changes** in the diff to avoid seeing loads of changes due to indent change!).

Fixes https://github.com/Dart-Code/Dart-Code/issues/1843.
Also see https://github.com/dart-lang/vm_service_drivers/pull/251.

### DIFF
--- a/src/extension/debug/dart_debug_impl.ts
+++ b/src/extension/debug/dart_debug_impl.ts
@@ -12,7 +12,7 @@ import { white } from "../../shared/utils/colors";
 import { config } from "../config";
 import { getLogHeader } from "../utils/log";
 import { safeSpawn } from "../utils/processes";
-import { DebuggerResult, ObservatoryConnection, SourceReportKind, Version, VM, VMClass, VMErrorRef, VMEvent, VMFrame, VMInstance, VMInstanceRef, VMIsolate, VMLibrary, VMMapEntry, VMObj, VMScript, VMScriptRef, VMSentinel, VMSourceReport, VMStack, VMTypeRef } from "./dart_debug_protocol";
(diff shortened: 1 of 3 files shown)


src/extension/debug/dart_debug_impl.ts
```diff
@@ -378,6 +392,15 @@ export class DartDebugSession extends DebugSession {
 		});
 	}
 
+	private subscribeToStreams() {
+		this.observatory.on("Isolate", (event: VMEvent) => this.handleIsolateEvent(event));
+		this.observatory.on("Extension", (event: VMEvent) => this.handleExtensionEvent(event));
+		this.observatory.on("Debug", (event: VMEvent) => this.handleDebugEvent(event));
+
+		const serviceStreamName = this.capabilities.serviceStreamIsPublic ? "Service" : "_Service";
```
devoncarew: This looks good - good to be able to switch on the service protocol version in general.

I think you'd also be good just subscribing to both streams and ignoring error results.

Perhaps add a todo: to remove the switch at some point?

DanTup: > I think you'd also be good just subscribing to both streams and ignoring error results.

That was my original plan, but making sure that if they both failed we had something in the log to debug (without always writing logs for the one that failed) was messy, so I figured I may as well do it properly (it seems likely we'll need this if changes for breakpoints across multiple debuggers happen anyway).

> Perhaps add a todo: to remove the switch at some point?

Rather than TODOs all around the place I've started periodically reviewing the Capabilities classes (of which there are now quite a few now) and chopping properties for old things out and fixing the related code appropriately. Unless they're adding complexity, I've been trying to keep them around a while though - this one seems pretty trivial (fun fact: most days Dart Code still has *hundreds* of users using SDK v1.24 🙃!)

DanTup:
The red builds are all from Flutter Web test failures.. they've always been a bit flaky on the CIs - though seems worse lately. They're all passing locally for this branch so I'm confident nothing is broken by it. I'll work on making them better when I do work to make them run on the unforked version.

Issue #1824, opened 2019-06-26
Switch to structured logs from dart:developer log()


DanTup:
We can get these logs through the service protocol and show them in the Debug Output which would give us more options in the future (for ex. filtering by source, so users can opt-in to logging from their dependencies, etc).

https://api.dartlang.org/stable/2.3.2/dart-developer/log.html

https://github.com/flutter/flutter-intellij/pull/3624

PR #1849, opened 2019-07-10
Show log events from dart:developer log() in the debug console (if enabled)

DanTup:
~~Required to do https://github.com/Dart-Code/Dart-Code/issues/1824.~~ Not so, Flutter's errors come through as `Flutter.Error` and are separate to `dart:developer`'s `log()`. This just exposes logs that would otherwise not be visible to the user.

### DIFF
--- a/package.json
+++ b/package.json
@@ -795,6 +795,12 @@
 					"description": "Whether to mark external pub package libraries as debuggable, enabling stepping into them while debugging.",
 					"scope": "resource"
 				},
+				"dart.showDartDeveloperLogs": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to show logs from dart:developer's log() function in the debug console.",
+					"scope": "resource"
+				},
 				"dart.enableCompletionCommitCharacters": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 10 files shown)


DanTup:
@devoncarew I added code to print `Logging` events, though I'm not sure how spammy this is going to be (we have no good way to filter it in VS Code here, so if lots of libraries write a lot of output, it's going to fill the users debug console).

I could write the output all to another pane (an Output channel), though we don't have any ability to link to source code in there (though I don't know that matters for these?).

devoncarew:
I think having it in-line with the other output makes sense. If things do get too verbose, we may want to add ways to toggle on and off specific channels, or hide the output altogether.

DanTup:
I was thinking that if a user pulls in some dependency that is really chatty and we don't already have some filter, their own output could get lost. There's no great way for us to build a UI to let them filter out sources (the best we can do right now is probably in the Workspace Settings JSON, excluding them by name). I have no idea how many people use `dart:developer`'s `log` function though. If it's practically nobody, then it's less urgent (if it's used a lot, I think we probably need some sort of filter before we ship this).

PR #1869, opened 2019-07-17
Add errors/stacks to dart:developer logging

### LINKED ISSUE #1824: Switch to structured logs from dart:developer log()

DanTup:
I originally missed these out. They're also async, so we need to chain future logs after this to ensure they're printed in the same order.

Screenshot 2019-07-17 at 10 53 36 am

Fixes #1824.

### DIFF
--- a/src/extension/debug/dart_debug_impl.ts
+++ b/src/extension/debug/dart_debug_impl.ts
@@ -151,13 +151,19 @@ export class DartDebugSession extends DebugSession {
 		process.on("error", (error) => {
 			this.logToUser(`${error}\n`, "stderr");
 		});
-		process.on("exit", (code, signal) => {
+		process.on("exit", async (code, signal) => {
 			this.processExited = true;
 			this.log(`Process exited (${signal ? `${signal}`.toLowerCase() : code})`);
 			if (!code && !signal)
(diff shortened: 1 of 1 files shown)


DanTup:
@devoncarew in the screenshot above, the exception text says `<collected>`. This seems to be what comes back from the VM when I call `toString()` on the exception. Is there anything we can do to avoid this? (it doesn't seem consistent, but it seems to happen almost every time).

devoncarew:
> This seems to be what comes back from the VM when I call `toString()` on the exception. Is there anything we can do to avoid this? (it doesn't seem consistent, but it seems to happen almost every time).

I don't know that there's anything we can do on the client side. cc @bkonyi; something to consider - when object refs are sent to clients, we may want some brief grace period before they can be considered for collection. A ring buffer of the last n objects sent to the client? Have objects flushed from a buffer after some short duration in the buffer?

PR #1870, opened 2019-07-17
Use Fluttered structured error events (behind preview flag)

DanTup:
@jacob314 @InMatrix @devoncarew I've made some progress on this, trying to make them look like IntelliJ's (so please let me know if they change - for ex the yellow ascii banner - so I can copy).

Two of the examples in @InMatrix's repo don't look too bad:

Screenshot 2019-07-17 at 12 58 56 pm

Screenshot 2019-07-17 at 12 59 42 pm

However the Material one looks really bad:

Screenshot 2019-07-17 at 1 00 34 pm
Screenshot 2019-07-17 at 1 00 42 pm

I think probably I'm rendering more nodes than I should (though it's not clear how I should decide which to show), but also some of it's just in the source (like the top line is like this in one big string). How should that be handled?

- [ ] Also needs tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -801,6 +801,12 @@
 					"description": "Whether to show logs from dart:developer's log() function in the debug console.",
 					"scope": "resource"
 				},
+				"dart.previewFlutterStructuredErrors": {
+					"type": "boolean",
+					"default": false,
+					"description": "Whether to enable Flutter's structured error support for improve error display.",
+					"scope": "resource"
+				},
 				"dart.enableCompletionCommitCharacters": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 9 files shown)


DanTup:
Also, there seems to be a race at startup - if an error occurs in startup code, it may be before structuredErrors has been toggled on. As far as I can tell, the service extension doesn't load until after I resumed the isolate, so I don't know if there's a way to avoid this?

devoncarew:
Some drive-by-comments:

- we'll need to adjust the `◢◤◢◤◢◤◢◤◢◤◢◤` chars (used by IntelliJ as well) to some other ansi chars; doesn't have to be for this PRR, but those chars are used by Flutter to signify layout overflow
- after a summary line, you'll want to output a single blank line
- IntelliJ is currently not printing out nodes where `type` == `ErrorSpacer`
- we don't recurse into child nodes if `style` == `shallow` (we still show local properties)

The impl. for IntelliJ is all at https://github.com/flutter/flutter-intellij/blob/master/src/io/flutter/logging/FlutterConsoleLogManager.java#L135.

devoncarew:
> As far as I can tell, the service extension doesn't load until after I resumed the isolate, so I don't know if there's a way to avoid this?

I've seen this race as well, but don't have a good solution.

DanTup:
Thanks for the notes! I had been following the IntelliJ implementation but I couldn't figure out all the differences (I suck at reading Java :-)).

I'll make those changes - and I'll update the banner to match whatever you do in IntelliJ once you've made a decision about what to do there.

PR #1891, opened 2019-07-24
Improve error display

DanTup:
Screenshot 2019-07-24 at 10 59 15 am

Screenshot 2019-07-24 at 11 12 38 am

@devoncarew I think these match your latest changes?

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -13,7 +13,7 @@ import { LogMessage } from "../shared/interfaces";
 import { PackageMap } from "../shared/pub/package_map";
 import { errorString, flatMap, throttle, uniq, uriToFilePath } from "../shared/utils";
 import { sortBy } from "../shared/utils/array";
-import { applyColor, grey } from "../shared/utils/colors";
+import { applyColor, grey, grey2 } from "../shared/utils/colors";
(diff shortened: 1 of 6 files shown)


devoncarew:
👍 looks good!

I think the only change I'm going to make to my PR at this point is to display properties before node children (rather than vice versa as I do now). This may or may not change the existing error message display - I don't know if the nodes we're currently displaying have both properties and nodes.

DanTup:
Some additional tweaks:

- Show properties before children
- Add blank line between hint/non-hint
- Highlight hints in lighter grey
- Don't render properties of Shallow nodes (were were previously only skipping children, so we had more output than IntelliJ, since it collapses them)

The only significant concern I have left is the Missing Material error. The "Here are the ancestors" section is *massive* but those nodes don't have any differences to the one above, so I can't see any way to identify them (possibly they'd need changes in Flutter?). IntelliJ probably would want to collapse them, but I think in VS Code we'd want to hide them (at least until we have some way to collapse).

Screenshot 2019-07-25 at 8 32 40 am

Screenshot 2019-07-25 at 8 33 01 am

Screenshot 2019-07-25 at 8 32 16 am
Screenshot 2019-07-25 at 8 32 27 am

DanTup:
Merging this so I can get `master` ready for a new beta build today, but will continue to iterate as things come up.

PR #2005, opened 2019-09-18
Remove the "Flutter: New Web Project" command and associated code/tests

### LINKED ISSUE #2004: Remove "Flutter: New Web Project" command
The old forked `flutter_web` project isn't maintained, and users are directed at using the unforked version. Stagehand no longer contains the template (https://github.com/dart-lang/stagehand/pull/633) so the command doesn't work anymore, and since there's no real distinction in web/not web projects it doesn't seem to be worthwhile having a specific command.

DanTup:
Fixes #2004.


richasay:
Why did you guys do this?  It is making testing for Flutter Web much more difficult.

DanTup:
@richasay the reasons for this are in the linked issue https://github.com/Dart-Code/Dart-Code/issues/2004 and were in the [release notes](https://dartcode.org/releases/v3-5/#flutter). The command was broken because it used the old way of creating a Flutter web project (which is no longer supported). The correct way is to use the standard **Flutter: New Project** command.

Note: Web support is still in preview, see the [Flutter site](https://flutter.dev/web) for more info.

richasay:
@DanTup  I guess it doesn't seem to make sense to rip one piece out rather than fix the stagehand issue.  The Flutter: New Project is not WEB friendly at all.  It creates a native device focused project with Test (not Web) dir, IOS and Android, Pubspec has none of the web dependencies, It requires a lot more manual configuration and makes "Preview" testing of web quite a but more difficult.  It's just frustrating that there was not much advanced warning that it was being eliminated.  All of the Flutter Web documentation still points the that process and so there is a lot of confusion around what was seen by many as the holy grail of Flutter promises, the trifecta of the one code to rule them all... @DanTup - I'm not blaming you... its just frustrating to get into a rhythm with an IDE and then have it breakdown.  Thanks

DanTup:
> I guess it doesn't seem to make sense to rip one piece out rather than fix the stagehand issue

It isn't really an issue to fix. Using Stagehand for Flutter web projects was always a temporary solution while web was forked - it was always going to go away in favour of `flutter create`. It may have happened sooner than I (or you) expected, but it was always going to happen and it doesn't make sense to reverse it. Web support is clearly marked as being preview, and using it is going to be bumpy for a while.

> Flutter: New Project is not WEB friendly at all. It creates a native device focused project with Test (not Web) dir

Running `flutter create` does create a `web` directory for me. You may need to upgrade Flutter (you need to be on the `master` branch, and also have enabled web support as detailed on the site linked above).

> , IOS and Android, Pubspec has none of the web dependencies, It requires a lot more manual configuration and makes "Preview" testing of web quite a but more difficult

I'm not sure which dependencies are missing, but if something isn't working right (and you've definitely on latest master) then it's worth filing issues at flutter/flutter to be looked at.

> It's just frustrating that there was not much advanced warning that it was being eliminated.

That's fair - and it caught me out too. The template was removed from Stagehand because the website now instructs to use `flutter create`. I had to remove the command because it didn't work now that Stagehand doesn't have the template. That said, moving to the Flutter version should be an improvement (and was always the long term goal).

> All of the Flutter Web documentation still points the that process

The getting started page for web (https://flutter.dev/docs/get-started/web) details the new flow (I believe that's why Stagehand removed the template), so I don't think that's entirely true.

I do understand the frustration, but I don't think there's much to do. This was always going to happen, and using early tech previews is always going to be bumpy. If you're having issues that make this worse, please do file them in flutter/flutter and I'm sure they'll be quickly resolved. The end goal is certainly not to make things more complicated :-)

Issue #1970, opened 2019-09-04
errors remain after running flutter generate


szotp:
I'm not sure if this is a problem with DartCode or some kind of analysis daemon.

1. I have a package that utilizes custom Builder, like r_flutter.
2. New person joins the team, and downloads git.
3. When they launch the project in VSCode, the code was not generated yet, so there are errors.
4. New person runs `flutter generate`, the command suceeds, but errors are still there.
5. New person executes "Reload window" and errors disappear.

Similiar issues may also happen after `flutter clean` or switching branch.

I think it would be ideal if VSCode could handle this automatically: do not perform analysis until files have been generated. 

EDIT: as requested I will divide this into 2 issues. The problem here is that error remain after running flutter generate - user needs to reload window to fix this.

DanTup:
I would expect the server to handle this automatically as the new files are generated. Are the errors the result of missing generated code, or missing packages?

Do you happen to have a small sample I could use to reproduce and test with?

> I think it would be ideal if VSCode could handle this automatically: do not perform analysis until files have been generated.

I don't think it's possible for the VS Code extension to know that there are files that need to be generated. But even if it did, I don't think it makes sense to remove all language functionality (including things like code completion) due to missing files. You might need to work on your code before you can get the generator to generate the files (for example if you have errors that prevent it).

I think there should be a better way to fix this that doesn't involve taking any functionality away when the files are missing.

szotp:
Here is the sample: [generation_sample.zip](https://github.com/Dart-Code/Dart-Code/files/3579075/generation_sample.zip)

What I had in mind was something similiar to `promptToRunPubGet` that asks user to run `flutter generate`, maybe it could be even done in the same operation?

1. User agrees to download packages.
2. After packages have been downloaded, DartCode checks if `builders:` option exists in pubspec.
3. If it exists, it runs `flutter generate` and updates the prompt message.
4. After that it restarts analysis server to hide errors.

If there are errors preventing generate from working then that's a pity, DartCode should continue working as it currently does.

DanTup:
> What I had in mind was something similiar to `promptToRunPubGet` that asks user to run `flutter generate`,

I think that's a reasonable idea (though I don't know how easily it can be detected right now - I've never used `flutter generate` :-)). Could you file a new issue to request that, and we can use this one for investigating why the errors don't go away on their own. I'd prefer to fix that (if we can) before restarting the server, since it would be much faster to not have to reinitialise everything.

[...]

PR #2033, opened 2019-10-06
Add command to reanalyze project

szotp:
This is slightly related to #1970.

Ideally it would not be needed, but I had few situations where I had to run "reload window" because analyzer was not handling updated packages, etc. 

So instead of "reload window" I could just force the analyzer to reanalyze. The additional bonus is that other extensions can also run this command, which will be useful for my extension for r_flutter.

I basically copied the implementation of `dart.openAnalyzerDiagnostics` so there are no tests or anything because the original command didn't have them.

### DIFF
--- a/package.json
+++ b/package.json
@@ -207,6 +207,11 @@
 				"category": "Dart"
 			},
 			{
+				"command": "dart.reanalyze",
+				"title": "Reanalyze",
+				"category": "Dart"
+			},
+			{
 				"command": "dart.changeSdk",
 				"title": "Change SDK",
 				"category": "Dart"
@@ -518,6 +523,10 @@
 				},
 				{
 					"command": "dart.openAnalyzerDiagnostics",
+					"when": "dart-code:dartProjectLoaded"
+				},
+				{
+					"command": "dart.reanalyze",
 					"when": "dart-code:dartProjectLoaded"
 				},
 				{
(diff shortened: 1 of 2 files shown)


DanTup:
> Ideally it would not be needed, but I had few situations where I had to run "reload window" because analyzer was not handling updated packages, etc.

The best fix here is to resolve those issues. This has come up before, but this is a workaround I would prefer not to have. If implemented, people would just use it without raising issues for the underlying bugs, and then complain that the extension is buggy and you have to keep reloading.

So, this should be considered a last resort and we should push on the other bugs first. If we hit blockers that mean they can't all be solved, then we should consider this. I don't think we should merge it unless we hit these though.

(Apologies with the delay for  https://github.com/Dart-Code/Dart-Code/issues/1970, I was out of the office for a few weeks last month so I didn't get much done.)

szotp:
Alright, I'm waiting for #1970 then, this PR can always be reopened.

szotp:
I just had another idea, `Flutter: Generate` command that does `flutter generate` and then reanalyze. Would that pass?

DanTup:
I  think a command to run `generate` makes sense (it should only be visible if you have a Flutter project open, and use the correct project root - similar to `flutter packages get` etc.).

Before adding an automatic reanalyze, I think we should try to figure out if it's an analysis server bug that it's required (it should pick up changes in files automatically, and incremental work is much faster than reanalyzing a large project).

I'm just going to do some digging with your repro from https://github.com/Dart-Code/Dart-Code/issues/1970.

Issue #2056, opened 2019-10-17
Syntax highlighting messes up when expressions contain `}`


DanTup:
```dart
void foo(String a) {
  print('${'test}test'}');
}
```

This would print the string `test}test`, but the `}` is messing up the syntax highlighting:

<img width="391" alt="Screenshot 2019-10-17 at 9 42 19 am" src="https://user-images.githubusercontent.com/1078012/66992832-6debc380-f0c2-11e9-84b8-fec031a4ae74.png">

DanTup:
This is slightly better with LSP semantic tokens, but not entirely fixed:

<img width="445" alt="Screenshot 2021-01-26 at 15 52 08" src="https://user-images.githubusercontent.com/1078012/105868832-7c779900-5fee-11eb-9603-8806624e3bdd.png">

The fix may be to remove any interpolated string expressions once Dart/Flutter stable channels fully support semantic tokens.

DanTup:
With d4d5fe8b, the interpolated expressions use the other parts of the grammar to parse so that strings are parsed correctly, rather than just looking for the next `}`.

<img width="1196" alt="Screenshot 2021-07-19 at 14 35 09" src="https://user-images.githubusercontent.com/1078012/126168295-a07d2cfb-4475-4b2c-a8d6-48351e968f1b.png">

DanTup:
PR to update the shared grammar is here:

https://github.com/dart-lang/dart-syntax-highlight/pull/19

Issue #105, opened 2016-08-16
Expressions in interpolated strings aren't coloured like simple variables


jrote1:
Would be good to have the variables a different color so that their easier to see

DanTup:
It looks like the [language syntax](https://github.com/Dart-Code/Dart-Code/blob/master/syntaxes/dart.json#L292) does know about these so this should be possible (I'm surprised it doesn't happen automatically).

TypeScript definitely highlights these differently, so we can review [theirs](https://github.com/Microsoft/vscode/blob/master/extensions/typescript/syntaxes/TypeScript.tmLanguage.json#L931).

ikhwanhayat:
I thought we already have this?

![image](https://cloud.githubusercontent.com/assets/35133/17721832/d82f33aa-645f-11e6-9332-f65952a27736.png)

Expressions inside the ${} aren't highlighted though.

IntelliSense works too! This is cool.

jrote1:
For me they are not highlighted in any scenario

ikhwanhayat:
@jrote1 Might be the color theme you use?  Mine is "Dark+ (default dark)".

DanTup:
Yeah, "Dark (Visual Studio)" doesn't seem to handle this nicely, but "Dark+" does. Not sure if I changed it, or this was default. Maybe worth reporting the inconsistency in the [Code repo](https://github.com/microsoft/vscode/issues)?

Recategorizing this as a bug as that's how it appears to the user; we should try to fix.

(@devoncarew I think we stole this file from you, so this might affect Atom too?)

DanTup:
Oh man, I just saw this sample on the Dart website:

```dart
List names = ['John', 'Sophena', 'Henrietta'];
print("${
  ((names) {
      return names[(new math.Random()).nextInt(names.length)];
  })(names)} is the most valueable member of our team");
```

Probably we'll have to settle with the whole expression being a single colour rather than nesting the colours.

DanTup:
I've made this better, but unfortunately it's not possible to make accurate using regex. We now allow lots of characters inside these expressions and still match (anything except any of `{}"'`), however it breaks down if you include nested braces or strings (the problem with strings is that you can put braces inside them).

Since JS regex (AFAIK) doesn't have support for counting braces to figure out which closing brace actually matches the expression (consider `${something('with a } in a string'){`) I don't think I can properly fix this (though if anyone has ideas, I'm all ears!).

jodinathan:
[...]

PR #2070, opened 2019-10-23
Properly highlight interpolated expressions

shroudedcode:
The way interpolated expressions are currently highlighted is quite hard to read:

Before

This annoyed me a lot, so I decided to update the Dart grammar to properly highlight the expressions within the curly braces and distinguish the beginning (`${`) and end (`}`) from the rest of the string:

After

### DIFF
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -318,14 +318,19 @@
 			"patterns": [
 				{
 					"captures": {
-						"2": {
-							"name": "variable.parameter.dart"
-						},
-						"3": {
+						"1": {
 							"name": "variable.parameter.dart"
 						}
 					},
-					"match": "\\$((\\w+)|\\{([^{}]+)\\})"
+					"match": "\\$(\\w+)"
+				},
+				{
+					"captures": {
+						"1": {
+							"name": "variable.parameter.dart"
+						}
+					},
+					"match": "\\$\\{([^{}]+)\\}"
 				},
 				{
 					"match": "\\\\.",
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks for the PR! Unfortunately this one is a bit of a minefield. There have been a few attempts to try and handle it better, but they've all had weird issues and I ended up reverted to something simpler.

Here's an example that [was on the Dart website](https://github.com/Dart-Code/Dart-Code/issues/105#issuecomment-240788864):

Screenshot 2019-10-23 at 4 37 45 pm

Since I last tried, I have made code blocks work in doc comments a little better (recursively), so I wonder if maybe a similar thing can be done here? It's possible it's too limited though - there are many open issues against VS Code asking for better ways of doing syntax highlighting (for ex https://github.com/microsoft/vscode/issues/585 and https://github.com/microsoft/vscode/issues/50140) but it's not clear if this will change anytime soon.

shroudedcode:
@DanTup I understand that, I'll give up in that case. :smile: I really hope a tool like Tree-sitter makes a breakthrough one day, parsing code with RegEx is just a mess.

DanTup:
Indeed - I've spent many hours debugging issues with this, and I still periodically find issues where I open a source file and everything is highlighted incorrectly (for example I found https://github.com/Dart-Code/Dart-Code/issues/2056 just last week!). The analyzer is able to provide reliably highlighting, but VS Code has no way to provide it programatically (I guess performance is a concern - having to round trip to a language server on each key press wouldn't be ideal).

Still, if VS Code gets something better that's still evaluated by Code (and it seems like there may have been more discussions around this recently), I'll definitely try to move over (ideally with something auto-generated from the Dart grammar file to ensure it's accurate).

Thanks for trying to improve this anyway! :-)

PR #2076, opened 2019-10-29
Add a tickbox to enable completion model

### LINKED ISSUE #2075: Add a user setting to enable analysis server's completion model
Should be a setting that adds `--enable-completion-model` to the analysis server flags.

Changing the setting should automatically restart the analysis server, and we should skip adding the flag for SDKs versions that don't support this flag (to avoid a crash if you switch back to an old SDK).

DanTup:
Screenshot 2019-10-29 at 8 04 19 am

Fixes #2075.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1352,6 +1352,12 @@
 					"description": "Whether to try to reuse existing DevTools windows instead of launching new ones. Only works for instances of DevTools launched by the DevTools server on the local machine.",
 					"scope": "window"
 				},
+				"dart.enableMachineLearningCodeCompletion": {
+					"type": "boolean",
+					"default": false,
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1352,6 +1352,12 @@
 					"description": "Whether to try to reuse existing DevTools windows instead of launching new ones. Only works for instances of DevTools launched by the DevTools server on the local machine.",
 					"scope": "window"
 				},
+				"dart.enableMachineLearningCodeCompletion": {
+					"type": "boolean",
+					"default": false,
+					"description": "Controls wether to use a machine-learned model for improved code completion rankings and suggestions.",
```
alexander-doroshko: wether -> whether =)

DanTup: Whoops, good catch - thanks!

Issue #1688, opened 2019-05-06
Allow dart syntax highlighting theming


xster:
In IntelliJ, you can theme local variables, instance variables, static variables, function params etc etc all differently. 

May be a dupe or may not be supported at all in vscode.

DanTup:
Our syntax highlighting is all regex-based using a TextMate grammar (https://github.com/Dart-Code/Dart-Code/blob/master/syntaxes/dart.json) so I don't think this is currently possible (it's also why we have strange quirks like #311) until VS Code gets an API for better colouring (https://github.com/Microsoft/vscode/issues/585).

DanTup:
Fixed by the semantic token work in https://github.com/dart-lang/sdk/commit/cb2ede57b7c91fb61c0cb05971cc1823faee2086.

In order to get the fix you'll need Dart-Code v3.18.0 (a preview release should be available later today/tomorrow), be using LSP Preview, and have a Dart SDK from after yesterday ([the most recent nightly](http://gsdview.appspot.com/dart-archive/channels/be/raw/latest/sdk/) works, and of course the next releases of Dart/Flutter SDKs will include this).

You can then use a theme that colours semantic tokens, or set them up in your own settings. For example, this sets static variables/properties to be rendered red:

<img width="1133" alt="Screenshot 2020-12-15 at 12 09 20" src="https://user-images.githubusercontent.com/1078012/102213287-6db9a480-3ece-11eb-899a-36c45a2b0651.png">

DanTup:
Due to some last minute issues, support for LSP 3.16 (which supports semantic tokens) won't be included in v3.18, but hopefully v3.19 next month instead.

Issue #311, opened 2017-05-25
Syntax highlighting colours function names differently when starting with capital


DanTup:
Not sure how easy this will be to fix, but TypeScript doesn't seem to have this problem. We should be able to identify functions from the parens?

![Colours](https://cloud.githubusercontent.com/assets/1078012/26466587/20aeab38-4187-11e7-9cb1-d858344a1030.png)

Skylled:
Isn't this because capitalization implies a Class name instead of a function?

DanTup:
That's the convention; but you don't have to follow it. If we were using the data from the analyzer, I'd expect they would be coloured correctly (based on actual AST, not guesses based on casing) so ideally we should be the same (whether it's reasonable or worthwhile, I don't know.. maybe we should just hope for https://github.com/Microsoft/vscode/issues/1967!)

Skylled:
Looking at this again on a whim, maybe we could assume that any ambiguous capitalized object that is followed by parens but not preceded by `new` or `const` is a function.

```
class IsClass {}

void NotClass(){}

var x = new IsClass();
var y = NotClass();
```

But that still won't help when assigning a function to a variable.
```
someObj.callback = NotClass // There's nothing here to detect from.
```

Additionally, the inverse problem is also true, lowerCamelCase class names are not treated correctly.
```
class alsoAClass {} // No colorizing
var z = new alsoAClass(); // Colored as a function
```

Seems like the best solution would be to get direct from the AST, as you said.

DanTup:
I think using the parens is probably the best way; seems more reliable than guessing based on case. I don't think Code is going to support dynamic colouring anytime soon :(

I'll see if I can change the regex at some point to handle this better.

weenzeel:
Found another case where upper cases is tripping up the syntax highlighting:

class FileMode {
  /// The mode for opening a file only for reading.
  static const read = const FileMode._internal(0);
  @Deprecated("Use read instead")
  static const READ = read;
  ...

In this snippet lower case read is scoped as source.dart but upper case READ is scoped as support.class.

DanTup:
@weenzeel thanks for the example. It's hard (for me) to fix some of these issues with the textmate grammar. I did hope to try and generate it from the language descriptions (in the hope it'd be accurate, including fixing the comment-highlighting bugs that keep coming up) - see https://github.com/dart-lang/sdk/issues/19298, but so far I haven't found a good way.

weenzeel:
[...]

PR #2203, opened 2020-01-07
[experimental] Semantic token highlighting

### LINKED ISSUE #2202: Use the analysis server for semantic highlighting
On the insiders channel, VS Code now provides apis for semantic token highlighting. This would allow Dart Code to enhance the static grammar-based highlighter with highlighting information from the analyzer.

In particular, the extension could use it to:
- be more accurate at semantic highlighting (members starting with an uppercase, named `get`/`set`, complex string literals, ...)
- highlight non-Dart files via analysis server plugins

To my knowledge, LSP doesn't support syntax highlighting (yet). So I think this would have to be implemented using the DAS protocol and a contributor for now. I've started some work on this and I hope I can open a PR next week.

## Current issues

- Not available in a stable VS Code release
- The analysis server doesn't highlight `null` and `throw` tokens, https://dart-review.googlesource.com/c/sdk/+/131071

simolus3:
## Changes

Recent insider builds of VS Code include an experimental api for semantic token highlighting. It allows extensions to report token types (like comment, string, keyword, class, function, variable, ...) and token modifiers (like static, abstract, member, ...). In this PR, I've added a contributor that's based on the Dart analysis server.

To make problems in the new highlighter easier to spot, I've temporarily removed the contributed grammar.

Server highlighting needs to be enabled via `"dart.analysisServerHighlighting": true`, `"editor.semanticHighlighting.enabled": true` also needs to be enabled.

### Problems and todos

- It seems like VS Code cancels a highlighting operation pretty fast if it takes to long. Sometimes files aren't analyzed at plugin start because the server takes to long. This is fixed after the first edit (and we still have the grammar-based highlighter), so I assume this shouldn't be much of an issue
- I haven't tested if this interferes with the grammar-based highlighter
- ~~I still need to go through more `HighlightRegionType`s from the analyzer and map them~~ done
- ~~Support nested highlight regions and multi-line tokens~~ done
- Of course, this is still blocked on VS Code and only available on an insiders build

## Screenshots

Note that all screenshots use server highlighting only. I also didn't go through all token types yet.
grafik

From an analyzer plugin:
grafik

Closes #2202, #1688

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,9 @@
 	"version": "3.8.0-dev",
 	"publisher": "Dart-Code",
 	"engines": {
-		"vscode": "^1.40.0"
+		"vscode": "^1.41.0"
 	},
+	"enableProposedApi": true,
 	"extensionKind": [
 		"workspace"
 	],
@@ -61,8 +62,9 @@
 				"configuration": "./syntaxes/dart-language-configuration.json"
 			}
 		],
-		"grammars": [
+		"$commented_out$grammars": [
 			{
+				"$reason": "Temporariliy commented out so that server highlights are more obvious. Undo before merge",
 				"language": "dart",
 				"scopeName": "source.dart",
 				"path": "./syntaxes/dart.json"
@@ -981,6 +983,12 @@
 					"type": "boolean",
 					"default": true,
 					"description": "Whether to use folding data from the Dart analysis server instead of the built-in VS Code indent-based folding.",
+					"scope": "window"
+				},
+				"dart.analysisServerHighlighting": {
+					"type": "boolean",
+					"default": false,
+					"description": "(Experimental) Whether to use syntax highlighting data form the Dart analysis server in addition to the built-in grammar.",
 					"scope": "window"
 				},
 				"dart.analysisExcludedFolders": {
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
 	"version": "3.8.0-dev",
 	"publisher": "Dart-Code",
 	"engines": {
-		"vscode": "^1.41.0"
+		"vscode": "^1.42.0"
 	},
 	"enableProposedApi": true,
 	"extensionKind": [
--- a/src/extension/analysis/analyzer.ts
+++ b/src/extension/analysis/analyzer.ts
@@ -46,6 +46,10 @@ function buildAnalyzerArgs(analyzerPath: string, dartCapabilities: DartCapabilit
 	if (config.analyzerInstrumentationLogFile)
 		analyzerArgs.push(`--instrumentation-log-file=${config.analyzerInstrumentationLogFile}`);
 
+	// Enable more detailed highlighting when enabled
+	if (config.analysisServerHighlighting)
+		analyzerArgs.push(`--useAnalysisHighlight2`);
+
 	// Allow arbitrary args to be passed to the analysis server.
 	if (config.analyzerAdditionalArgs)
 		analyzerArgs = analyzerArgs.concat(config.analyzerAdditionalArgs);
--- a/src/extension/analysis/open_file_tracker.ts
+++ b/src/extension/analysis/open_file_tracker.ts
@@ -15,7 +15,7 @@ export class DasFileTracker implements IAmDisposable {
 	private readonly flutterOutlines: { [key: string]: FlutterOutline } = {};
 	private readonly occurrences: { [key: string]: Occurrences[] } = {};
 	private readonly folding: { [key: string]: FoldingRegion[] } = {};
-	private readonly highlights: {[key: string]: HighlightRegion[]} = {};
+	private readonly highlights: {[key: string]: HighlightRegion[] } = {};
 	private readonly pubRunTestSupport: { [key: string]: boolean } = {};
 	private lastPriorityFiles: string[] = [];
 	private lastSubscribedFiles: string[] = [];
--- a/src/extension/analysis/open_file_tracker.ts
+++ b/src/extension/analysis/open_file_tracker.ts
@@ -1,8 +1,9 @@
 import { Disposable, TextDocument, Uri, window, workspace } from "vscode";
-import { FlutterOutline, FoldingRegion, Occurrences, Outline } from "../../shared/analysis_server_types";
+import { FlutterOutline, FoldingRegion, HighlightRegion, Occurrences, Outline } from "../../shared/analysis_server_types";
 import { IAmDisposable, Logger } from "../../shared/interfaces";
 import { fsPath } from "../../shared/utils/fs";
 import { WorkspaceContext } from "../../shared/workspace";
+import { config } from "../config";
 import { isUsingLsp } from "../extension";
 import { locateBestProjectRoot } from "../project";
 import * as util from "../utils";
@@ -14,6 +15,7 @@ export class DasFileTracker implements IAmDisposable {
 	private readonly flutterOutlines: { [key: string]: FlutterOutline } = {};
 	private readonly occurrences: { [key: string]: Occurrences[] } = {};
 	private readonly folding: { [key: string]: FoldingRegion[] } = {};
+	private readonly highlights: {[key: string]: HighlightRegion[]} = {};
 	private readonly pubRunTestSupport: { [key: string]: boolean } = {};
 	private lastPriorityFiles: string[] = [];
 	private lastSubscribedFiles: string[] = [];
@@ -34,6 +36,7 @@ export class DasFileTracker implements IAmDisposable {
 			delete this.occurrences[path];
 			delete this.folding[path];
 			delete this.pubRunTestSupport[path];
+			delete this.highlights[path];
 			this.updateSubscriptions();
 		}));
 		this.disposables.push(window.onDidChangeVisibleTextEditors((e) => this.updatePriorityFiles()));
@@ -41,6 +44,8 @@ export class DasFileTracker implements IAmDisposable {
 		this.disposables.push(this.analyzer.registerForFlutterOutline((o) => this.flutterOutlines[o.file] = o.outline));
 		this.disposables.push(this.analyzer.registerForAnalysisOccurrences((o) => this.occurrences[o.file] = o.occurrences));
 		this.disposables.push(this.analyzer.registerForAnalysisFolding((f) => this.folding[f.file] = f.regions));
+		this.disposables.push(this.analyzer.registerForAnalysisHighlights((f) => this.highlights[f.file] = f.regions));
+
 		// Handle already-open files.
 		this.updatePriorityFiles();
 		this.updateSubscriptions();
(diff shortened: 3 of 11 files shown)


DanTup:
Nice! I'll try it out and review it properly soon, but at a glance the code and screenshots look great! Have you noticed any issues (like bad colouring while typing due to latency) when using it?

We might need to come up with a better way to wait for the highlights - currently the code waits a while (which could maybe simplified something like [this](https://github.com/Dart-Code/Dart-Code/blob/3e867caebc1af757b8327aeb51639781712d57e3/src/extension/providers/dart_document_symbol_provider.ts#L13)) which works for when the file is first opened, but I think when typing it's likely to just get "the last highlights" so the offsets might end up incorrect (we might need to evict the existing cached highlights on-change or similar).

I wonder if there's a way we can conditionally disable the built-in syntax so we can ship behind the flag initially (it's not ideal testing if both are active when it's enabled). If not, we might want to ship a custom vsix build and call it out in a release for people to test out, then ship it always-on (removing the old one) from a future release.

Thanks!

simolus3:
> Have you noticed any issues (like bad colouring while typing due to latency) when using it

So far it appears like VS Code offsets the existing tokens if it doesn't get a response after an edit. I'll also try to use it for a longer duration after I finished the remaining mappings and see if everything feels natural.

> We might need to come up with a better way to wait for the highlights

Ah cool, I didn't know about `waitFor`. I had the idea of making the `DasFileTracker` expose promises (or `Event`s?) instead of a synchronous method. That way we could resolve pending promises as soon as analyzer notifications come in, we wouldn't have to wait for a fixed duration (of course we'd still need to respect cancellation tokens). What do you think?

There are some more complicated mappings that I still need to do. For instance, the analyzer sometimes reports multiline tokens (e.g. for multiline-comments and probably for `'''`-string literals as well).  We'd have to split those tokens at line endings, otherwise VS Code won't accept them.
Sometimes tokens are also nested: `'hello\nworld'` would report a string literal token from 0 to 16, and a "valid string escape" region from char 5 to 7. We'd have to transform that to a string literal, followed by the valid string escape, followed by another string literal. I'll see if I can come up with something clever here.

> I wonder if there's a way we can conditionally disable the built-in syntax so we can ship behind the flag initially

I don't have a lot of experience with VS Code plugin development so I'm not 100% sure. But it appears like grammars have to be set in the `package.json` and can't be registered dynamically.

DanTup:
Sorry for the delay in getting back to this, I've not been around as much recently.

>  That way we could resolve pending promises as soon as analyzer notifications come in, we wouldn't have to wait for a fixed duration (of course we'd still need to respect cancellation tokens). What do you think?

There's a `onDidChangeSemanticTokens` event that the provider can raise, so I think we could just subscribe to the analyzer event and fire that, which will cause VS Code to re-request tokens (this is how some things like the flutter outline tree notify for changes).

> We'd have to split those tokens at line endings, otherwise VS Code won't accept them.

If this isn't documented, it may be worth raising with VS Code to check it's not just a bug (and if it's not, that it's explicitly documented!).

> it appears like grammars have to be set in the package.json and can't be registered dynamically.

Yeah, though recently I was reminded that if we were to drop this, then opening a file while the analyzer is doing its initial scan could result in no tokens for a while (this won't affect your language since it'll be blank anyway, but it would seem like a regression for Dart). I think it probably makes sense to keep the original grammar as a fallback and for immediate colouring for Dart and then have the provider improve it (assuming it layers on top nicely).

I haven't reviewed the code that's here yet, so let me know if it's in a reasonable state for me to do so, or if you're likely to make any significant changes. We could also create a branch to land it into while it's still being worked on to avoid having to do everything in one PR if that makes sense.

simolus3:
Thanks, I didn't know about `onDidChangeSemanticTokens`. I'll integrate that and check the line-ending issue. 

> I haven't reviewed the code that's here yet, so let me know if it's in a reasonable state for me to do so

I don't think there will be any significant changes, but I want to clean some things up and write more tests for the token transformations if they turn out to be necessary. It shouldn't take long and I'll do that over the weekend, so it's probably best to not properly review this until next week.

simolus3:
I took another look - the [documentation](https://github.com/microsoft/vscode/blob/6fb14fddace8dbbfb9b47db916c3521e89312242/src/vs/vscode.proposed.d.ts#L293) does point out that tokens can't be multiline, so it looks like we need to split them.
I still have to do a transformation to remove nested highlight regions. For instance, if we had something like `///comment [link] more`, the analysis server would report the whole thing as a single comment (offset 0, length 22) with a documentation region in between (offset 11, length 6). For `[link]` to be a different token, we need to split that to a comment token from 0..10, followed by a documentation token from 11..16 and another comment from 17..21. I've asked [here](https://github.com/microsoft/vscode/issues/86415#issuecomment-596143316) to check if this transformation is really necessary on our side.

simolus3:
@DanTup I'm wondering if it makes sense to drop this and instead wait for LSP-based semantic highlighting? It looks like highlighting will be standardized with the next LSP release.

I'm not super familiar with the analysis server, but I could then try to make it support highlighting over LSP. I guess that's the best long-term approach either way. So if this will be obsolete in a couple of months and we don't have to do all the token transformations in here, it might be better to just wait and then use the LSP directly?

DanTup:
@simolus3 sorry for the delay, I'd been hoping to get back to this sooner.

The latest release of the extension does let you enable LSP mode, so if this API will be added to the LSP protocol soon, then I think that might be a good idea to avoid duplicated work. I do want to switch to LSP by default as soon as I can, but it's going to need plugin support. I have been looking at in the background - I've got the first request working when using the AngularDart plugin this week, though I have some other things that take priority in the short-term.

If you're interested in having a go there, I'll aim to tidy up my work so far and push it somewhere.

simolus3:
>  but it's going to need plugin support

Just to be clear: Each individual plugin will need to support LSP as well? I was hoping that the analysis server could act as a translating proxy here, similar to what it's doing with the Dart-specific protocol.
Are there any public resources describing how plugin communication will look like with LSP? I'm maintaining my own plugin, and obviously want it to continue working with LSP.

> If you're interested in having a go there

I would love to take at current LSP work - especially if plugins need changes as well. Hopefully most of this can be implemented in `package:analyzer_plugin` rather than the individual plugins. I'm happy to add support for more requests if all plugins can benefit from that.

Either way, I'll close this PR and wait for LSP to standardize semantic highlighting.

DanTup:
> Just to be clear: Each individual plugin will need to support LSP as well?

No sorry, I described it badly :) I'm testing with the AngularDart plugin, and the LSP server is translating LSP requests back into DAS-protocol requests and sending them to plugins (and then converting the plugin responses to LSP responses).

So far I've implemented only the definition request (as a proof-of-concept that plugins can work), but I'll also need to look at how notifications will work (things like errors). I'm hoping they're also quite straight forward (just translate them to LSP equivs), though I'm not sure how they're handled if multiple plugins (or plugin + base functionality) provide data for the same files (to avoid them overwriting each others errors) so I'll need to do a little investigation.

My initial change is here (though it needs much tidying up):

https://dart-review.googlesource.com/c/sdk/+/142378

It's possible in future plugins may want to deal with LSP (eg. to support LSP things that the server does not outside of LSP), but for now I think the translation will work fine.

simolus3:
Thanks for clearing that up, the DAS-protocol to LSP response translation definitely makes sense. 
I could assist by implementing other plugin functionality as dependent CLs if that helps.

DanTup:
Contributions would definitely be appreciated! It's not my top priority right now so I'm progressing it mainly when blocked on other things. I've updated the change above that addresses some of the comments and adds a test for the definition request using a mock plugin.

I think the main thing that would prevent this working with your plugin in VS Code if you enabled LSP preview (and ran the server from source, using that branch) would be that the initialization handshake doesn't include file extensions from plugins (if the plugins are initialised by the time this happens, it probably won't be too difficult to collect there `interestingFiles`).

If you do decide to have a go at implementing bits, let me know what you're working on so I can avoid overlapping - and if I can be of any assistance, do shout (I'm usually responsive [in discord](https://discord.gg/xSTPEqm) when I'm around).

subzero911:
It appeared that because of sematic token errors we cannot setup the color highlighting for variables (check the topic above)

DanTup:
@subzero911 this is unfortunately a known issue. There is a plan to move to semantic highlighting provided by the language server (which is what this PR was progress towards - https://github.com/Dart-Code/Dart-Code/issues/2202). There are many other issues with the TextMate grammar (for ex. https://github.com/Dart-Code/Dart-Code/issues/311 and https://github.com/Dart-Code/Dart-Code/issues/2056).

The Dart extension is going to move to [the LSP protocol](https://microsoft.github.io/language-server-protocol/), so implementing it in the current version is less worthwhile because the work would be lost and must be done again in LSP.

Semantic highlighting in LSP is still work-in-progress (https://github.com/microsoft/language-server-protocol/issues/18) but once it's finalised there, work can start on adding it to the LSP server. Then it would be available for VS Code themes when using the LSP mode (which can be opted in to using the `dart.previewLsp` setting until it's also ready to enable by default).

Hope this makes sense!

DanTup:
@simolus3 FYI (in case you thought I'd forgotten 😄) - since this API looks fairly stable now, I've made a start on implementing this in the LSP server (though nothing will land until LSP v3.16 is finalised):

https://dart-review.googlesource.com/c/sdk/+/158393/

It has some basic tests for Dart (none for plugins, though it does merge the results so they might "just work") though I haven't got as far as testing it for real in a client yet!

simolus3:
This is awesome, thanks for keeping me updated! Do you know how I can checkout CL branches to try this out? I imagine the push-based model of server plugins might cause some issues with the pull-based LSP model, especially for services with low latency like highlights. I'm happy to try this out, maybe it's not too bad.

Also, you mentioned splitting tokens at line endings as a todo. Do you think we might have to split nested tokens too? I think it might become an issue if we highlight nested string literals for instance. I'm happy to port the approach from this PR over if you think it's necessary.

DanTup:
> Do you know how I can checkout CL branches to try this out?

I don't know for Gerrit, but I pushed the branch to GH if that helps (you can use `dart.analyzerPath` to run the server from source, but you'll need to make sure you're on a recent build (eg. [nightly](http://gsdview.appspot.com/dart-archive/channels/be/raw/latest/sdk/))):

https://github.com/DanTup/sdk/tree/lsp-3-16-semantic-tokens

(also be ware server startup will be slower from source)

> Also, you mentioned splitting tokens at line endings as a todo. Do you think we might have to split nested tokens too?

I'm not sure - I couldn't find anything about that in the LSP docs and because the character deltas appear to be from the start of the previous token (not the end), I thought maybe they're intended to be supported (though I haven't got as far as any that overlap yet).

(FWIW I'm out of the office next week, so this might not progress much more - and I may be less responsive than normal - until I'm back).

simolus3:
No problem! I'll try to run your version with my plugins sometime next week then :)

And yeah, I meant overlaps in general when I mentioned nested tokens. Judging by https://github.com/microsoft/vscode/issues/86415#issuecomment-596190099 we might need that transformation in the server, but since it mostly highlights atomic identifiers and keywords there's little chance of that actually happening.

simolus3:
I couldn't get this to work, not even for Dart. I'm using your SDK fork, the current version of the extension from git and VS Code insiders. In the server diagnostics, I see that server capabilities contain a `semanticTokensProvider` but the client capabilities don't mention semantic tokens. That probably indicates a misconfiguration on my end, do you see anything obvious that I'm missing?

```json
{
    "dart.additionalAnalyzerFileExtensions": ["moor"],
    "dart.analyzerDiagnosticsPort": 9876,
    "dart.analyzerInstrumentationLogFile": "/tmp/analyzer_instrumentation.txt",
    "dart.sdkPath": "/path/to/current/dev/sdk",
    "dart.analyzerPath": "/your/dart/fork/pkg/analysis_server/bin/server.dart",
    "dart.previewLsp": true,
    "editor.semanticHighlighting.enabled": true,
    "dart.analyzeAngularTemplates": true,
}
```

DanTup:
@simolus3 I hadn't tried in VS Code yet, only tests in the server.

I just had a look, and the node lsp client with semantic token support hasn't been released yet. I can't figure out how to run it from source, however [this tag](https://github.com/microsoft/vscode-languageserver-node/releases/tag/release%2Fclient%2F7.0.0-next.9) was made a few hours ago, so I think once that version is published to npm, I can make a build of VS Code that uses that(preview) version to do some testing there.

simolus3:
No worries! I thought semantic highlighting already worked in VS Code, if that's not the case I can also wait a little longer :)

DanTup:
It does work in VS Code, however for LSP there's [a client library](https://www.npmjs.com/package/vscode-languageclient) that needs to also be updated (that's a shared library maintained by MS that translates between the VS Code APIs and LSP). Until the LSP spec is finalised, that won't ship in the release version, but we can use the pre-release builds once they're made (I did look, but couldn't figure out how to reference it from source).

That might happen today based on that tag, so if it's before I finish (or otherwise, probably tomorrow), I'll try to make it all work and build a version of Dart-Code we can use for testing.

DanTup:
I've opened https://github.com/microsoft/language-server-protocol/issues/1077 asking for some explicit info about nested/intersecting tokens, as the comment you linked says the characterDelta is from the end of the previous token, but the LSP spec says it's from the start.

PR #2232, opened 2020-01-24
Add CodeLens links for Flutter DartPad samples

### LINKED ISSUE #2151: Replace "Open Widget Sample" with in-line Code-Lens links
Equiv of https://github.com/flutter/flutter-intellij/pull/4139

DanTup:
This adds CodeLens links for Flutter files that have `{@tool dartpad` snippets that open the interactive online samples:

Screenshot 2020-01-24 at 15 13 12

Screenshot 2020-01-24 at 15 13 24

It has some assumptions/caveats:

- Only works for files inside `(Flutter SDK)/packages/flutter/lib/src/material/app_bar.dart`
- Assumes the leaf-most folder name is the package (eg. in the above, "material") and that the `name` from the `Outline` node is the classname, and uses these in the URL as `${sample.libraryName}/${sample.className}-class.html`
- Always links to `https://api.flutter.dev` regardless of your Flutter branch/channel
- Assumes there's an element with the ID `#${sample.libraryName}.${sample.className}.1` at the correct location (this jumps down the page - if this assumption becomes invalid, you'll end up at the top of the docs page for that class)

@devoncarew I think these are the same for the IntelliJ version too (except the last one - though you might wish to add) but shout if anything seems wrong here.

Fixes #2151.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1121,6 +1121,12 @@
 					"description": "Whether to show Code Lens actions in the editor for quick running/debugging tests.",
 					"scope": "window"
 				},
+				"dart.showDartPadSampleCodeLens": {
+					"type": "boolean",
+					"default": true,
+					"description": "Whether to show Code Lens actions in the editor for opening online DartPad samples.",
+					"scope": "window"
+				},
 				"dart.showTodos": {
 					"type": "boolean",
 					"default": true,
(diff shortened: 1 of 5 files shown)


devoncarew:
Great!

> I think these are the same for the IntelliJ version too (except the last one - though you might wish to add) but shout if anything seems wrong here.

Yup, this is what IntelliJ is doing (modulo the last item). I elected to not try and link to a specific anchor - it was unclear to me how stable that would be going forward. Linking to package+class seems stable (w/ the caveat that for most cases the user will need to scroll a bit to see the actual dartpad instance on the page).

DanTup:
> I elected to not try and link to a specific anchor - it was unclear to me how stable that would be going forward

Yeah, I did consider leaving it out, but it seems fairly stable (they're used in the IDs for `flutter create --sample` and always ascend from 1) and if they change it just reverts to not scrolling anyway (there are no errors, it just doesn't scroll down anywhere). I also thought it felt a bit weird given the text on the code lens if we didn't scroll (since then it's just like "open API docs" and it might raise questions about why we only do it for some specific set of classes).

PR #2243, opened 2020-01-31
Change Flutter Q1 survey date

DanTup:
@pq @jayoung-lee FYI!

Note: I changed the close time from 5pm to 4pm UTC because otherwise it came out at 10am (because you go into PST?) and wasn't sure if which timezone the survey close date is in (I'd rather stop showing the prompt an hour before it ends than continue an hour after).

### DIFF
--- a/src/shared/constants.ts
+++ b/src/shared/constants.ts
@@ -60,10 +60,10 @@ export const takeSurveyAction = "Take Survey";
 // To confirm dates in PST(/PDT), paste this into Chrome dev console:
 // new Date(Date.UTC(...)).toLocaleString("en-US", {timeZone: "America/Los_Angeles"})
 // Paste the results into comments for easier reviewing!
-// "2/24/2020, 9:00:00 AM"
-export const surveyStart = Date.UTC(2020, 1 /* Month is 0-based!! */, 24, 17, 0);
-// "3/2/2020, 9:00:00 AM"
-export const surveyEnd = Date.UTC(2020, 2 /* Month is 0-based!! */, 2, 17, 0);
+// "3/3/2020, 9:00:00 AM"
+export const surveyStart = Date.UTC(2020, 2 /* Month is 0-based!! */, 3, 17, 0);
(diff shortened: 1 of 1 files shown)


pq:
@jayoung-lee: I *think* the dates in my PR are what you wanted?  

https://github.com/flutter/flutter-intellij/pull/4348

If not let me know ASAP.  It may or may not be too late to cherry-pick...

jayoung-lee:
I think Danny is using UTC and 2 here means March...So that's why it looks
different? It's March 3rd to  March 10th (9am PDT& PST)! One hour
difference should not matter , but thanks Danny for taking care of it!

JaYoung

DanTup:
Yeah, in JavaScript dates have 0-based months, because 🤷‍♂️

I always update the comments to include a US-formatted version of the date in LA time (generated by executing the `Date.UTC` call below it):

```ts
// "3/3/2020, 9:00:00 AM" <-- This is the start time in PST (UTC-8)
export const surveyStart = Date.UTC(2020, 2 /* Month is 0-based!! */, 3, 17, 0);

// "3/10/2020, 9:00:00 AM" <-- This is the end time in *PDT* (UTC-7)
export const surveyEnd = Date.UTC(2020, 2 /* Month is 0-based!! */, 10, 16, 0);
```

I just noticed that the end date came out at 10am when I updated the dates, so I chopped an hour off. It's possible yours will be 10am local time (PDT) because your code explicitly sets PST, but I'm not sure and it sounds like that won't be a problem (I wasn't just if the survey was closing strictly at this time, but it sounds not).

DanTup:
I often get confused by your timezones, because in summer we're "BST" (British Summer Time) so I read "PST" as "Pacific Summer Time" when it's the other way around 🤨

jayoung-lee:
Yeah, that is totally fine :)  I just talked to Phil, too. Thanks! Woohoo!

jayoung-lee:
I actually had to look it up again, because we call it "summer time" in
Korea too .... 😀

Issue #524, opened 2018-02-04
Save-All sends multiple reload requests


DanTup:

DanTup:
Added a 200ms debounce but also raised https://github.com/Microsoft/vscode/issues/42913 in the hope of improving this (if we make the delay too big we're slowing down the refresh cycle but if it's too small we still send multiple requests which may either result in multiple restarts or the subsequent ones being dropped because reload was already in progress).

PR #2298, opened 2020-03-13
Add a setting to ignore analyzer error codes.

rickybrent:
Intended as a work around for https://github.com/dart-archive/angular_analyzer_plugin/issues/524 , but useful for suppressing any unwanted analyzer warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -998,6 +998,15 @@
 					},
 					"scope": "resource"
 				},
+				"dart.analysisIgnoredErrors": {
+					"type": "array",
+					"default": [],
+					"description": "An array of error codes to ignore from Dart analysis. This option should usually be set at the Workspace level.",
+					"items": {
+						"type": "string"
+					},
+					"scope": "window"
+				},
 				"dart.debugSdkLibraries": {
 					"type": "boolean",
 					"default": false,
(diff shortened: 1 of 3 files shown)


DanTup:
Is this not something that can be handled with the analysis_options file?

https://dart.dev/guides/language/analysis-options#ignoring-rules

DanTup:
@rickybrent ping!

rickybrent:
Sorry for the late reply -- I wanted to test some of the other options mentioned in the bug (#524) and completely lost track of time!

I don't think it is, though it could be user error -- I'm actually mostly on the backend side of our dart efforts and the main developer uses vim. Setting implicit-dynamic to true doesn't seem to have any effect, and unfortunately "two_way_binding_output_type_error" isn't a recognized as an error code by the analyzer -- adding it to analysis_options just creates a new "unrecognized_error_code" warning.

All that said ... I'm not sure this is the right place to make such a fix or workaround. Ideally I'd have opened a pull request fixing #524 on the analyzer plugin instead, but adding an option to your excellent dart plugin was much easier. :sweat_smile:  

I completely understand if you'd rather not add this as an option. :)

DanTup:
Ah, you're right - it seems like the server only understands its own errors, and not those from plugins.

I agree it would be better handled in the server, so I've opened https://github.com/dart-lang/sdk/issues/41531. If it doesn't seem possible to address there, we could do it here - but that should probably be a last resort, as fixing it here will apply only to VS Code (and not when you run the command line tools - for ex. on CI servers, or if there are IntelliJ/etc. users also working on it).

DanTup:
Going to close this as I don't think it makes sense to fix here, but more importantly - it would get lost when moving to LSP anyway (which is currently behind a preview flag, but will become the main implementation soon).

This will need to be fixed in the plugin (the issue linked above is in an archived repro, so it may need re-raising in a non-archived repo!) or the servers ignore issue fixed (there's an SDK issue linked above).

Thanks for the the PR anyway - sorry the resolution isn't so simple!

PR #2325, opened 2020-03-28
Update analyzer.ts

devoncarew:
- change the client ID we send in for analysis server analytics

This ID (not well documented) is meant to represent the general platform that the analysis server is running from. This might be something like IntelliJ, Android Studio, VS Code, gitpod, ...

The client version similarly should be related to the version of the platform (although this isn't as critical from an analytics point of view). It might be nice to have the VS Code version in here.

cc @DanTup

### DIFF
--- a/src/extension/analysis/analyzer.ts
+++ b/src/extension/analysis/analyzer.ts
@@ -39,7 +39,7 @@ function buildAnalyzerArgs(analyzerPath: string, dartCapabilities: DartCapabilit
 		analyzerArgs.push(`--port=${config.analyzerDiagnosticsPort}`);
 
 	// Add info about the extension that will be collected for crash reports etc.
-	analyzerArgs.push(`--client-id=Dart-Code.dart-code`);
+	analyzerArgs.push(`--client-id=VS-Code`);
 	analyzerArgs.push(`--client-version=${extensionVersion}`);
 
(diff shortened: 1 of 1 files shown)


DanTup:
Any preference on the version - should I change it to VS Code version? I think the plugin version is probably more useful, but it does seem strange to have them mismatched between VS Code/extension (maybe there's value in having a `--plugin-version` or something?).

devoncarew:
> Any preference on the version - should I change it to VS Code version? I think the plugin version is probably more useful, but it does seem strange to have them mismatched between VS Code/extension (maybe there's value in having a `--plugin-version` or something?).

Yeah, I don't know. We _do_ want the platform coming through here - that's valuable to us.  The Dart-Code version is probably more generally useful than the VS Code platform version, but like you said that then leaves a mismatch between the tw0 bits of metadata.

devoncarew:
We could add more granularity here but I'm not sure it's worth the additional cli flags. `--client-id` and `--client-version`, `--plugin-id` and `--plugin-version`, ...

DanTup:
I don't see where this goes, so don't have any real opinion. I'll merge this as-is (the failures are unrelated - some are issues in VS Code Insiders, and some are breaks in Flutter's web debugging that's being looked at) but if you want me to switch it to the VS Code version instead, just let me know!

PR #2338, opened 2020-04-08
Switch to reading Flutter survey data from the web

DanTup:
Equiv of https://github.com/flutter/flutter-intellij/pull/4449.

@jayoung-lee @pq FYI! I have tested it with my own copy of the JSON file with varying dates (plus updated the automated tests) but please let me know when you have dates for the next survey and update the file, so I can double-check it. Thanks!

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -6,7 +6,7 @@ import { Analyzer } from "../shared/analyzer";
 import { DaemonCapabilities, FlutterCapabilities } from "../shared/capabilities/flutter";
 import { dartPlatformName, flutterExtensionIdentifier, flutterPath, HAS_LAST_DEBUG_CONFIG, isWin, IS_LSP_CONTEXT, IS_RUNNING_LOCALLY_CONTEXT, platformDisplayName } from "../shared/constants";
 import { LogCategory } from "../shared/enums";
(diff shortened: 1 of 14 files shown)


DanTup:
In VS Code we're not previleged enough to be able to run code in the UI thread (we can't even run code in the same *process* as the UI/editor 😁) so that shouldn't be an issue. I can't see anything in that PR, but did you have to change anything for DNS/proxy issues?

If the file can't be fetched in VS Code, it'll just write the connection error/timeout to the dev tools console (and ofc, not show a survey prompt at all).

jayoung-lee:
I have not set the dates for the Q2 survey yet, but I will do so soon (and update the json file) and ping you all. Thanks!

pq:
> did you have to change anything for DNS/proxy issues?

Well, I *tried*.  I think that moving to the platform `HttpRequests` class rather than my hand-rolled URL processing should handle it but need to push out a dev build for testing to be sure.

jayoung-lee:
Alright, the json file is updated with Q2 survey information!

DanTup:
Great, thanks! I tested with dates before and during the dates in there and the prompt appeared when I'd expect. The change will ship in this months release.

jayoung-lee:
Awesome. Thank you, Danny!

PR #2506, opened 2020-05-29
Enable ignore lint quick fix

### LINKED ISSUE #2487: Enable assist to ignore lints by default
When I start working on a new code base one of the first things I tend to do is add the effective dart lints.
This typically yields thousands of errors.
A moderate no. of the lints are not valid.

It would be really convenient if the dart-code offered an 'add ignore' option quick fix for these lints.

BootBlock:
Fixes #2487.


DanTup:
Thanks!

PR #2507, opened 2020-05-29
Rename preview flag for file renames/moves

### LINKED ISSUE #2482: Remove preview flag for file renames/moves
Perhaps rename the flag, and default to try - but leave a flag, so people can turn this on and off (otherwise, add a prompt like TypeScript).

BootBlock:
Fixes #2482.


DanTup:
Thanks!

PR #2540, opened 2020-06-09
Switch dart.progress for DAP progress events

### LINKED ISSUE #2516: Migrate dart.progress events to use DAP's progressEvent
https://github.com/microsoft/debug-adapter-protocol/issues/92

BootBlock:
Fixes #2516.

### DIFF
--- a/src/debug/dart_debug_impl.ts
+++ b/src/debug/dart_debug_impl.ts
@@ -138,8 +138,6 @@ export class DartDebugSession extends DebugSession {
 			return;
 		}
 
-		this.sendEvent(new ProgressStartEvent(debugLaunchProgressId, "Launching"));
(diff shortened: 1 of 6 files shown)


DanTup:
Thanks! I caused some conflicts with this, so I've rebased your branch over master and fixed up.

I also changed this to pass an empty title:

```
this.sendEvent(new ProgressStartEvent(debugLaunchProgressId, "", "Launching..."));

```

It's not clear if this is allowed (the spec says title is mandatory), but I've filed https://github.com/microsoft/language-server-protocol/issues/1025 for clarification. If it's not allowed, I think there should be a way to allow us to have single-message updatable progress without needing to have two messages joined with colons.

PR #2578, opened 2020-06-23
Add preview flag for running DevTools embedded

DanTup:


String-error-404:
how to get the dart dev tools  in vs code

DanTup:
@String-error-404 when you're in a debug session, you should be able to click the Inspector icon on the debug toolbar, or use the DevTools commands from the VS Code command palette.

PR #2598, opened 2020-06-30
Add support for the Flutter snap on Linux

MarcusTomlinson:
At Canonical we've developed a convenient way to install and use Flutter on Linux.

By simply running:

```
snap install flutter --classic
```

You'll have all the dependancies you need to begin developing Flutter applications. From here you just type: `flutter create`, `flutter run`, etc. as normal to perform flutter tasks.

The Flutter repo is stored under `$HOME/snap/flutter/common/flutter/`, but the actual flutter script (executed by the call to `flutter`) is installed to `/snap/flutter/current/flutter.sh`. This allows users to switch flutter versions as needed via the home directory, while taking advantage of the dev environment provided by the snap.

To get this working in VS Code, we need the extension to detect whether the flutter snap is installed, and if so, be aware of the separate repo and script paths for execution.

Furthermore, because snaps are installed at root-level, the Flutter repo is only populated on first user run of the snap. We therefore must also check if the repo exists on startup, and initialise it if not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 Dart Code extends [VS Code](https://code.visualstudio.com/) with support for the
 [Dart](https://www.dartlang.org/) programming language, and provides tools for
-effectively editing, refactoring, running, and reloading [Flutter](https://flutter.io/)
+effectively editing, refactoring, running, and reloading [Flutter](https://flutter.dev/)
 mobile apps, and [AngularDart](https://angulardart.org) web apps.
 
 


src/extension/sdk/utils.ts
```diff
@@ -237,22 +239,38 @@ export class SdkUtils {
...
+			if (!fs.existsSync(flutterSnapSdkPath + "/.git")) {
+				await window.withProgress(
+					{
+						location: ProgressLocation.Notification,
+						title: "Initializing Flutter snap",
+					},
+					async () => {
+						await util.promisify(child_process.exec)(flutterSnapBinPath);
+					});
+			}
```
DanTup, line 255: I'm thinking this only makes sense to do if we don't find a Flutter SDK using the normal path. If a user has set up Flutter without using Snap (and added to `PATH` or set `dart.flutterSdkPath`(s)) it would be strange to kick this off and then ignore those settings.

src/shared/utils.ts
```diff
@@ -123,6 +123,9 @@ export function usingCustomScript(binPath: string, binArgs: string[], options?:
 			? options.customScriptReplacesNumArgs
 			: 1; // Default to removing one arg.
 		binArgs = binArgs.slice(numArgsToRemove);
+	} else if (fs.existsSync(flutterSnapBinPath) && binPath.endsWith("/flutter")) {
+		// If installed, use flutter from the snap instead of the SDK path.
+		binPath = flutterSnapBinPath;
```
DanTup, line 128: It feels like this change should be unnecessary - this code is specifically to handle user/repo-specified custom locations, and it should fall back to the detected SDK. If you remove this, what doesn't work correctly?

src/extension/sdk/utils.ts
```diff
@@ -237,22 +239,38 @@ export class SdkUtils {
...
-			firstFlutterMobileProject && path.join(firstFlutterMobileProject, ".flutter"),
-			firstFlutterMobileProject && path.join(firstFlutterMobileProject, "vendor/flutter"),
-			process.env.FLUTTER_ROOT,
-		].concat(paths).filter(notUndefined);
-
-		const flutterSdkPath = this.findFlutterSdk(flutterSdkSearchPaths);
-		// Since we just blocked on a lot of sync FS, yield.
-		await resolvedPromise;
+		let flutterSdkPath;
+		if (fs.existsSync(flutterSnapBinPath)) {
```
DanTup, line 243: Maybe worth adding an `isLinux` check at the start of this.

src/extension/sdk/utils.ts
```diff
@@ -237,22 +239,38 @@ export class SdkUtils {
...
+			flutterSdkPath = flutterSnapSdkPath;
+			// If the Flutter snap is installed but not initialised, initialise it now.
+			if (!fs.existsSync(flutterSnapSdkPath + "/.git")) {
+				await window.withProgress(
```
DanTup, line 247: I'm thinking we might want to ask the user first before starting this. Although they're unlikely to say no, it really bugs me when a VS Code extension just starts downloading (potentially large) files without any prompt - especially when VS Code can eagerly activate extensions for many unexpected reasons (for example if you open a large non-Dart project and VS Code fails to walk the tree within 7 seconds, it will [automatically activate the Dart extension](https://github.com/microsoft/vscode/issues/73656)!)

DanTup:
I think I may have misunderstood this slightly..

> If the Flutter snap is installed but not initialised

I had assumed this code ran on a clean install of Ubuntu (eg. `flutterSnapBinPath` would exist by default), however re-reading that comment now I'm not so sure. Would this only be created if the user has already taken an action to install this (outside of VS Code), and this just performs some required initialisation step?

If so, is there a reason snap doesn't automatically initialise when installing? This change feels a little more strange to have in the extension if it still requires the user to have gone and performed some other install step first (but not a second initialisation step).

MarcusTomlinson:
So no, this is not installed on Ubuntu out of the box. This is an application (snap) that users opt into installing via `snap install flutter`.

Once the user installs the snap, `flutterSnapBinPath` will exist but `flutterSnapSdkPath` will not. This is because snaps are installed at root-level (i.e. with no awareness of which user will actually use it). Therefore, it has to be on first execution of the snap (i.e. calling `flutter`) that the SDK path is populated in that user's home directory.

This is actually quite transparent to a user of the terminal. The first time you call `flutter run` for example, the snap script will first "initialise" (download) the SDK directory, then proceed immediately to executing the run command. From VS Code however, we must force this initialisation way upfront as to ensure everything is in place before SDK detections begin (especially the detection of the Dart SDK which is within the Flutter one).

src/shared/utils.ts
MarcusTomlinson: If we remove this, the binPath will be set to `$HOME/snap/flutter/common/flutter/bin/flutter` and not `/snap/flutter/current/flutter.sh`.

Calling the former will result in build errors if any dev dependancies are not installed on the users system (for example, CMake or Ninja). This is the convenience provided by the snap: The latter is a script that ensures environment variables such as PATH, LDFLAGS, etc. are configured correctly to point into the snap package, such that the user need not manually do anything more than `snap install flutter` to get going with their Flutter development.

src/extension/sdk/utils.ts
MarcusTomlinson: Good point. I'll fix this when I'm back at my desk tomorrow :)

src/shared/utils.ts
DanTup: > Calling the former will result in build errors if any dev dependancies are not installed on the user's system

Won't this have already been taken care of by the initialisation during the SDK searching?

MarcusTomlinson: Build dependancies (CMake, Ninja, etc.) are installed together with the snap under `/snap/flutter/current/...`, however in order for flutter to know where those are, some environment variables need to be tweaked (PATH, LDFLAGS, etc.). `/snap/flutter/current/flutter.sh` does this. Calling `flutter` directly from the SDK directory does not.

DanTup: Oh, I thought this was just a script to do the initialisation. So the first time this script runs, it will initialise, and then subsequent runs it will forward on to the real Flutter binary? What's the output of the process on the first run?

I think it may be better to try and wrap it up in something like a `WorkspaceConfig` (similar to the Bazel one). The Bazel one overrides specific Flutter sub-commands scripts (`flutterRunScript` etc.) but there's no reason we couldn't also have a base `flutterScript` field that allows overriding the flutter binary rather than the sub-commands.

The entry point for the Bazel one is in `src/extension/sdk/utils.ts` and looks like this:

```ts
processWorkspaceType(findGitRoot, processKnownGitRepositories);
processWorkspaceType(findBazelWorkspaceRoot, (l, c, b) => processBazelWorkspace(l, c, b, config.previewBazelWorkspaceCustomScripts));
```

The first arguments (`findGitRoot`/`findBazelWorkspaceRoot`) detect whether a workspace is of a certain type (in this case it's not really workspace specific, but we can detect if the snap is installed) then the second argument (`progressKnownGitRepositories` etc.) will update the `WorkspaceConfig` (eg. in this case, set a new `flutterScript` field that would be used in the code above (this would make it generic instead of Snap-specific)).

I think that way, all snap-specific code is inside those two functions (find + process) rather than dotted across multiple places. If other OSes had similar things, this would also make it easier to implement them without extending all of these cases.

WDYT?

It may also be good to get an integration test around this (eg. on GitHub actions we use an Ubuntu box, install the snap, mock the prompt to automatically approve, and ensure we can load successfully), but we can do that once we've worked the implementation details out.

Thanks!

MarcusTomlinson: Ha, it’s funny you mention doing this as a workspace config. That’s precisely how I originally implemented this and thought perhaps it was the wrong place :) I agree, it’s much cleaner that way, far less diff. Alright, I'll go in that direction again.

DanTup: Ah, doh :) I think it makes more sense wrapped up there - it was specifically to try and avoid leaking logic like this around the place, and instead allow a simpler way to override behaviour based on the environment.

The only possibly niggle, is that it might prevent us from suppressing the initialisation in the case where the user installed (but didn't initialize) the snap package, but then intends to use a VS-Code-configured path. It's probably a bit of an edge case though, so if the rest of the code is cleaner and isolated as a result, it's probably not a big problem.

Shout if anything isn't clear!

MarcusTomlinson: Oh right. Yeah... it simply will not work without the explicit initialization logic as I have it now because the SDK path will not exist for the extension to detect. Chicken and egg situation.

DanTup: I was still thinking this would exist, but inside the `processSnapWorkspace`(?) function (so `processWorkspaceType` would likely need to be come async). So the equiv of the `find` function would return truthy if the Snap exists (whether it's initialised or not), then the equiv of the `process` function would initialise if required (being async so holding up the rest of the code while this happens), then write the paths into the config and continue.

DanTup:
@MarcusTomlinson are you still planning to work on this? If not, I'm happy to have a go at the changes discussed above. (Or if you're stuck on any part of it, let me know!)

MarcusTomlinson:
> @MarcusTomlinson are you still planning to work on this? If not, I'm happy to have a go at the changes discussed above. (Or if you're stuck on any part of it, let me know!)

Hey Danny, sorry for the delay, I'm busy moving this to another branch. I'll PR something this morning. I suspect I'll need some help with the testing yeah.

MarcusTomlinson:
@DanTup it may not be a totally rubbish idea for you to give this a go, if you'd like to. I'm happy to prod on in parallel, even if just as a learning exercise tbh.

PR #2608, opened 2020-07-06
Update from flutter.io to flutter.dev

timsneath:
These were the only two remaining instances I could see in the codebase. Fixes dialogs like this:
image


DanTup:
Thanks!

PR #2613, opened 2020-07-07
Add support for Flutter snap on Linux

MarcusTomlinson:
At Canonical we've developed a convenient way to install and use Flutter on Linux.

By simply running:

```
snap install flutter --classic
```

You'll have all the dependancies you need to begin developing Flutter applications. From here you just type: `flutter create`, `flutter run`, etc. as normal to perform flutter tasks.

The Flutter repo is stored under `$HOME/snap/flutter/common/flutter/`, but the actual flutter script (executed by the call to `flutter`) is installed to `/snap/flutter/current/flutter.sh`. This allows users to switch flutter versions as needed via the home directory, while taking advantage of the dev environment provided by the snap.

To get this working in VS Code, we need the extension to detect whether the flutter snap is installed, and if so, be aware of the separate repo and script paths for execution.

Furthermore, because snaps are installed at root-level, the Flutter repo is only populated on first user run of the snap. We therefore must also check if the repo exists on startup, and initialise it if not.


DanTup:
@MarcusTomlinson looks great, thanks! I'm going to merge into a branch for now so I can try it out when I have a VM set up and also to try and include a test for it in GitHub actions. I may ping you if I get stuck of have questions. Thanks!

PR #2660, opened 2020-07-22
Fix "tasks unexpectedly provided a task of type" error

szotp:
Enable previewBuildRunnerTasks in flutter project and then execute
Command Palette > Tasks: Run task > Show all tasks...

You will see this error:
The task provider for "pub" tasks unexpectedly provided a task of type "flutter".

Probably because pub is hardcoded here:
https://github.com/Dart-Code/Dart-Code/blob/b6aac604c02af46d66f99e01650c2b4b50c9edfc/src/extension/extension.ts#L337

So I changed the code to always use "pub".

### DIFF
--- a/package.json
+++ b/package.json
@@ -1950,15 +1950,6 @@
 						"type": "string"
 					}
 				}
-			},
-			{
-				"type": "flutter",
-				"required": [],
-				"properties": {
-					"command": {
-						"type": "string"
-					}
-				}
 			}
 		],
 		"problemMatchers": [
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -331,7 +331,8 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean 
 
 	// Task handlers.
 	if (config.previewBuildRunnerTasks) {
-		context.subscriptions.push(vs.tasks.registerTaskProvider("pub", new PubBuildRunnerTaskProvider(sdks)));
+		const provider = new PubBuildRunnerTaskProvider(sdks);
+		context.subscriptions.push(vs.tasks.registerTaskProvider(provider.type, provider));
 	}
 
 	// Snippets are language-specific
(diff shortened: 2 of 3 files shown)


DanTup:
I'm not seeing this error - when I pick "Show all Tasks" I just get this empty box:

Screenshot 2020-07-23 at 14 13 33

Where does the error appear for you?

DanTup:
It's possible the better fix would be conditionally using pub/flutter here:

https://github.com/Dart-Code/Dart-Code/blob/41a7f4c39edc991a7ff4641008713df020683667/src/extension/extension.ts#L337

Though I'm curious why I'm not seeing the same error you do.

DanTup:
Ok, the issue was that I didn't have `bulid_runner` referenced.

I think a better fix would be to use the correct ID based on the type of project (for example if you have a workspace with multiple projects, we should show the correct one for each).

Something more like this:

[patch.txt](https://github.com/Dart-Code/Dart-Code/files/4966113/patch.txt)

If I open a workspace with a Flutter + Dart project, this will then show the correct type:

Screenshot 2020-07-23 at 14 23 44

WDYT?

szotp:
eh, I wish that flutter did not create this pub/flutter division, it seems completely pointless.

But your proposal looks good, I will update my MR later.

szotp:
Ok, I have updated.

But, while doing this, I had another idea: maybe task provider should just have a different name that works both for pub and flutter, and then decide what to call when task is being ran.

Instead of pub/flutter: build_runner watch use:

Option A: build_runner: watch
Option B: dart: build_runner watch

If this won't work for some reason, please merge as it is.

DanTup:
I think it probably would - I'm not sure why I previously separated them in the UI. I don't think anyone running build_runner would be specifically looking for "flutter" so I think it could all just be using a `type` of `pub`. If you want to strip that out and simplify it (so the only difference is probably inside `createBuildRunnerCommandBackgroundTask` where it picks the `program`/`args`), that sounds good to me (you could also then remove the `flutter` `taskDefinition` from `package.json`).

Thanks!

szotp:
Let's just use pub then. PR updated.

DanTup:
LGTM, thanks!

PR #2894, opened 2020-10-21
Add TODO comment snippet

lsaudon:
[Flutter Style Todos](https://dart-lang.github.io/linter/lints/flutter_style_todos.html)

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -126,6 +126,13 @@
 				"  $0",
 				"});"
 			]
+		},
+		"todo": {
+			"prefix": "todo",
+			"description": "Insert a TODO comment.",
+			"body": [
+				"// TODO($1): $0"
+			]
 		}
 	}
 }


DanTup:
Thanks for the PR! Unfortunately snippets in this file are not scoped and can show up when you don't want them to. My goal is to reduce the number here rather than increase them. For example, this snippet will show up when writing strings:

Screenshot 2020-10-21 at 16 10 26

This is not really desirable (particularly for users that won't use the snippet). A better option if you find this useful would be to [add it to your own snippets file](https://code.visualstudio.com/docs/editor/userdefinedsnippets#_create-your-own-snippets).

In the future it would be nice to move snippets to the analysis server (where it can take the context into account), but it's not something that's supported today.

Thank you for the effort anyway!

lsaudon:
It's the same problem for all snippets.

DanTup:
Yep, that's correct - that's why I'd like to reduce them/move them to the server (which has some context and could conditionally include them). There's an open issue for supporting them in the server at https://github.com/dart-lang/sdk/issues/37831.

PR #3014, opened 2020-12-16
Exclude .dart_tool folder from upgrade packages list

### LINKED ISSUE #3013: .dart_tool folder shows up in package list
When running **Flutter: Upgrade Packages** in the gallery:

Screenshot 2020-12-15 at 14 30 47

BootBlock:
Prevents the `.dart_tool` folder from appearing within the **Flutter: Upgrade Packages** list.

Fixes #3013.

### DIFF
--- a/src/shared/utils/fs.ts
+++ b/src/shared/utils/fs.ts
@@ -101,7 +101,7 @@ export async function findProjectFolders(logger: Logger, roots: string[], exclud
 	}));
 	const projectFoldersChecks = await Promise.all(projectFolderPromises);
 	const projectFolders = projectFoldersChecks
-		.filter((res) => res.exists)
+		.filter((res) => res.exists && !res.folder.includes(".dart_tool"))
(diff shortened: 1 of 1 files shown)


src/shared/utils/fs.ts
```diff
@@ -101,7 +101,7 @@ export async function findProjectFolders(logger: Logger, roots: string[], exclud
 	}));
 	const projectFoldersChecks = await Promise.all(projectFolderPromises);
 	const projectFolders = projectFoldersChecks
-		.filter((res) => res.exists)
+		.filter((res) => res.exists && !res.folder.includes(".dart_tool"))
```
DanTup, line 104: We should probably more this to the filter on line 94 (or an additional line next to it) to avoid the `hasPubspec` etc. checks on folders we'd then discard anyway.

We should also probably also put `${path.sep}` on each end just to ensure we only get folders with exactly that segment.

PR #3027, opened 2020-12-21
Add support for Flutter integration tests

BootBlock:
This PR adds support for integration testing but I have some questions.

- Should debugging be supported?
  - If so, how do we get the VM service URL?
- Does Flutter drive support JSON output so we can put the results in the test tree?
- Should we support `attach` (Flutter drive does have `use-existing-app`)?

### DIFF
--- a/src/debug/debug_entry.ts
+++ b/src/debug/debug_entry.ts
@@ -2,6 +2,7 @@ import { DebugSession } from "vscode-debugadapter";
 import { DartDebugSession } from "./dart_debug_impl";
 import { DartTestDebugSession } from "./dart_test_debug_impl";
 import { FlutterDebugSession } from "./flutter_debug_impl";
+import { FlutterIntegrationTestDebugSession } from "./flutter_integration_test_debug_impl";
 import { FlutterTestDebugSession } from "./flutter_test_debug_impl";
 import { WebDebugSession } from "./web_debug_impl";
 import { WebTestDebugSession } from "./web_test_debug_impl";
@@ -13,6 +14,7 @@ const debuggers: { [key: string]: any } = {
 	"dart": DartDebugSession,
 	"dart_test": DartTestDebugSession,
 	"flutter": FlutterDebugSession,
+	"flutter_integration_test": FlutterIntegrationTestDebugSession,
 	"flutter_test": FlutterTestDebugSession,
 	"web": WebDebugSession,
 	"web_test": WebTestDebugSession,
--- a/src/debug/flutter_integration_test_debug_impl.ts
+++ b/src/debug/flutter_integration_test_debug_impl.ts
@@ -1,0 +1,49 @@
+import path = require("path");
+import { TerminatedEvent } from "vscode-debugadapter";
+import { DebugProtocol } from "vscode-debugprotocol";
+import { flutterPath } from "../shared/constants";
+import { FlutterLaunchRequestArguments } from "../shared/debug/interfaces";
+import { LogCategory } from "../shared/enums";
+import { safeSpawn } from "../shared/processes";
+import { DartDebugSession } from "./dart_debug_impl";
+
+export class FlutterIntegrationTestDebugSession extends DartDebugSession {
+	constructor() {
+		super();
+
+		this.allowWriteServiceInfo = false;
+		this.parseVmServiceUriFromStdOut = false;
+		this.logCategory = LogCategory.FlutterTest;
+	}
+
+	protected async attachRequest(response: DebugProtocol.AttachResponse, args: any): Promise<void> {
+		return this.launchRequest(response, args);
+	}
+
+	protected spawnProcess(args: FlutterLaunchRequestArguments): any {
+		// TODO: Do we support this?
+		const isAttach = args.request === "attach";
+		this.allowTerminatingVmServicePid = !isAttach;
+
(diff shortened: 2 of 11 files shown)


DanTup:
@dnfield @jonahwilliams do you know the answers to these?

1. How can we run `flutter drive` in a way that we can get a VM service URL and start paused (so we can add breakpoints before code runs)?
    - For Dart VM we'd use `--enable-vm-service=0 --pause_isolates_on_start=true -DSILENT_OBSERVATORY=true --write-service-info=/tmp/random-file.json`
    - For `flutter run` we'd use `--machine --start-paused` and get the VM Service in the [app.debugPort](https://github.com/flutter/flutter/blob/4d663472e2cf70cf6229d8371dc3381d20e35df9/packages/flutter_tools/doc/daemon.md#appdebugport) message
    - For `flutter test` we also use `--machine --start-paused` and the VM service URL is in the `test.startedProcess` event
2. Can `flutter drive` output the same JSON that `flutter test` does, so we could use it to populate tests in the test tree in the editor?

Thanks!

jiahaog:
The issue today is that `flutter drive` does not understand the semantics of integration tests, which makes implementing debugging support more complicated. Ideally, this would live in the flutter tool as an abstraction for IDE tooling to use – I have been working on https://github.com/flutter/flutter/issues/66264 to solve this :) 

Regarding (2), it is tracked in https://github.com/flutter/flutter/issues/51135. This issue only tracks supporting machine readable output from the host side driver script though. We might also want some form of tooling to propagate the test results that run on the device together in a well formed manner.

DanTup:
@jiahaog how different is using `package:integration_test` to what was available before?

We currently support debugging (both the app and the tests) and having the test results shown in the test tree by bypassing `flutter drive` and just starting two debug sessions (one runs the instrumented app on the device using `flutter run` and the other runs the `package:flutter_driver` tests using `flutter test`).

Would something similar be possible with `package:integration_test`, or do they work too differently?

jiahaog:
Using `flutter run` and `flutter test` to bypass `flutter drive` is certainly quite interesting. I think this flow should work as well for integration tests, just that I'm not sure if it's a good idea for IDE plugins to own the logic to do so.

The main difference between Flutter Driver tests and Integration Tests is that for the former, the tests are part of the host script, and for the latter, the host side script is typically a placeholder script. Users write their assertions in a Dart script which is built into the app which runs on the device. As such, some sort of tooling will be needed to pull the results from the app.

DanTup:
> I'm not sure if it's a good idea for IDE plugins to own the logic to do so.

Yeah, I agree with that. It was a bit of a workaround while there was nothing better (it seems like it required two debug sessions to debug both tests and app). If I understand correctly, the new way won't require two debugging sessions as the tests and app are all running in the same process on the device (which seems like a great change).

Is there any way at all for us to get at the VM service for debugging as things are today? If not, we can force the session into noDebug mode for now (although I think debugging will be a highly-requested feature).

> Users write their assertions in a Dart script which is built into the app which runs on the device. As such, some sort of tooling will be needed to pull the results from the app.

Do these tests still use `package:test`? I think the current output of `flutter_test` is just `package:test`'s JSON reporter, so if that's usable on the device and it prints to stdout (which is routed back through the tool), we may be able to consume that without too much work?

jiahaog:
> Is there any way at all for us to get at the VM service for debugging as things are today?

One thing to try is to scrape the logs of `flutter drive` or explicitly pass the port to `flutter drive` (via `--host-vmservice-port`) to achieve this if support for debugging is needed now.

> Do these tests still use package:test? I think the current output of flutter_test is just package:test's JSON reporter, so if that's usable on the device and it prints to stdout (which is routed back through the tool), we may be able to consume that without too much work?

Yes, integration tests use `package:test`, but the executable portion of package:test (which handles the JSON reporting) isn't used at all. You could write a custom host side driver script that extends the recommended driver to [wire the output](https://github.com/flutter/flutter/blob/1cb0a24a466fb1d1cf8ad95b15c9857ab9817293/packages/integration_test/lib/integration_test_driver.dart#L72) somewhere else as a short term solution.

I've discussed this earlier today with @jonahwilliams, and the approach we're going to try is to implement integration tests as part of `flutter test`. The idea is that instead of running the tests in the flutter_tester, e.g. `flutter test test/foo_test.dart`, `flutter test integration_test/baz_test.dart` would run the tests in a driven app. This will make it much easier for users (and tools like this) to leverage the existing capabilities of `flutter test` for integration tests, and get debugging support for free (or with minimal work). I'm planning to start a POC on this hopefully next week :)

DanTup:
> I've discussed this earlier today with @jonahwilliams, and the approach we're going to try is to implement integration tests as part of flutter test. The idea is that instead of running the tests in the flutter_tester, e.g. `flutter test test/foo_test.dart`, `flutter test integration_test/baz_test.dart` would run the tests in a driven app. 

That sounds perfect, and I think will eliminate the need for this PR (and any follow-up work for handling debugging/test reporting). I won't merge this unless it turns out that won't work (or won't be available anytime soon).

Please let me know if you want me to do some testing from VS Code once you have something. Thanks!

DanTup:
Based on discussion above, closing this in favour of #3133 (WIP) which will recognise `integration_test` as a test folder and generally behave like the `test` folder, except it will pass through the selected device to the debugger and add `-d deviceId` to the args.

PR #3041, opened 2021-01-01
Specified type in dart snippet

Semvrij:
It is a [better practice](https://dart-lang.github.io/linter/lints/always_specify_types.html) to specify type annotations.

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -54,7 +54,7 @@
 			"prefix": "for",
 			"description": "Insert a for loop.",
 			"body": [
-				"for (var i = 0; i < ${1:count}; i++) {",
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks for the PR - however I think this change is quite subjective. The Flutter style guide linked to from the lint is for the Flutter repo itself rather than a global recommendation. Personally, I would not use that lint (although ofc I do follow it when working in the Flutter repo).

I'd like to move these snippets to the language server (they're currently not context-specific so show up in places they should not). Once there, they would be able to query the lints and generate code that takes lints into account, so I think it would be best to make this change as part of that work to avoid this changing for everyone without an option to change back.

PR #3046, opened 2021-01-04
Update package.json

kagouda:
A small modification based on the recent change of the 'New Project' command to 'New Application Project'.


DanTup:
Good catch, thanks!

PR #3106, opened 2021-01-27
Fix typo in Coordindator naming

BootBlock:
Fixes all instances of `coordindator` appearing rather than `coordinator`.


DanTup:
Oops!

PR #3148, opened 2021-02-18
[Flutter] Create widgets Stateless and Stateful with imports

LeonardoRosaa:
I don't know if this feature is util for `Dart-Code`.

Changes: 
 - Creating widgets Stateless and Stateful with imports of [Material](https://api.flutter.dev/flutter/material/material-library.html), [Cupertino](https://api.flutter.dev/flutter/cupertino/cupertino-library.html) or [Widgets](https://api.flutter.dev/flutter/widgets/Widget-class.html) libraries.

### DIFF
--- a/snippets/flutter.json
+++ b/snippets/flutter.json
@@ -14,6 +14,55 @@
 				"}"
 			]
 		},
+		"Flutter stateless widget material": {
+			"prefix": "stlessm",
+			"description": "Insert a StatelessWidget with material import",
+			"body": [
+				"import 'package:flutter/material.dart';",
+				"",
+				"class $1 extends StatelessWidget {",
+				"  @override",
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks for the contribution! Although the goal is to not add any more snippets here, with the intention to move snippets to the language server instead. The snippets here show up in _all_ contexts, even where they are not valid, which pollutes the completion list.

By moving them to the server (https://github.com/dart-lang/sdk/issues/37831), they can be context-sensitive, and the imports could also be automatically inserted when required, without needing special snippets for them. This means you could use the snippet in the middle of a file and get the imports (with them inserted into the correct place).

Ofc, if these snippets are useful to you in the meantime, you could include them in your VS Code User Snippets file and they should still show up for you.

I hope this makes sense!

Issue #3308, opened 2021-04-28
Allow void to be coloured like a type


tjx666:
![image](https://user-images.githubusercontent.com/41773861/116456669-191f3080-a895-11eb-8c33-b80d1bf3575e.png)

UmairJibran:
according to the latest version of the file, `void` and `var` are categorized as `preemptive type`, const has been moved to `modifier` and the `String` is highlighted as a normal class.

UmairJibran:
However, I have created a pull request with the requested changes, let's see if they accept it or not.

DanTup:
Sorry for the delay! I think `void` is is a little different to a normal type (for example there are places where you cannot put `void` but could put a type). I'm not sure if colouring it like a type is more correct than as a keyword, and some people may prefer it as is (I do - I like being able to see `void` at a glance and know whether a function is returning a value or not).

If there's demand for being able to colour it like a type, perhaps it would be better to add a modifier to the semantic token so that it can be customised?

@UmairJibran I'll comment on the PR about your change - I'm not sure I entirely understand the goal - although I will note that if we do make changes here, they should also be reflected in LSP semantic tokens, otherwise you'll see them load as one colour, and then change to another colour once the semantic tokens arrive.

nbros:
I know `void` is a type in Dart but I still see it as the _absence_ of a type, so I like it colored this way too (like a keyword).

tjx666:
I see that `void` is colorized as type in typescript file.

DanTup:
@tjx666 I think there are valid arguments for colouring it both ways. Given at least _some_ people prefer it to be coloured like a keyword and that's what people are already used to, I think any change should be opt-in.

Adding a semantic token modifier will allow anyone who wants it to be coloured like a type to change its colour, and the default would remain as-is.

DanTup:
I landed a change in the SDK in https://github.com/dart-lang/sdk/commit/3690f549b02f11787358ed6e12df96b55dcceaf7 that adds a `void` modifier to the `keyword` token, so it's now possible to customise the colour of the `void` keyword in your VS Code settings:

```js
"editor.semanticTokenColorCustomizations": {
	"rules": {
		"keyword.void": {
			"foreground": "#ffff99",
			"fontStyle": "underline"
		}
	}
},
```

[...]

PR #3316, opened 2021-05-02
add syntax Highlighting as "modifier"

UmairJibran:
This Pull Request is in response to issue #3308 

Add Syntax Highlighting as Modifier for the following data types/keywords
- bool
- String
- int
- void
- var

### DIFF
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -305,7 +305,7 @@
 					"name": "keyword.operator.logical.dart"
 				},
 				{
-					"match": "(?<!\\$)\\b(static|final|const|required|late)\\b(?!\\$)",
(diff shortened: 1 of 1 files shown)


DanTup:
Can you clarify exactly what the intended change here is? It currently causes `bool` `String` and `int` to change from being coloured like types to being coloured like keywords (something I don't think is a good change). It also duplicates `void` and `var` into the "modifier" section when they currently have their own section (directly below). It doesn't make sense to duplicate them, but I'm also unsure of the advantage of folding them into another category (currently they can be coloured differently, and this would prevent that).

UmairJibran:
Sorry, it's a schoolboy error from my end, I'll close this pull request to avoid further confusion.
Thanks for clarifying it to me.

Issue #3315, opened 2021-05-02
Support running emulators when running in remote/container environments


Delgan:
**Describe the bug**
Hi!
I'm using the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension with the Flutter' one on Linux. I can successfully configure Flutter, create an emulator, launch it, and run my application. 

The problem: when I press `F5` the existing device is not listed and there is no "Create Android emulator" option.
Also, when I press `Ctrl`+`Shift`+`P` and search for "Flutter: Launch Emulator" there is no such command.

Note that if I create the emulator with `flutter emulators --create` and then start it with `flutter emulators --launch` then I'm able to run my application using `F5` and it's working perfectly fine. However, the Flutter extension won't mention the device if it is not launched, even after restarting VSCode (although it's somehow detected by the Flutter daemon according to logs below).

**To Reproduce**
1. Open a local Flutter project
2. Install `Remote Development` VSCode extension
3. Use the following `Dockerfile` to configure `Remote - Containers` extension (important: `KVM_GID` needs to be modified):
```Dockerfile
FROM ubuntu:20.04

ARG USERNAME=vscode
ARG USER_UID=1000
ARG USER_GID=$USER_UID

# The container is configured for non-root user but we need to access "/dev/kvm" to run emulators.
# We have to map the group owner of "/dev/kvm` from host (which is "kvm") to inside the container.
# So we'll create a new "kvm" group with the id from host returned in "cat /etc/group | grep kvm"). 
ARG KVM_GID=108

RUN groupadd --gid $USER_GID $USERNAME \
    && useradd --uid $USER_UID --gid $USER_GID -m $USERNAME \
    && apt-get update \
	&& DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        sudo \
        openjdk-8-jdk-headless \
	    wget curl git xz-utils zip unzip libglu1 \
        python3 python3-dev python3-venv python3-pip python3-tk \
        libpulse-dev libxcomposite-dev libxcursor-dev libasound2-dev kmod qemu-kvm \
        ssh vim locate \
	&& apt-get autoremove -y \
	&& apt-get clean -y \
	&& rm -rf /var/lib/apt/lists/* \
    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME \
[...]

PR #3352, opened 2021-05-18
Support launching emulator devices in remote workspaces

Delgan:
Hey @DanTup.

I just saw your fix for #3315. Thanks for that!

I also noticed the "Launch Emulator" command wasn't available in remote workspace. I guess this is because of the `when` clause in `"package.json"`: https://github.com/Dart-Code/Dart-Code/blob/f38e02c2e3384975d4d051484836a7984e31d64d/package.json#L987-L988

I supposed this simple additional condition using the new configuration parameter would allow to run the command in a devcontainer.

Thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -985,7 +985,7 @@
 				},
 				{
 					"command": "flutter.launchEmulator",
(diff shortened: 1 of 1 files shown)


DanTup:
Good spot, thanks!

PR #3371, opened 2021-05-31
Match latest flutter.json

helin24:
This is to clean up VSCode's view of flutter.json, which has changed significantly since this was last updated. This simplifies DevTools startup since we no longer have a separate script to start DevTools; this may change in the future to either using the daemon (see comments for more details) or another method.

### DIFF
--- a/src/extension/sdk/dev_tools/manager.ts
+++ b/src/extension/sdk/dev_tools/manager.ts
@@ -53,17 +53,6 @@ export class DevToolsManager implements vs.Disposable {
 	}
 
 	private async handleEagerActivationAndStartup(workspaceContext: DartWorkspaceContext) {
-		if (workspaceContext.config?.activateDevToolsEagerly) {
-			try {
-				await this.preActivate(true);
-				this.logger.info(`Finished background activating DevTools`);
-			} catch (e) {
-				this.logger.error("Failed to background activate DevTools");
-				this.logger.error(e);
-				vs.window.showErrorMessage(`Failed to activate DevTools: ${e}`);
-			}
-		}
-
 		if (workspaceContext.config?.startDevToolsServerEagerly) {
 			try {
 				if (workspaceContext.config?.startDevToolsServerEagerly)
--- a/src/extension/sdk/dev_tools/manager.ts
+++ b/src/extension/sdk/dev_tools/manager.ts
@@ -11,7 +11,7 @@ import { DartWorkspaceContext, DevToolsPage, Logger, SomeError } from "../../../
 import { CategoryLogger } from "../../../shared/logging";
 import { UnknownNotification } from "../../../shared/services/interfaces";
 import { StdIOService } from "../../../shared/services/stdio_service";
-import { disposeAll, usingCustomScript } from "../../../shared/utils";
+import { disposeAll } from "../../../shared/utils";
 import { getRandomInt } from "../../../shared/utils/fs";
 import { waitFor } from "../../../shared/utils/promises";
 import { envUtils, isRunningLocally } from "../../../shared/vscode/utils";
@@ -78,7 +78,7 @@ export class DevToolsManager implements vs.Disposable {
 	}
 
 	private async preActivate(silent: boolean): Promise<void> {
-		this.devToolsActivationPromise = this.pubGlobal.backgroundActivate(devtoolsPackageName, devtoolsPackageID, silent, this.workspaceContext.config?.devtoolsActivateScript);
+		this.devToolsActivationPromise = this.pubGlobal.backgroundActivate(devtoolsPackageName, devtoolsPackageID, silent);
 		await this.devToolsActivationPromise;
 	}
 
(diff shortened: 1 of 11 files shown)


src/extension/sdk/dev_tools/manager.ts
```diff
@@ -98,22 +98,18 @@ export class DevToolsManager implements vs.Disposable {
 
 		if (!this.devtoolsUrl) {
 			this.devToolsStatusBarItem.hide();
-			// Don't try to check for install/version if we're using a custom script.
-			if (!this.workspaceContext.config?.devtoolsActivateScript) {
```
DanTup, line 102: I presume the removal of this means it's now ok to use `pub global activate devtools` and `pub global run devtools` for these workspaces?

src/extension/sdk/dev_tools/manager.ts
```diff
@@ -420,11 +415,9 @@ class DevToolsService extends StdIOService<UnknownNotification> {
 
 		this.spawnedArgs = this.getDevToolsArgs();
 
-		const { binPath, binArgs } = usingCustomScript(
-			path.join(workspaceContext.sdks.dart, pubPath),
-			this.spawnedArgs,
-			workspaceContext.config?.devtoolsRunScript,
-		);
+		// TODO(helin24): Use daemon instead to start DevTools if internal workspace
```
DanTup, line 418: Curious what "daemon" is in this context (or more specifically what its lifetime is and whether it allows us to still use the devtools_server?).

src/shared/utils/workspace.ts
```diff
@@ -54,13 +54,8 @@ export function tryProcessBazelFlutterConfig(logger: Logger, config: WritableWor
...
 
-		config.activateDevToolsEagerly = !!flutterConfig.devtoolsActivateScript;
-		config.dartSdkHomeLinux = makeFullPath(flutterConfig.dartSdkHome?.linux);
-		config.dartSdkHomeMac = makeFullPath(flutterConfig.dartSdkHome?.macos);
-		// This one replaces the args "global activate devtools"
-		config.devtoolsActivateScript = makeScript(flutterConfig.devtoolsActivateScript, 3);
-		// This one replaces the args "global run devtools"
-		config.devtoolsRunScript = makeScript(flutterConfig.devtoolsRunScript, 3);
+		// TODO(helin24): Check where this is used
+		config.activateDevToolsEagerly = false;
```
DanTup, line 58: IIRC this was because the activation script for internal workspaces was slow, so we did it eagerly at startup. If it's not just using normal devtools and is not slow, this field can probably be entirely removed.

DanTup:
Some of the test failures are because of the VM printing that it started DevTools (waiting on https://dart-review.googlesource.com/c/sdk/+/199960 to land to be able to disable this). I've pushed 05135219a9f1ef87408b36b9c970b7a4e777fdce to `master` in the meantime to make the builds green, so if you rebase you shouldn't get all the unrelated failures (I'll restore those tests in future after the new switch lands).

src/extension/sdk/dev_tools/manager.ts
helin24: We can, although optimally for internal applications we can use `daemonScript` to launch a flutter daemon and then request DevTools to start from there. There are some changes planned for DevTools launch more generally, so I don't think this is a priority and perhaps we should continue using `pub global` for now.

src/extension/sdk/dev_tools/manager.ts
helin24: This is related to the above comment - externally DevTools can be started by first running `flutter daemon` and then passing a message like `[{"message":"devtools.launch"}]`. Then the daemon will print out the address of the DevTools server and this can be used the same as when DevTools is launched with `pub global`. (Above I mentioned `daemonScript` which does the same thing as `flutter daemon` internally). I may need to know more about how `devtools_server` is used to determine whether everything would be the same otherwise.

src/shared/utils/workspace.ts
helin24: Do you mean startup of the workspace or of the app? I may leave this here for now because I want to think more about how we want to launch DevTools internally and this may still be relevant if we decide to use the internal daemon script.

src/extension/sdk/dev_tools/manager.ts
DanTup: Ah got it - I wasn't sure if the daemon you meant was similar to the `flutter run` daemon, which would be more tricky as it has a different lifetime (it's once per debug session, whereas the DevTools server and `flutter daemon` can live for the whole editor session).

Right now devtools_server is used for some additional functionality like re-connecting DevTools instances when you start a new debug session. We can't currently run DevTools without the server, although it's possible a lot of its functionality could be migrated away or dropped. For example, embedded views we can refresh ourselves to reconnect them to new apps - this isn't the case for external browser windows we spawn though.

If using the Pub version for now works, sounds like this doesn't need resolving immediately though.

src/shared/utils/workspace.ts
DanTup: The workspace. Originally we would only try activate DevTools right before we needed to run it (which we did when the user asked for it). My understanding was that the internal activate script was too slow to run then so this flag promoted it to being run at workspace activation (so it should be ready before the user runs a debug session or tries to run DevTools).

Things have changed a little since in that now we try to _run_ DevTools eagerly too, so we can pass it to Flutter (`--devtools-server-address`) so it's possible this could be folded into that anyway (at least for Flutter apps).

helin24: Okay I see. After taking a second look I think it makes sense to remove this `activateDevToolsEagerly` check. If I want to run DevTools eagerly it looks like I should make sure `startDevToolsServerEagerly` is set to true - does this require a change within the workspace file or elsewhere?

DanTup: `startDevToolsServerEagerly` should already be set to true for Flutter projects during activation:

https://github.com/Dart-Code/Dart-Code/blob/8ccb1ab3211d3a9c43a4e3b59da60d7dc2decb73/src/extension/extension.ts#L200-L207

It might be that your workspace isn't currently being detected as Flutter based on other comments - if that will be fixed you should get this for free, but otherwise could set it explicitly in `tryProcessBazelFlutterConfig` with the other fields.

DanTup (approved):
LGTM!

PR #3378, opened 2021-06-02
Enable device daemon for bazel workspace (WIP)

helin24:
I think for now it'll be okay to assume that any internal workspace is a Flutter workspace, since the added functionality (showing device menu in this case) doesn't interfere with other workspaces. We may want to change this later to require an indication of Flutter in a `.code-workspace` file.

The internal "Flutter SDK" path isn't really a path with a binary to run Flutter, since we rely on scripts to execute the equivalent  flutter commands instead, but it is necessary that we are in the right working directory to execute the scripts. Versioning isn't important since we can assume all internal users will be on a close to the latest version or will update if not.

The script for running a daemon to find devices can have stdout that results in lines like: `[3:27:46 PM] [FlutterDaemon] [Error] ` in the debug console for the extension - I'm not sure where this is coming from and why the stdout from this process is recognized as errors.

I think some similar changes are required to do things like show the flutter outline and show "Flutter" in the spot to the left of the device menu, so let me know if there's a better place to put this logic where all of these activities could be handled gracefully.

### DIFF
--- a/src/extension/sdk/utils.ts
+++ b/src/extension/sdk/utils.ts
@@ -243,43 +243,49 @@ export class SdkUtils {
 				this.logger.info(`Found Fuchsia project that is not vanilla Flutter`);
 		}
 
-		const flutterSdkSearchPaths = [
-			workspaceConfig?.flutterSdkHome,
-			config.flutterSdkPath,
-			// TODO: These could move into processFuchsiaWorkspace and be set on the config?
-			fuchsiaRoot && path.join(fuchsiaRoot, "lib/flutter"),
-			fuchsiaRoot && path.join(fuchsiaRoot, "third_party/dart-pkg/git/flutter"),
-			firstFlutterMobileProject,
-			firstFlutterMobileProject && extractFlutterSdkPathFromPackagesFile(firstFlutterMobileProject),
-			firstFlutterMobileProject && path.join(firstFlutterMobileProject, ".flutter"),
-			firstFlutterMobileProject && path.join(firstFlutterMobileProject, "vendor/flutter"),
-			process.env.FLUTTER_ROOT,
-			isLinux ? "~/snap/flutter/common/flutter" : undefined,
-			"~/flutter-sdk",
-			"~/.flutter-sdk",
-		].concat(paths).filter(notUndefined);
+		let flutterSdkPath;
+		if (workspaceConfig.flutterBazel) {
+			hasAnyFlutterProject = true;
+			hasAnyFlutterMobileProject = true;
+			flutterSdkPath = workspaceConfig?.flutterSdkHome;
+		} else {
+			const flutterSdkSearchPaths = [
+				config.flutterSdkPath,
+				// TODO: These could move into processFuchsiaWorkspace and be set on the config?
+				fuchsiaRoot && path.join(fuchsiaRoot, "lib/flutter"),
+				fuchsiaRoot && path.join(fuchsiaRoot, "third_party/dart-pkg/git/flutter"),
+				firstFlutterMobileProject,
+				firstFlutterMobileProject && extractFlutterSdkPathFromPackagesFile(firstFlutterMobileProject),
+				firstFlutterMobileProject && path.join(firstFlutterMobileProject, ".flutter"),
+				firstFlutterMobileProject && path.join(firstFlutterMobileProject, "vendor/flutter"),
+				process.env.FLUTTER_ROOT,
+				isLinux ? "~/snap/flutter/common/flutter" : undefined,
+				"~/flutter-sdk",
+				"~/.flutter-sdk",
+			].concat(paths).filter(notUndefined);
 
-		let flutterSdkResult = this.findFlutterSdk(flutterSdkSearchPaths);
+			let flutterSdkResult = this.findFlutterSdk(flutterSdkSearchPaths);
 
-		// Handle the case where the Flutter snap has not been initialised.
-		if (!flutterSdkResult.sdkPath && flutterSdkResult.candidatePaths.includes(snapBinaryPath)) {
-			// Trigger initialization.
-			this.logger.info(`No Flutter SDK found, but a 'flutter' binary did point at ${snapBinaryPath} so attempting to initialize...`);
-			await initializeFlutterSdk(this.logger, snapFlutterBinaryPath);
+			// Handle the case where the Flutter snap has not been initialised.
+			if (!flutterSdkResult.sdkPath && flutterSdkResult.candidatePaths.includes(snapBinaryPath)) {
+				// Trigger initialization.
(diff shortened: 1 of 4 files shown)


DanTup:
> The internal "Flutter SDK" path isn't really a path with a binary to run Flutter, since we rely on scripts to execute the equivalent flutter commands instead

If the path isn't a real SDK, won't it break the things that don't have custom scripts? For example, if you modified a `pubspec.yaml` we may try to run `(path)\flutter pub get` when you save the file?

> Versioning isn't important since we can assume all internal users will be on a close to the latest version or will update if not.

Right now if we don't have a valid version, we may assume version 0.0.0 (and therefore _not_ enable any new functionality). I think previously we had a `versionFile` on the workspace config, but if there isn't a real file perhaps we should add something like a `flutterVersion` (which could be set really high) or a flag to assume the latest version (which we could pass to the DartCapabilities/FlutterCapabilities and have the getters in them handle that).

src/shared/interfaces.ts
```diff
@@ -53,6 +53,7 @@ export interface WritableWorkspaceConfig {
 	flutterVersionFile?: string;
 	useLsp?: boolean;
 	useVmForTests?: boolean;
+	flutterBazel?: boolean;
```
DanTup, line 56: I think it may be better to name this based on what it does rather than what triggered it - for example `enableFlutterMode` or `forceFlutterMode` or similar. This would be a little more generic and might be usable for other workspace types (for example it might allow simplifying some of the other pubspec checks into the more specific workspace functions).

src/extension/sdk/utils.ts
```diff
@@ -243,43 +243,49 @@ export class SdkUtils {
...
-
-			// Then search again.
-			this.logger.info(`Snap initialization completed, searching for Flutter SDK again...`);
-			flutterSdkResult = this.findFlutterSdk(flutterSdkSearchPaths);
-		}
+		let flutterSdkPath;
+		if (workspaceConfig.flutterBazel) {
+			hasAnyFlutterProject = true;
+			hasAnyFlutterMobileProject = true;
+			flutterSdkPath = workspaceConfig?.flutterSdkHome;
```
DanTup, line 250: Setting these probably gets the Flutter functionality enabled, but there might also need to be some changes to `isInFlutterProject`(?) to trigger the Flutter paths for specific files (eg. when launching projects, or running commands like Get Packages that choose between Dart/Flutter).

helin24:
> If the path isn't a real SDK, won't it break the things that don't have custom scripts? For example, if you modified a pubspec.yaml we may try to run (path)\flutter pub get when you save the file?

Yes, I'll have to look at functions that use the SDK and make sure there are alternative scripts/logic for the bazel case. I'm not sure how messy this would be though. It may make sense to add an internal script that acts like the SDK and reroutes to appropriate actions or returns something reasonable for actions that aren't supported, but there are probably cases where adding some additional logic to the extension is unavoidable.

> perhaps we should add something like a flutterVersion (which could be set really high) or a flag to assume the latest version (which we could pass to the DartCapabilities/FlutterCapabilities and have the getters in them handle that)

I think either of these would be good options. Do you think it's important to keep the `workspaceConfig.flutterVersionFile` property?

src/extension/sdk/utils.ts
helin24: I'm not sure I understand the context here - I'll try to schedule time with you tomorrow to discuss in real time.

src/extension/sdk/utils.ts
```diff
@@ -243,43 +243,49 @@ export class SdkUtils {
...
+			flutterSdkPath = flutterSdkResult.sdkPath;
+		}
 
 		// Since we just blocked on a lot of sync FS, yield.
 		await resolvedPromise;
 
 		// If we're a Flutter workspace but we couldn't get the version, try running Flutter to initialise it first.
 		// Do this before searching for the Dart SDK, as it might download the Dart SDK we'd like to find.
-		if (hasAnyFlutterProject && flutterSdkPath) {
+		if (hasAnyFlutterProject && flutterSdkPath && !workspaceConfig.forceFlutterMode) {
```
DanTup, line 288: In hindsight this new name seems less good here. No need to change in this PR, but maybe it's worth adding an additional flag to suppress Flutter initialisation and also set it for the Bazel workspace config?

DanTup (approved):

DanTup:
> The script for running a daemon to find devices can have stdout that results in lines like: `[3:27:46 PM] [FlutterDaemon] [Error] ` in the debug console for the extension

I missed this previously - did you figure it out? Is it happening only with the Bazel workspace daemon? Is there anything after the `[Error]` part? The `FlutterDaemon` class uses `StdIOService` which logs `stderr` like that:

https://github.com/Dart-Code/Dart-Code/blob/3af8a9c119a7adc4b4a59df4c630d149ed136e11/src/shared/services/stdio_service.ts#L71

It also logs a few other things like that, but usually invalid JSON messages.

DanTup:
@helin24 the test failure I saw was because we used the Flutter SDK as the working directory when spawning the daemon. I don't think there's any good reason for that and it failed here because `my-flutter-sdk` isn't a real Flutter SDK.

If you rebase to include bc8364d5d3c20b384f98263dc3d24280a7929410 I think it should go green.

DanTup:
On second thoughts - you may have mentioned you needed the working directory set to that? If so, then we should revert bc8364d5d3c20b384f98263dc3d24280a7929410 and just create the `my-flutter-sdk` folder in the repo (might need a placeholder file in it) instead.

helin24:
> The `FlutterDaemon` class uses `StdIOService`
> It also logs a few other things like that, but usually invalid JSON messages.

Ah I see. I think it may be that JSON is expected but the internal script logs non-JSON as it's starting up the daemon. Once the daemon is started then the output is JSON.

DanTup:
There's a (slightly strangely-named) flag `treatHandlingErrorsAsUnhandledMessages` that will prevent non-JSON messages from being treated as errors. Perhaps setting that to `true` in the `FlutterDaemon` subclass will resolve that.

helin24:
> There's a (slightly strangely-named) flag `treatHandlingErrorsAsUnhandledMessages` that will prevent non-JSON messages from being treated as errors. Perhaps setting that to `true` in the `FlutterDaemon` subclass will resolve that.

I looked at this more closely and it seems there are just some internal printouts that are unexpectedly stderr, so I think any changes I make for this will be internal and not part of the extension.

helin24:
I'm not able to reproduce the remaining test failures locally, so merging this for now.

DanTup:
lgtm! I checked the failures on master, and there were three issues:

1. VS Code crashing at startup
2. NPM complaining of a corrupt TypeScript download
3. Test failures on Windows where we tried to clean up (killFlutterTester) but hadn't activated the extension

The first two are just unrelated flakes. The last one I'm not sure why it didn't always fail, but I've made `killFlutterTester` just bail out early in the case of not having an active extension (68e53562).

PR #3397, opened 2021-06-09
Set bazel project flutter SDK version to latest

helin24:
The version file isn't needed for bazel projects, so just setting it to latest. I may be able to remove the `flutterVersionFile` completely.

### DIFF
--- a/src/extension/sdk/utils.ts
+++ b/src/extension/sdk/utils.ts
@@ -286,7 +286,8 @@ export class SdkUtils {
 		// If we're a Flutter workspace but we couldn't get the version, try running Flutter to initialise it first.
 		// Do this before searching for the Dart SDK, as it might download the Dart SDK we'd like to find.
 		if (hasAnyFlutterProject && flutterSdkPath && !workspaceConfig.skipFlutterInitialization) {
-			const flutterNeedsInitializing = !getSdkVersion(this.logger, { sdkRoot: flutterSdkPath, versionFile: workspaceConfig?.flutterVersionFile, isLatest: workspaceConfig?.isFlutterVersionLatest })
+			const flutterVersion = workspaceConfig?.flutterVersion ?? getSdkVersion(this.logger, { sdkRoot: flutterSdkPath });
+			const flutterNeedsInitializing = !flutterVersion
 				|| !fs.existsSync(path.join(flutterSdkPath, "bin/cache/dart-sdk"));
 
 			if (flutterNeedsInitializing)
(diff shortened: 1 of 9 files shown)


src/extension/sdk/utils.ts
```diff
@@ -286,7 +286,7 @@ export class SdkUtils {
 		// If we're a Flutter workspace but we couldn't get the version, try running Flutter to initialise it first.
 		// Do this before searching for the Dart SDK, as it might download the Dart SDK we'd like to find.
 		if (hasAnyFlutterProject && flutterSdkPath && !workspaceConfig.skipFlutterInitialization) {
-			const flutterNeedsInitializing = !getSdkVersion(this.logger, { sdkRoot: flutterSdkPath, versionFile: workspaceConfig?.flutterVersionFile })
+			const flutterNeedsInitializing = !getSdkVersion(this.logger, { sdkRoot: flutterSdkPath, versionFile: workspaceConfig?.flutterVersionFile, isLatest: workspaceConfig?.isFlutterVersionLatest })
 				|| !fs.existsSync(path.join(flutterSdkPath, "bin/cache/dart-sdk"));
```
DanTup, line 290: It may be better to skip the call to `getSdkVersion` here and just use the flag directly - or even return the version directly in the workspace config instead of a flag:

```ts
const flutterVersion = workspaceConfig?.flutterVersion ?? getSdkVersion(...);
const flutterNeedsInitializing = !flutterVersion || !fs.existsSync(path.join(flutterSdkPath, "bin/cache/dart-sdk");
```

src/shared/constants.ts
```diff
@@ -204,3 +204,5 @@ export const defaultLaunchJson = JSON.stringify(
 	},
 	undefined, "\t"
 );
+
+export const LATEST_SDK_VERSION = "latest";
```
DanTup, line 208: It's probably worth a comment on this explaining why it's used (and what the general behaviour should be when it's used).

We could also name this something like 'MAX_VERSION' and set it to something high like `999.999` to avoid having to special-case in `versionIsAtLeast`. I'm not certain if that's better, but it could avoid accidentally trying to treat the string as a version number elsewhere that could fail.

DanTup (commented):
> I may be able to remove the `flutterVersionFile` completely.

Seems reasonable to me.

Added a few nits + there's a lint failure (need to use triple-equals). You might need the ESLint VS Code extension for the lints to show up in the editor.

src/shared/constants.ts
helin24: `MAX_VERSION` is fine with me. I initially wanted something that would be readable for display but just changed to checking for matching to `MAX_VERSION` instead.

helin24:
Oh it was a lint failure! I saw the bottommost error about a file not existing and was confused. Thanks for the eslint suggestion, installed!

DanTup (approved):
> Oh it was a lint failure! I saw the bottommost error about a file not existing and was confused.

Seems like GH actions upload-artifact doesn't support silently skipping when folders don't exist (folders don't exist if the tests never ran). I've added a workaround in d0e7f385323c7a4c024c7c831d959532c3d72c13 that will ensure the folders exist that should hopefully make it clearer in these cases.

The failing bot about coverage I believe is unrelated - we're currently enabling LSP randomly for 50% of people and I'm not handling this properly on the bot (that is, the LSP bot is forcing LSP bot, and the non-LSP bot is getting a 50/50 chance to enable it). I'll fix that up on master.

This all LGTM!

PR #3424, opened 2021-06-22
Switch arg order for rudimentary bazel run

helin24:
I needed only minor changes for a basic bazel run app configuration to work, but there are probably other situations where bazel will differ from a standard app due to things like:
- Excluding some arguments
- Arguments having a different flag name
- I think the change in ordering I made for where the target is passed is okay for standard apps too, but if not, argument order may need to be adjustable

Maybe some of these differences could be codified in the workspace, e.g. have an 'arguments' JSON that looks like:
```
{"arguments": [
  {"device": {}},
  {"mode": {"replaceFlagName": "otherMode"}},
  {"target": {"excludeFlagName": "true"}
] }
```

Alternatively, I could add an internal script that translates from standard arguments to the arguments we need. That would be more flexible but seems like it might be tough to make sure we're alerted to changes.

I'm missing something for getting debugging breakpoints to work; may need to hook up something additional for that.

### DIFF
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -150,7 +150,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 			&& !isInsideFolderNamed(debugConfig.program, "tool")
 			&& !isInsideFolderNamed(debugConfig.program, ".dart_tool")) {
 			// Check if we're a Flutter or Web project.
-			if (isFlutterProjectFolder(debugConfig.cwd)) {
+			if (isFlutterProjectFolder(debugConfig.cwd) || this.wsContext.config.forceFlutterMode) {
 				debugType = DebuggerType.Flutter;
 			} else if (isInsideFolderNamed(debugConfig.program, "web") && !isInsideFolderNamed(debugConfig.program, "test"))
 				debugType = DebuggerType.Web;
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -150,7 +150,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 			&& !isInsideFolderNamed(debugConfig.program, "tool")
 			&& !isInsideFolderNamed(debugConfig.program, ".dart_tool")) {
 			// Check if we're a Flutter or Web project.
-			if (isFlutterProjectFolder(debugConfig.cwd) || this.wsContext.config.forceFlutterMode) {
+			if (isFlutterProjectFolder(debugConfig.cwd) || this.wsContext.config.forceFlutterDebug) {
 				debugType = DebuggerType.Flutter;
 			} else if (isInsideFolderNamed(debugConfig.program, "web") && !isInsideFolderNamed(debugConfig.program, "test"))
 				debugType = DebuggerType.Web;
--- a/src/debug/flutter_debug_impl.ts
+++ b/src/debug/flutter_debug_impl.ts
@@ -184,10 +184,6 @@ export class FlutterDebugSession extends DartDebugSession {
 
 	protected spawnRunDaemon(isAttach: boolean, args: FlutterLaunchRequestArguments, logger: Logger): RunDaemonBase {
 		let appArgs = [];
-		if (!isAttach || args.program) {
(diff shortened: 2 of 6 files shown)


src/extension/providers/debug_config_provider.ts
```diff
@@ -150,7 +150,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 			&& !isInsideFolderNamed(debugConfig.program, "tool")
 			&& !isInsideFolderNamed(debugConfig.program, ".dart_tool")) {
 			// Check if we're a Flutter or Web project.
-			if (isFlutterProjectFolder(debugConfig.cwd)) {
+			if (isFlutterProjectFolder(debugConfig.cwd) || this.wsContext.config.forceFlutterMode) {
```
DanTup, line 153: With this change, the `forceFlutterMode` flag is doing now doing two things:

- Forcing the _workspace_ into Flutter mode (this activates the daemon, enables Flutter commands)
- Forcing any debug session to use the Flutter debug adapter

These are subtly different, since a Flutter workspace doesn't usually mean the Flutter debugger would always be used (for example if you run a `tool/foo.dart` in a Flutter project, normally it would run with the Dart debugger - but this will cause it to run with Flutter).

I suspect that's intended, though it's worth adding a comment to `forceFlutterMode` to indicate this behaviour (or splitting it into two flags, one for forcing the workspace mode, and one for forcing the debug type) to make it clear.

DanTup (approved):
LGTM!

Test failures are unrelated - the dev builds started failing recently (perhaps a Dart/Flutter change), I'm taking a look at them today. Feel free to merge this if it's ready.

src/extension/providers/debug_config_provider.ts
helin24: I see, I think it makes sense to split it up here then. I may eventually want to add some other checks for whether to use Flutter mode during debug.

DanTup:
@helin24 looking at `spawnRunDaemon`, I'm thinking maybe it's doing too much in the debug adapter. Things in like `deviceId`, `flutterMode`, `trackWidgetCreation`, `flutterPlatform` etc. should maybe be injected into the `args` while in the client (in `DebugConfigProvider`) and the debug adapter only deal with the things specific to connecting the debugger (things like `--start-paused`).

This would simplify the debug adapter a little, and it's likely in the future the adapter (flutter_debug_impl.ts) will be moved to Flutter tools, so the more generic and less specific it is, the better (it would allow other clients using it to make their own decisions about these args, rather than relying on these made up flags that are mapped to switches).

If we did that, I think it would be simpler to build different args for Bazel workspaces (with some tidying up of `DebugConfigProvider`, perhaps there could be a subclass of that for Bazel which is registered instead, and just overrides the building of the args).

I don't know if that would solve the entire issue though - we'd still be selecting a few args in the debug adapter like `--start-paused` (based on the `noDebug` setting). Would you need to change that too? (If so, then maybe subclassing `FlutterDebugSession` would make sense instead, and a new entry in `DebuggerType`?).

DanTup:
The issue with breakpoints in debug adapters not being hit after opening a new workspace turns out to be caused by a VS Code issue:

https://github.com/microsoft/vscode/issues/123508

We use an environment variable to tell the extension to use a server-mode version of the debug adapter we started (which the main VS Code instance has a debugger connected to), but that's being lost when the extension host restarts (which happens when you open a new folder).

A temporary workaround (that should not be committed) to avoid constantly restarting debug sessions, you can comment out the condition around this code so that it always runs regardless of the env variable. This will prevent debugging from working if you launch just the "Extension" launch config because those servers won't be running - do this only while using the "Extension (Debuggable) + DAs" config:

https://github.com/Dart-Code/Dart-Code/blob/9732feed03f954fe82725e011b0846d3f27d3df6/src/extension/providers/debug_adapter_descriptor_factory.ts#L16-L20

DanTup:
> The issue with breakpoints in debug adapters not being hit after opening a new workspace turns out to be caused by a VS Code issue:
>
> microsoft/vscode#123508

FYI it looks like this has been fixed for the next VS Code release (and available in Insiders already).

helin24:
Great! I've gotten used to switching workspaces the way you recommended, but I can try it again once it's released.

PR #3441, opened 2021-06-29
Add css color decoration support

Semvrij:
Added color decoration support for [CSS Colors](https://pub.dev/packages/css_colors).


DanTup:
@devoncarew ultimately I'd like to move this to the server (https://github.com/dart-lang/sdk/issues/44682), but I think it's worthwhile having here meantime. Do you think it's worth adding to tools_metadata, and then using that to copy here so IntelliJ/Android Studio could use them too?

devoncarew:
> @devoncarew ultimately I'd like to move this to the server ([dart-lang/sdk#44682](https://github.com/dart-lang/sdk/issues/44682)), but I think it's worthwhile having here meantime. Do you think it's worth adding to tools_metadata, and then using that to copy here so IntelliJ/Android Studio could use them too?

Yeah, given that this is shipped by flutter.dev, I think it's worth updating tools_metadata for it. Also, +1 to longer term making changes so we can support the `textDocument_documentColor` API.

DanTup:
@Semvrij thanks for the PR! As eluded to above, the source of these colour mappings is https://github.com/flutter/tools_metadata, which is shared with other editors to allow us all to show the same previews.

Are you interested in sending a PR there? There's [a Dart script](https://github.com/flutter/tools_metadata/blob/master/tool/colors/update_colors.dart) that extracts colours from the source files using a regex and builds some resource files from it.

If not, I'm happy to do it and then we can update this PR to update the existing JSON file (we'll still need the regex change to find them though, until the analysis server is able to provide this info  as described in the issue linked above). Thanks!

Semvrij:
Hi @DanTup,

I am not quite familiar with flutter's tools_metadata. Can you give me some more information on how to implement it? Otherwise, it might be better if you could do it.

DanTup:
@Semvrij there are two scripts in the tools_metadata repo:

- [tool/colors/update_colors.dart](https://github.com/flutter/tools_metadata/blob/master/tool/colors/update_colors.dart) extacts colours from the Flutter SDK files and writes some intermediate Dart files (in [tool/colors/generated](https://github.com/flutter/tools_metadata/tree/master/tool/colors/generated))
- [tool/colors/generate_files.dart](https://github.com/flutter/tools_metadata/blob/master/tool/colors/generate_files.dart) references those files and generates JSON/properties files

I think the best way to handle this would be:

- Add css_colors to the pubspec so that the package is available locally (without downloading it from GitHub)
- In `update_colors.dart`, use `Isolate.resolvePackageUri('package:css_colors/css_colors.dart')` to get the path to the source file and use a regex to extract all of the color names to write a file in the `tool/colors/generated` folder similar to the existing Flutter ones
- In `generate_files.dart`, add the new generated file to the `generatePropertiesFiles` and `generateJsonFiles` functions

If you'd like to have a go, I'm happy to answer questions/provide pointers if you get stuck. I'm also happy to do it if that's easier for you. Once that's done, we should update this PR slightly so the JSON is in the same file as the others (and in the same format/order as whatever we commit to tools_metadata for easier updating in future). The main goal is just to have tools_metadata as the source of truth for these mappings.

Semvrij:
> * In `update_colors.dart`, use `Isolate.resolvePackageUri('package:css_colors/css_colors.dart')` to get the path to the source file ...

Thanks! This was exactly what I was getting stuck on. I did not know how to get the path of the file.

DanTup:
@Semvrij thanks for the other PR! Since the output JSON is the same as we have here, we can just move it from `css_colours.ts` into the same `colours.ts` file as the others (that file should ultimately be generated by joining together the JSON files from tools_metadata but it's currently just a copy/paste).

With that done, this all looks good to me, thanks!

Semvrij:
Done! @DanTup

DanTup:
@Semvrij super, thanks!

PR #3451, opened 2021-07-01
Use daemon to start bazel DevTools (WIP)

helin24:
This sets up a running DevTools server using a script specified in the workspace, but trying to get the embedded inspector page fails with "Dart DevTools was unable to launch Chrome so your default browser was launched instead", along with "Request must contain a "method" key." (I don't see a default browser opening either)

I see that this logic goes through `launchInEmbeddedWebView` but I'm not sure what parameters are expected here for a successful launch and not sure what I may be missing.

Also, is there a hook on workspace close to shut down the daemon process?

### DIFF
--- a/src/extension/sdk/dev_tools/manager.ts
+++ b/src/extension/sdk/dev_tools/manager.ts
@@ -8,8 +8,8 @@ import { FlutterCapabilities } from "../../../shared/capabilities/flutter";
 import { vsCodeVersion } from "../../../shared/capabilities/vscode";
 import { CHROME_OS_DEVTOOLS_PORT, devToolsPages, isChromeOS, pubPath, reactivateDevToolsAction, skipAction } from "../../../shared/constants";
 import { LogCategory, VmService } from "../../../shared/enums";
-import { DartWorkspaceContext, DevToolsPage, Logger, SomeError } from "../../../shared/interfaces";
-import { CategoryLogger } from "../../../shared/logging";
+import { DartWorkspaceContext, DevToolsPage, FlutterWorkspaceContext, Logger, SomeError } from "../../../shared/interfaces";
+import { CategoryLogger, EmittingLogger } from "../../../shared/logging";
 import { UnknownNotification } from "../../../shared/services/interfaces";
 import { StdIOService } from "../../../shared/services/stdio_service";
 import { disposeAll } from "../../../shared/utils";
@@ -19,6 +19,7 @@ import { envUtils, isRunningLocally } from "../../../shared/vscode/utils";
 import { Analytics } from "../../analytics";
 import { DebugCommands, debugSessions } from "../../commands/debug";
 import { config } from "../../config";
+import { FlutterDaemon } from "../../flutter/flutter_daemon";
 import { PubGlobal } from "../../pub/global";
 import { getToolEnv } from "../../utils/processes";
 import { DartDebugSessionInformation } from "../../utils/vscode/debug";
@@ -57,8 +58,9 @@ export class DevToolsManager implements vs.Disposable {
 	private async handleEagerActivationAndStartup(workspaceContext: DartWorkspaceContext) {
 		if (workspaceContext.config?.startDevToolsServerEagerly) {
 			try {
-				if (workspaceContext.config?.startDevToolsServerEagerly)
-					await this.spawnIfRequired(true);
+				if (workspaceContext.config?.startDevToolsServerEagerly) {
+					await this.spawnIfRequired(workspaceContext.config?.startDevToolsFromDaemon ? new FlutterDaemon(new EmittingLogger(), workspaceContext as FlutterWorkspaceContext, this.flutterCapabilities) : undefined, true);
+				}
 			} catch (e) {
 				this.logger.error("Failed to background start DevTools");
 				this.logger.error(e);
@@ -84,7 +86,7 @@ export class DevToolsManager implements vs.Disposable {
 		return page.id;
 	}
 
-	private async spawnIfRequired(silent = false): Promise<string | undefined> {
+	private async spawnIfRequired(flutterDaemon?: FlutterDaemon, silent = false): Promise<string | undefined> {
 		// If we're mid-silent-activation, wait until that's finished.
 		await this.devToolsActivationPromise;
 
@@ -102,7 +104,16 @@ export class DevToolsManager implements vs.Disposable {
 				return undefined;
 			}
 			this.capabilities.version = installedVersion;
-			if (silent) {
+			if (flutterDaemon != null) {
+				const result = await flutterDaemon.serveDevTools();
+				this.devtoolsUrl = new Promise<string>((resolve, reject) => {
+					if (result.host != null && result.port != null) {
+						resolve(`http://${result.host}:${result.port}/`);
+					} else {
+						reject("Unable to serve DevTools");
+					}
+				});
+			} else if (silent) {
 				this.devtoolsUrl = this.startServer();
 			} else {
 				this.devtoolsUrl = vs.window.withProgress({
--- a/src/extension/sdk/dev_tools/manager.ts
+++ b/src/extension/sdk/dev_tools/manager.ts
@@ -106,7 +106,7 @@ export class DevToolsManager implements vs.Disposable {
 			if (flutterDaemon != null) {
 				const result = await flutterDaemon.serveDevTools();
 				this.devtoolsUrl = new Promise<string>((resolve, reject) => {
-					if (result.host != null && result.port != null) {
+					if (result.host && result.port) {
 						resolve(`http://${result.host}:${result.port}/`);
 					} else {
 						reject("Unable to serve DevTools");
--- a/src/extension/sdk/dev_tools/manager.ts
+++ b/src/extension/sdk/dev_tools/manager.ts
@@ -8,8 +8,8 @@ import { FlutterCapabilities } from "../../../shared/capabilities/flutter";
 import { vsCodeVersion } from "../../../shared/capabilities/vscode";
 import { CHROME_OS_DEVTOOLS_PORT, devToolsPages, isChromeOS, pubPath, reactivateDevToolsAction, skipAction } from "../../../shared/constants";
 import { LogCategory, VmService } from "../../../shared/enums";
-import { DartWorkspaceContext, DevToolsPage, FlutterWorkspaceContext, Logger, SomeError } from "../../../shared/interfaces";
-import { CategoryLogger, EmittingLogger } from "../../../shared/logging";
+import { DartWorkspaceContext, DevToolsPage, IFlutterDaemon, Logger, SomeError } from "../../../shared/interfaces";
+import { CategoryLogger } from "../../../shared/logging";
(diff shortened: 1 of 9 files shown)


src/extension/sdk/dev_tools/manager.ts
```diff
@@ -57,8 +58,9 @@ export class DevToolsManager implements vs.Disposable {
 	private async handleEagerActivationAndStartup(workspaceContext: DartWorkspaceContext) {
 		if (workspaceContext.config?.startDevToolsServerEagerly) {
 			try {
-				if (workspaceContext.config?.startDevToolsServerEagerly)
-					await this.spawnIfRequired(true);
+				if (workspaceContext.config?.startDevToolsServerEagerly) {
+					await this.spawnIfRequired(workspaceContext.config?.startDevToolsFromDaemon ? new FlutterDaemon(new EmittingLogger(), workspaceContext as FlutterWorkspaceContext, this.flutterCapabilities) : undefined, true);
```
DanTup, line 62: Will this spawn a second daemon process? If so, I think we should instead pass the existing daemon in (constructed in `extension.ts` - the code that creates it might need moving to be earlier in the activation, but I don't think that should cause any problems).

src/extension/sdk/dev_tools/manager.ts
```diff
@@ -102,7 +104,16 @@ export class DevToolsManager implements vs.Disposable {
 				return undefined;
 			}
 			this.capabilities.version = installedVersion;
-			if (silent) {
+			if (flutterDaemon != null) {
+				const result = await flutterDaemon.serveDevTools();
+				this.devtoolsUrl = new Promise<string>((resolve, reject) => {
+					if (result.host != null && result.port != null) {
```
DanTup, line 110: Can `host` and `port` be explicitly `null`, or would they be missing from the JSON? If missing, I these might need to check `undefined` instead (or both?).

I usually to rely on thruthiness for simple things like this to avoid needing to be explicit about the difference - eg `if (result.host && result.port)` (it doesn't work if falsey values like `0` are valid though).

src/shared/flutter/daemon_interfaces.ts
```diff
@@ -78,3 +78,8 @@ export interface ShowMessage {
 export interface SupportedPlatformsResponse {
 	platforms: PlatformType[];
 }
+
+export interface ServeDevToolsResponse {
+	host: string;
+	port: string;
```
DanTup, line 84: Based on the code above, this maybe should include `| null` (or `| undefined`)?

DanTup (commented):
> trying to get the embedded inspector page fails with "Dart DevTools was unable to launch Chrome so your default browser was launched instead"

We normally try to launch DevTools by calling a VM Service extension (so it can locate Chrome and launch that):

https://github.com/flutter/devtools/blob/64cffbed6366329ad05e44d48fa2298367643bb6/packages/devtools_server/lib/src/server.dart#L695

This usually works by:
- We spawn DevTools using the command line with `--machine` (at startup or on first session)
- We send any VM Service URIs to the server and DevTools server registers the `launchDevTools` service extension with them (when a debug session starts)
- We call the VM Service extension when we need to launch an external DevTools window

That's probably not going to be valid here, so the fix is probably to update `canLaunchDevToolsThroughService` to be set to false here (perhaps with a workspace settings flag to disable using the service extension to launch DevTools). I think long-term, that code may all go away anyway (the benefit we get from devtools_server is reduced now we're defaulting to embedded Windows, so it may not be worth maintaining this APIs/complexity in DevTools).

(that said, I think this could should be bypassed entirely if launching in the embedded view).

> along with "Request must contain a "method" key."

I'm not certain what's causing, but if you can the **Dart: Capture Logs** command and then repro, seeing the exact payload being sent might help.

> Also, is there a hook on workspace close to shut down the daemon process?

Many of the classes have a `disposables` collection and a `dispose` method. Anything that has things that need cleaning up will be registered in the extensions "subscriptions" which VS Code will call `dispose()` on when the extension is unloading.

The original daemon spawned in `extension.ts` is added directly to the extension contexts subscriptions and has a `dispose` method (which calls down to `StdIOService`'s `dispose` that terminates the process):

https://github.com/Dart-Code/Dart-Code/blob/master/src/extension/extension.ts#L452-L456

(see comments below - I think we might want to avoid spawning a daemon within the DevTools manager, it may be better to try and pass this one through)

src/extension/sdk/dev_tools/manager.ts
```diff
@@ -102,7 +103,16 @@ export class DevToolsManager implements vs.Disposable {
 				return undefined;
 			}
 			this.capabilities.version = installedVersion;
-			if (silent) {
+			if (flutterDaemon != null) {
```
DanTup, line 106: nit: if we add `private readonly` in front of `flutterDaemon` in the constructor, that would avoid needing to pass it down through the method calls, and we could check `workspaceContext.config?.startDevToolsFromDaemon` directly here which would be a little more explicit than `flutterDaemon` being set in the case where it's being used for DevTools (otherwise, it's worth a comment noting that's the case).

helin24:
The daemon seems to not get disposed properly in the bazel case, and I'm not sure why that is. Is there a spot where there is additional logic for disposal that I should check?

Regarding launching the window, `canLaunchDevToolsThroughService` is `false` and so eventually `launchInEmbeddedWebView` runs. I think this is the path we want? I don't see anything relevant in capture logs when I click on 'Dart DevTools' and get the empty window, e.g.:
```
Fri Jul 02 2021 [10:11:39 GMT-0700 (Pacific Daylight Time)] Log file started
[10:11:43 AM] [General] [Info] Using only available debug session
[10:11:51 AM] [General] [Info] Exposing URL: http://127.0.0.1:56264/Ps35_NBnNIk%3D/
[10:11:51 AM] [General] [Info] Mapping URL: http://127.0.0.1:56264/Ps35_NBnNIk%3D/
[10:11:51 AM] [General] [Info] Mapped URL: http://127.0.0.1:56264/Ps35_NBnNIk%3D/
[10:11:51 AM] [General] [Info] Final URI: http://127.0.0.1:56264/Ps35_NBnNIk%3D/
[10:11:51 AM] [General] [Info] Final URL: http://127.0.0.1:56264/Ps35_NBnNIk%3D/
Fri Jul 02 2021 [10:11:59 GMT-0700 (Pacific Daylight Time)] Log file ended
```

Anything else along this path I should check?

DanTup:
> The daemon seems to not get disposed properly in the bazel case, and I'm not sure why that is. Is there a spot where there is additional logic for disposal that I should check?

I would try enabling the [daemon log](https://dartcode.org/docs/logging/#flutter-daemon) which should log the stdin/stdout for the process so you can see if your shutdown message is being sent (and its response). How are you shutting it down when you expect it to run? If you close VS Code or open another folder it should shut the daemon down, but if you're hitting the Stop button in VS Code it's possible (I'm not certain) that it would just terminate VS Code and might not shut down cleanly.

> Regarding launching the window, `canLaunchDevToolsThroughService` is false and so eventually `launchInEmbeddedWebView` runs. I think this is the path we want? 

Yep, I think so - but in that case I wouldn't have expected "Dart DevTools was unable to launch Chrome so your default browser was launched instead" to be shown - are you still seeing that?

helin24:
Thanks, I looked at the daemon log and I'm not seeing the shutdown message. It seems that the place I put it (https://github.com/Dart-Code/Dart-Code/pull/3451/files#diff-2f46814827e2ed95cdb578341aff004f3e6e2da7f43acd6d8b1686b4db02733cR889) may be too late in the shutdown process. It may be that the parent daemon process has already been shut down, but without an explicit 'shutdown' command earlier on, there's a DevTools process hanging on? Is there a code pointer for where/when the flutter daemon process is stopped?

Regarding the inspector window launch, I think I was seeing an old message. Now I am no longer seeing the "unable to launch Chrome" message, and I can tell checking breakpoints that it runs [this line of code](https://github.com/Dart-Code/Dart-Code/blob/master/src/extension/sdk/dev_tools/embedded_view.ts#L65) with a URI that I can successfully navigate to in a separate Chrome window. Once that finishes I end up with a blank window though:
Screen Shot 2021-07-08 at 3 55 01 PM

DanTup:
> Thanks, I looked at the daemon log and I'm not seeing the shutdown message. It seems that the place I put it (https://github.com/Dart-Code/Dart-Code/pull/3451/files#diff-2f46814827e2ed95cdb578341aff004f3e6e2da7f43acd6d8b1686b4db02733cR889) may be too late in the shutdown process

Ah, you added it after the logs are stopped, so the logging was shut down before it was logged. I swapped the order like this:

Screenshot 2021-07-09 at 16 06 38

And it shows up in the log for me:

```
[4:06:11 PM] [FlutterDaemon] [Info] <== [{"id":"3","result":{"platforms":["linux","macos","windows","ios","android","web"]}}]
[4:06:16 PM] [FlutterDaemon] [Info] ==> [{"id":"4","method":"daemon.shutdown"}]
[4:06:16 PM] [FlutterDaemon] [Info] <== [{"id":"4"}]
Fri Jul 09 2021 [16:06:16 GMT+0100 (British Summer Time)] Log file ended
```

> Regarding the inspector window launch, I think I was seeing an old message. [...]
> Once that finishes I end up with a blank window though:

If you click DevTools in the status bar then "Open in browser" does it load there correctly? If not, it may be easier to understand why there (eg. if we built the wrong URL). If it's fine externally, you can use Help -> Toggle Developer Tools in VS Code to open the Chromium dev tools which might help (eg. Console tab, or examining the iframe to see its src etc).

helin24:
Oh that was silly of me! Yes, now I see the shutdown message in logs and the DevTools server is also properly shut down. Not sure why lack of logging would keep it from executing, but maybe that caused an error? Hopefully this isn't a race condition where the shutdown might sometimes not get executed.

Ah here's the error message from trying to open widget inspector embedded: 
```
Refused to display 'http://127.0.0.1:9100/' in a frame because it set 'X-Frame-Options' to 'sameorigin'.
```

It works when I open in browser. Do you know why that would be different for the external case? I imagine the address should be similar.

DanTup:
> Yes, now I see the shutdown message in logs and the DevTools server is also properly shut down. Not sure why lack of logging would keep it from executing, but maybe that caused an error?

I can't think why - I would've expected it to have worked as you had it, but just not logged correctly. There shouldn't be a race as long as the `await` is working correctly (waiting for the response) and the daemon isn't giving that response until it's really done any shutdown it needs to.

src/shared/flutter/daemon_interfaces.ts
helin24: I think truthiness check makes sense for this case, so in that case I think you mean that we don't need null/undefined here?

DanTup: If the value _could_ be `null` or `undefined`, we should include it here (otherwise it looks like those are not valid values and we may accidentally write code that assumes it's never `null`/`undefined` in the future).

I'm assuming it can be null because of the code above (`result.host != null && result.port != null`), but I'm not certain (this depends on what the internal daemon can emit).

helin24: Oops I misunderstood what code we were looking at. It looks like these fields could both be null.

PR #3474, opened 2021-07-14
 Add a "Dart: Add Dependency" command

### LINKED ISSUE #3306: Add a command to "add package" that alls "pub add"
Pub has both a command for adding new packages, and an API to get all package names for completion. With these, we could add a command palette command to add packages that would avoid having to modify the pubspec directly.

In the past, some issues have been reported that turned out to be subtly-incorrect indentation in pubspec, so handing off the editing to Pub for simple tasks like this might save some of those.

DanTup:
- [x] tests

Possible future improvements:

- [x] quick-fix on missing package URIs? (if we can distinguish from those that are just errors in the filename part)
- [ ] improve error if trying to add Flutter package to a Dart project
- [ ] show (and cache) package descriptions as you move through them
- [x] improve adding Flutter package to non-Flutter project (if Flutter SDK available)

Fixes #3306.

### DIFF
--- a/package.json
+++ b/package.json
@@ -133,6 +133,11 @@
 				"category": "Dart"
 			},
 			{
+				"command": "dart.addDevDependency",
+				"title": "Add Dev Dependency",
+				"category": "Dart"
+			},
+			{
 				"command": "dart.toggleDartdocComment",
 				"title": "Toggle Dartdoc Comment",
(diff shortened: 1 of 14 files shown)


fjmurillovBuller:
hola, esta excelente, estaremos conectados para aprender y emprender

PR #3483, opened 2021-07-20
Add support hexadecimal uppercase x

Semvrij:
This pull request adds support for hexadecimal values with a uppercase letter x in the prefix. For example, `0XFFFFFFFF` instead of a lowercase x `0xFFFFFFFF`.


DanTup:
Thanks!

Issue #2059, opened 2019-10-20
Templates like if, for do not fit to context


vogella:
Looks to me that the if, for and other standard templates are available independent of the content.

For example, I can use them in the import section which IMHO is wrong.

See screenshot.

![image](https://user-images.githubusercontent.com/139910/67158594-54ed4780-f33a-11e9-8cce-4e12c13ca249.png)

DanTup:
This is unfortunately just how snippets work in VS Code - they have no context. The best fix is for the snippets to come from the server (https://github.com/dart-lang/sdk/issues/37831) but for now they'll always be visible.

That said - when I try to reproduce your issue, I get normal completion items for the import, and the snippets are pushed all the way to the bottom:

<img width="451" alt="Screenshot 2019-10-21 at 6 34 38 pm" src="https://user-images.githubusercontent.com/1078012/67228580-7bda7500-f431-11e9-95e5-f41bb1631161.png">

<img width="591" alt="Screenshot 2019-10-21 at 6 34 49 pm" src="https://user-images.githubusercontent.com/1078012/67228584-7ed56580-f431-11e9-8781-66c4f520b5fc.png">

I'm curious why you're not seeing the same. Does your completion otherwise work as expected?

vogella:
Mine is also at the bottom but I scrolled to it.

DanTup:
Oh, I see!

In that case I think this is working as expected, and requires the above SDK work to (reliably) improve it. The main issue is that in the editor we don't really know the context of where the cursor is (nor do our snippets define in what contexts they are valid).

DanTup:
In the upcoming release of Dart-Code, server snippets will be enabled by default if your SDK supports them (which if you're on stable won't be until the next SDK releases, but if you're on master/dev, it probably already does). The server-provided snippets will only show up in contexts where they make sense (so if/for etc. won't show up at top level, and things like `stful` will only show up at top level).

PR #3492, opened 2021-07-25
feat(snippets): create page snippet

### LINKED ISSUE #3493: Create page snippet for navigation API
I've been working with Flutter 2.0 and the navigator API and I think it would be really great if we could have a snippet for creating pages.  This will definitely reduce waste of time by people manually typing out a page widget every time they need to create a new page.

**Describe the solution you'd like**
I would like a snippet that can create a page widget setup.

I have created my own snippets for doing this as shown in my PR.

sno2:
Closes #3493 
I've been working with Flutter 2.0 and the navigator API and I think it would be really great if we could have a snippet for creating pages.  This is one that I've been using globally in my editor and hopefully others may enjoy the convenience as well.

### DIFF
--- a/snippets/flutter.json
+++ b/snippets/flutter.json
@@ -72,6 +72,29 @@
 				"  }",
 				"}"
 			]
+		},
+		"Flutter page widget with stateless view": {
+			"prefix": "pge",
+			"description": "Insert a Page with a StatelessWidget view",
+			"body": [
+				"class $1Page extends Page {",
+				"  const $1Page() : super(key: const ValueKey('$1 Page'));",
+				"",
+				"  Route createRoute(BuildContext context) {",
+				"    return MaterialPageRoute(",
+				"        settings: this, builder: (BuildContext context) => $1View());",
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks for the PR! Unfortunately due to how snippets show up _everywhere_ regardless of context I'm avoiding adding any more here. My goal is to move the existing snippets to the language server instead which could show only the snippets that are relevant based on the location of the cursor (see https://github.com/Dart-Code/Dart-Code/issues/2059).

I also get quite a lot of requests like this, but adding all of those snippets to the base extension would cause them to show up for everyone, and a lot of them are a little subjective to how people write code. Often a better option is to add snippets locally (which I think you've done) or in a separate extension ([for example this one](https://marketplace.visualstudio.com/items?itemName=Nash.awesome-flutter-snippets)) where users can more actively choose to have them rather than automatically getting them when installing the main Dart/Flutter extensions.

I hope this all makes sense, and I appreciate the effort all the same! Thanks!

PR #3505, opened 2021-07-30
Allow bazel target that's not a main file

helin24:
I had wanted to use only main files as targets to minimize divergence from the external workflow, but it would be nice to have an exception for now. It will take more time to investigate how much work it would be to get teams to change their configurations to using main files only.


DanTup (approved):
LGTM

PR #3517, opened 2021-08-05
Add support for DevTools to open external links

tomgilder:
See https://github.com/flutter/devtools/issues/3251 - external links are currently failing when DevTools are embedded in VSCode due to the "allow-popups" sandbox restriction being enabled.

There might be a simpler solution to this that I'm unaware of.

I'm not very experienced with VSCode/TypeScript development, please check carefully 😁

Especially not sure if I got the dispose logic right.

Buddy PR on DevTools: https://github.com/flutter/devtools/pull/3252

### DIFF
--- a/src/extension/sdk/dev_tools/embedded_view.ts
+++ b/src/extension/sdk/dev_tools/embedded_view.ts
@@ -1,10 +1,11 @@
 import * as vs from "vscode";
 import { Event, EventEmitter } from "../../../shared/events";
 import { DevToolsPage } from "../../../shared/interfaces";
-import { firstNonEditorColumn } from "../../../shared/vscode/utils";
+import { envUtils, firstNonEditorColumn } from "../../../shared/vscode/utils";
 import { DartDebugSessionInformation } from "../../utils/vscode/debug";
 
 const pageScript = `
+const vscode = acquireVsCodeApi();
 window.addEventListener('message', (event) => {
 	const message = event.data;
 	const devToolsFrame = document.getElementById('devToolsFrame');
@@ -25,6 +26,9 @@ window.addEventListener('message', (event) => {
 			// https://github.com/flutter/devtools/issues/2775
 			window.dispatchEvent(new KeyboardEvent('keydown', message.data));
 			break;
+		case "launchUrl":
+			vscode.postMessage({command: 'launchUrl', data: message.data});
+			break;
 	}
 });
 `;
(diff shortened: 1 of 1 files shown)


src/extension/sdk/dev_tools/embedded_view.ts
```diff
@@ -25,6 +26,9 @@ window.addEventListener('message', (event) => {
 			// https://github.com/flutter/devtools/issues/2775
 			window.dispatchEvent(new KeyboardEvent('keydown', message.data));
 			break;
+		case "launchUrl":
+			vscode.postMessage({command: 'launchUrl', data: message.data});
+			break;
```
DanTup, line 31: I wonder if we could call the post the same `did-click-link` message that VS Code seems to (to avoid having our own)?

https://github.com/microsoft/vscode/search?q=did-click-link

DanTup (commented):
I added a note on https://github.com/flutter/devtools/pull/3252#pullrequestreview-723367816. If we can't come up with something better and DevTools is happy to have that change, I think this looks reasonable. It would be good to avoid leaking too much VS Code stuff into DevTools if we can help it though.

src/extension/sdk/dev_tools/embedded_view.ts
tomgilder: Doesn't look like it to me; it would require calling `hostMessaging.postMessage('did-click-link', url)` but the [VSCode message method](https://github.com/microsoft/vscode/blob/99c8e788d46c935aecb16129919c6041eed92250/src/vs/workbench/contrib/webview/browser/pre/main.js#L947) requires the posted command to be "onmessage" or "do-update-state".

src/extension/sdk/dev_tools/embedded_view.ts
```diff
@@ -25,6 +26,9 @@ window.addEventListener('message', (event) => {
 			// https://github.com/flutter/devtools/issues/2775
 			window.dispatchEvent(new KeyboardEvent('keydown', message.data));
 			break;
+		case "launchUrl":
+			vscode.postMessage({command: 'launchUrl', data: message.data});
```
DanTup, line 30: I think we should change this tso the URL is included inside ` data` object instead of directly. That way if we need to ever pass additional data with the URL in future (for example whether it should open in another embedded window vs external, or something), it would be non-breaking to change it (and not require coordination with a DevTools change).

DanTup (changes requested):
I think there's one change we should make in case we need to extend in future, but otherwise looks good to me - thanks!

src/extension/sdk/dev_tools/embedded_view.ts
tomgilder: Done

DanTup:
Thanks! I'm OOO this week, but will check the dispose is all good next week and merge this for the next version :-)

DanTup (approved):

DanTup:
This all looks good to me, and seems to work fine. It does trigger an error because of the change in DevTools to also call allow the original event to be handled, but I think I agree with Jacob that it's better to always call that. It's unlikely the sandbox rules will change, so it should never trigger to windows being opened, and the error is explainable.

Thanks!

PR #3589, opened 2021-10-04
style: add return type of main

Ascenio:
This could be subjective, but I think it's good to have a return type on every function. Every time I use the `main` snippet I find myself writing void anyway. You could also argue that I could make my own snippet, and that's true. But because dart is typed after all and `fun` uses void, why not?

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -4,7 +4,7 @@
 			"prefix": "main",
 			"description": "Insert a main function, used as an entry point.",
 			"body": [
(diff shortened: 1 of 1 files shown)


Ascenio:
I think there's something funny going on with tests.

DanTup:
Thanks!

I don't believe either of the test failures are related to this. One is a known problem and the other looks like the SDK update check got a status 200 but not valid JSON and didn't handle that correctly. I'll improve the handling of that on the main branch.

Issue #3620, opened 2021-10-18
Subscribe to analysis.navigation for visible editors and use the results for instant navigation


JoeyScarr:
In my company's codebase, code navigation is intolerably slow, to the point where most devs just use global find instead of the jump-to-def/find-all-references functionality.

Is there some reason why Dart-Code doesn't subscribe to `analysis.navigation` events from the language server? FWIW Dart for Webstorm does this and uses the results for code navigation, making it instant.

I have a hacked up version of Dart-Code locally that does this and it vastly improves the experience for me. Happy to send a PR if anyone's interested (though it's not very hard to implement).

JoeyScarr:
Thanks for the quick response!

> Also - can you clarify if both of these are slow, or just find-refs? My understanding is that the navigation event only provided the information required for Go-to-Definition, so I wouldn't expect to be able to improve Find-References with it - is that incorrect?

Ah, you're right. It's only Go to Definition that's improved by this.

> By default, Dart-Code now uses LSP and not the original server protocol (if you set dart.previewLsp: false it will use the original protocol, but that will be removed in the future). This means the navigation needs to go via the server.

To provide some context, we're still stuck on Dart 1 for the short-medium term, hence we can't use the LSP yet.

I'll have to get back to you once I figure out how to provide a repro+trace that's both useful and shareable, but another piece of information that might be useful is we have a gencode library that has ~900 parts. It used to be one big file and we recently split it up -- performance was bad before but now it's even worse. It seems like the analysis server is constantly re-analyzing, even though nothing in the giant library is changing.

It's possible that the situation might improve if we split up the gencode into individual pieces that must be imported separately, and/or if we switch to Dart 2 and use newer versions of the analysis server, but both of those are a significant chunk of work. If you have any other ideas for short-term improvement, let me know.

FWIW, my suggested approach in the OP helps mitigate the situation somewhat, but it's still pretty bad as soon as one of the gencode files is opened; it seems that triggers the analyzer to keep analyzing forever.

JoeyScarr:
Sent a PR: https://github.com/Dart-Code/Dart-Code/pull/3649

Not sure if it's the cleanest solution but will leave up to your judgement whether it's mergeable or not.

DanTup:
Fixed by #3649. Thanks!

PR #3649, opened 2021-11-03
Subscribe to navigation events on open files

JoeyScarr:
Use that to navigate if available, otherwise fall back to requesting navigation directly from the analysis server.

### DIFF
--- a/src/extension/providers/dart_reference_provider.ts
+++ b/src/extension/providers/dart_reference_provider.ts
@@ -3,9 +3,10 @@ import { flatMap } from "../../shared/utils";
 import { fsPath } from "../../shared/utils/fs";
 import { toRange, toRangeOnLine } from "../../shared/vscode/utils";
 import { DasAnalyzerClient } from "../analysis/analyzer_das";
+import { DasFileTracker } from "../analysis/file_tracker_das";
 
 export class DartReferenceProvider implements ReferenceProvider, DefinitionProvider {
-	constructor(private readonly analyzer: DasAnalyzerClient) { }
+	constructor(private readonly analyzer: DasAnalyzerClient, private readonly fileTracker: DasFileTracker) { }
 
 	public async provideReferences(document: TextDocument, position: Position, context: ReferenceContext, token: CancellationToken): Promise<Location[] | undefined> {
 		// If we want to include the decleration, kick off a request for that.
@@ -33,11 +34,16 @@ export class DartReferenceProvider implements ReferenceProvider, DefinitionProvi
 	}
 
 	public async provideDefinition(document: TextDocument, position: Position, token: CancellationToken): Promise<DefinitionLink[] | undefined> {
-		const resp = await this.analyzer.analysisGetNavigation({
-			file: fsPath(document.uri),
-			length: 0,
-			offset: document.offsetAt(position),
-		});
+		let resp1 = this.fileTracker.getNavigationTargets(fsPath(document.uri), document.offsetAt(position));
+		if (!resp1) {
+			resp1 = await this.analyzer.analysisGetNavigation({
+				file: fsPath(document.uri),
+				length: 0,
+				offset: document.offsetAt(position),
+			});
+		}
+		if (!resp1) return undefined;
+		const resp = resp1;
 
 		if (token && token.isCancellationRequested)
 			return;
--- a/src/extension/analysis/file_tracker_das.ts
+++ b/src/extension/analysis/file_tracker_das.ts
@@ -1,5 +1,5 @@
 import { CancellationToken, Disposable, TextDocument, Uri, window, workspace } from "vscode";
-import { FlutterOutline, FoldingRegion, Occurrences, Outline } from "../../shared/analysis_server_types";
+import { AnalysisGetNavigationResponse, AnalysisNavigationNotification, FilePath, FlutterOutline, FoldingRegion, NavigationRegion, Occurrences, Outline } from "../../shared/analysis_server_types";
 import { IAmDisposable, Logger } from "../../shared/interfaces";
 import { disposeAll } from "../../shared/utils";
 import { fsPath } from "../../shared/utils/fs";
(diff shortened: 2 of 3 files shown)


JoeyScarr:
This is the change I mentioned in #3620.

src/extension/providers/dart_reference_provider.ts
```diff
@@ -33,11 +34,16 @@ export class DartReferenceProvider implements ReferenceProvider, DefinitionProvi
 	}
 
 	public async provideDefinition(document: TextDocument, position: Position, token: CancellationToken): Promise<DefinitionLink[] | undefined> {
-		const resp = await this.analyzer.analysisGetNavigation({
-			file: fsPath(document.uri),
-			length: 0,
-			offset: document.offsetAt(position),
-		});
+		let resp1 = this.fileTracker.getNavigationTargets(fsPath(document.uri), document.offsetAt(position));
```
DanTup, line 37: nit: Is there any reason to name this `resp1` and not just `resp` and remove the assignment on line 46?

src/extension/providers/dart_reference_provider.ts
```diff
@@ -33,11 +34,16 @@ export class DartReferenceProvider implements ReferenceProvider, DefinitionProvi
...
+		if (!resp1) {
+			resp1 = await this.analyzer.analysisGetNavigation({
+				file: fsPath(document.uri),
+				length: 0,
+				offset: document.offsetAt(position),
+			});
+		}
+		if (!resp1) return undefined;
```
DanTup, line 45: Is it possible for this to be undefined/null at this point? The original code seems to assume `analysisGetNavigation` can't return undefined, so I think it should still be the case after the above has run?

DanTup (commented):
I think this looks fine (with two minor nits/comments), although I think it may introduce a race that could be hit if the server is busy that might result in inconsistent results:

- Notification is sent from server, client caches it
- Server gets busy so it's not responding to requests
- Modify a file (notification is sent to server)
- Try to navigate - it will use the stale results from the previous notification (since the server hasn't sent the new one yet)

With the previous code, the request would go to the server and be queued/async, so should return the latest results.

I'm not sure this is an issue worth worrying about though (both because it seems unlikely, and because almost all users are using LSP), so I don't think it blocks landing this.

I'll wait for your response to the comments below, but otherwise I think we can merge this. Thanks for the contribution!

src/extension/providers/dart_reference_provider.ts
JoeyScarr: Probably not. I just did it to satisfy the type system; if you remove this it seems to want non-null assertions on the usages of `resp` below. That's probably better.

src/extension/providers/dart_reference_provider.ts
JoeyScarr: Same as above, was just to satisfy the type system. Feel free to remove, you'll just have to add a couple of `1`s in the code below.

JoeyScarr:
> I think this looks fine (with two minor nits/comments), although I think it may introduce a race that could be hit if the server is busy that might result in inconsistent results:
> 
> * Notification is sent from server, client caches it
> * Server gets busy so it's not responding to requests
> * Modify a file (notification is sent to server)
> * Try to navigate - it will use the stale results from the previous notification (since the server hasn't sent the new one yet)
> 
> With the previous code, the request would go to the server and be queued/async, so should return the latest results.
> 
> I'm not sure this is an issue worth worrying about though (both because it seems unlikely, and because almost all users are using LSP), so I don't think it blocks landing this.
> 
> I'll wait for your response to the comments below, but otherwise I think we can merge this. Thanks for the contribution!

There is another issue I think, which is that if the DAS is slow to send back the navigations for a file after it's modified, the stored navigation offsets will be wrong. This could be solved by tracking offset changes from edits and taking them into account when finding the right navigation link to use, but will leave up to you to decide whether you care.

DanTup:
> There is another issue I think, which is that if the DAS is slow to send back the navigations for a file after it's modified, the stored navigation offsets will be wrong. This could be solved by tracking offset changes from edits and taking them into account when finding the right navigation link to use,

Yeah, that's the same issue really. Another option could be to just blow away the cached data whenever a file is modified, so in that case it'll hit the server. It could be done with a listener on `workspace.onDidChangeTextDocument` in the constructor, similar to the existing open/close calls.

I'm happy to make that change after merging this, or you're welcome to add (or, if you think that will introduce problems, let me know).

src/extension/providers/dart_reference_provider.ts
DanTup: I'd prefer not to have the `!` so this is fine - I just thought TypeScript was a little cleverer and wouldn't have reported it 🙄

JoeyScarr:
> > There is another issue I think, which is that if the DAS is slow to send back the navigations for a file after it's modified, the stored navigation offsets will be wrong. This could be solved by tracking offset changes from edits and taking them into account when finding the right navigation link to use,
> 
> Yeah, that's the same issue really. Another option could be to just blow away the cached data whenever a file is modified, so in that case it'll hit the server. It could be done with a listener on `workspace.onDidChangeTextDocument` in the constructor, similar to the existing open/close calls.
> 
> I'm happy to make that change after merging this, or you're welcome to add (or, if you think that will introduce problems, let me know).

Sounds fine, feel free to add that. I've been using the plugin with just this patch and the issue doesn't seem to crop up too often, so I'm not too fussed either way. Thanks for the review!

DanTup:
I decided not to bother with the change above, as I don't think it would be 100% reliable anyway. Eg:

- you modify document (1)
- we clear the local cache
- you modify document (2)
- we clear the local cache
- server sends notification from edit (1)
- you Go-to-Definition, and use the stale results

So since we can't really guard against it reliably, I'm not sure it's worth trying to cover a subset of those cases. The chance of it occurring seems fairly slim anyway (and ofc this code will ultimately go away).

Thanks for the contribution!

PR #3650, opened 2021-11-03
Add icons to device labels

### LINKED ISSUE #3647: Show icons against devices based on their type
With https://github.com/microsoft/vscode/issues/75000 fixed, we might be able to add icons to distinguish the difference devices types more quickly:

Screenshot 2019-06-06 at 4 55 59 pm

It'll require the next VS Code update (due this/next week?).

guidezpl:
Screen Shot 2021-11-03 at 18 26 34

Screen Shot 2021-11-03 at 22 55 25

Closes #3647


guidezpl:
@DanTup I don't have permission to request a review, please review :)

DanTup (approved):
LGTM, thanks! :-)

DanTup:
Oh, maybe we should add a mobile icon for "Create Android Emulator"?

guidezpl:
It might be more clear to only show icons for available devices and keep the commands as is

Before open
Screen Shot 2021-11-04 at 14 12 24
After open
Screen Shot 2021-11-04 at 14 13 22

DanTup:
Yeah, I think you're right - although the alignment looks a bit weird. Is there any blank/spacer we could use so the text all lines up? If not, then I'll merge as-is. Thanks!

guidezpl:
There isn't, but just in case, here's what it looks like with icons:
Screen Shot 2021-11-05 at 03 57 57. WDYT?

DanTup:
I think that looks ok, but I wouldn't say miles better so I'll leave it to you.

It might be better if we could have a separator like shown here:

image

But I can't find the code doing that - I'm not sure if it's VS Code doing it from some special API for notebooks, or if we can do this ourselves 🤔

guidezpl:
It's coming next release apparently! I filed https://github.com/Dart-Code/Dart-Code/issues/3655.  I'll go with icons

DanTup:
Great!

There's a failing test on the PR:

```
  2) device_manager
       overrides real emulators with custom definitions:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ '$(play) Start My emulator override'
- 'Start My emulator override'
      + expected - actual

      -$(play) Start My emulator override
      +Start My emulator override
```

To run, select Flutter or Flutter LSP tests in the debug pane (you can also change the test to `it.only(` to run just that test) - although the fix is probably fairly clear :-)

Thanks!

DanTup:
Thanks!

PR #3871, opened 2022-03-12
Fix configuration groups ordering

guidezpl:
Give every group an order (same-order = alphabetical) 

Before | After
--- | --- 
 
Screen Shot 2022-03-12 at 11 36 33 | Screen Shot 2022-03-12 at 11 36 20


DanTup:
Nice, thanks!

PR #3965, opened 2022-05-17
Call `super.dispose()` at the end of `dispose()` in `stanim` snippet

nilsreichardt:
From the Flutter docs of the [dispose](https://api.flutter.dev/flutter/widgets/State/dispose.html) method:
> Implementations of this method should end with a call to the inherited method, as in super.dispose().

Therefore, we should call `super.dispose()` at the end.


DanTup:
Thanks!

With newer versions of the extension/SDKs, the snippets come from the server so I've applied the same fix there at https://dart-review.googlesource.com/c/sdk/+/245003/.

nilsreichardt:
Great, thanks!

Issue #3899, opened 2022-04-01
Use super parameters in snippets


guidezpl:
**Is your feature request related to a problem? Please describe.**
Feature: https://github.com/dart-lang/language/issues/1855, Flutter repo work: https://github.com/flutter/flutter/issues/100575

**Describe the solution you'd like**
Tweak the snippets to adopt this feature

DanTup:
@bwilkerson for code generated by the server (for ex. snippets), should super parameters be used whenever supported, or only if the lint is also enabled?

bwilkerson:
I think it's reasonable for snippets to make use of any valid language feature that makes the generated code better.

I think that it's reasonable to not produce some style of code when there's a lint that prohibits it (that is, to allow lints to reduce the set of valid language features), but I don't think we want to require users to "enable" new language features before we start producing them in generated code. Or, at least, we don't want to require them to enable it again; they already had to opt in once by increasing the minimum SDK version.

goderbauer:
+1 I think if the package has a lower SDK bound of at least 2.17.0-0 (indicating that super params are supported) the macros should use them.

DanTup:
Working on a change at https://dart-review.googlesource.com/c/sdk/+/240281/.

It will only work for server-provided snippets, which are currently something [you can turn on](https://dartcode.org/releases/v3-38/#experimental) if on Flutter `master` using the `dart.enableServerSnippets` setting but will automatically be enabled when your SDK supports them in an upcoming release of the VS Code extension.

DanTup:
Landed in https://github.com/dart-lang/sdk/commit/fa024be728cc0f7c5053d1ceb79f72cbd9746dbb.

eseidelGoogle:
Re: snippets, do https://github.com/Dart-Code/Dart-Code/blob/master/snippets/flutter.json also need to change?

DanTup:
@eseidelGoogle those snippets shipped in the extension cannot adopt to language versions, so updating them will result in errors for anyone on the current stable SDKs (which don't support super params). Those snippets are being phased out - newer Dart/Flutter SDKs (not the current stable version, but future versions) will get their snippets from the server, which can adopt to language versions (and lints).

[...]

PR #3999, opened 2022-06-01
Migrate to super-initializer parameters

CoderUni:
Dart 2.17 now supports [super-initializer parameters](https://dart.dev/guides/language/language-tour#super-parameters). 
```dart
  const HomePage({
    Key? key,
  }) : super(key: key);
```
becomes
```dart
  const HomePage({super.key});
```
This will help reduce boilerplate.

### DIFF
--- a/snippets/flutter.json
+++ b/snippets/flutter.json
@@ -5,7 +5,7 @@
 			"description": "Insert a StatelessWidget",
 			"body": [
 				"class $1 extends StatelessWidget {",
-				"  const $1({ Key? key }) : super(key: key);",
+				"  const $1({super.key});",
 				"",
 				"  @override",
 				"  Widget build(BuildContext context) {",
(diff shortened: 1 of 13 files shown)


DanTup:
Thanks for the PR! However it's deliberate that we haven't migrated the extension-shipped snippets to use super-params because we want to be compatible with older SDK versions for users that haven't upgraded.

Users that are on the latest versions will now get snippets from the language server (which can adapt them based on language version) - which have been updated to use super params, but that change has not shipped in a stable Flutter SDK yet (see https://github.com/Dart-Code/Dart-Code/issues/3899 / https://github.com/dart-lang/sdk/commit/fa024be728cc0f7c5053d1ceb79f72cbd9746dbb).

The other changes are in test files which would be fine today, although something on my list is to make the CI bots run with the oldest version of the SDKs we support, so I think it may also be best not to migrate them to super params while we're supporting older versions.

Hope this make sense!

PR #4051, opened 2022-07-07
Return to using devtools script for bazel

helin24:
We no longer need to start DevTools from daemon, so that code can be deleted. Instead we can use a custom script in the bazel context.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2741,5 +2741,11 @@
 	"optionalDependencies": {
 		"bufferutil": "^4.0.6",
 		"utf-8-validate": "^5.0.9"
+	},
+	"__metadata": {
+		"id": "f57f68ea-9ee8-42b5-9a97-041d3e4278c4",
+		"publisherDisplayName": "Dart Code",
+		"publisherId": "a606ecbb-a8fb-4c35-b29a-5e8842d82af2",
+		"isPreReleaseVersion": false
 	}
 }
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -2741,5 +2741,11 @@
 	"optionalDependencies": {
 		"bufferutil": "^4.0.6",
 		"utf-8-validate": "^5.0.9"
+	},
+	"__metadata": {
```
DanTup, line 2745: Do you know where this came from? I haven't seen it before (although I'm on an older version of vsce because I had issues with a previous upgrade, so perhaps a newer version is doing this?)

src/test/test_projects/hello_world/pubspec.yaml
```diff
@@ -16,3 +16,7 @@ dependencies:
 
 dev_dependencies:
   test: '>=0.12.34'
+ # test
```
DanTup, line 19: This is left over from some tests that didn't clean up (perhaps they didn't complete correctly) and could be reverted.

DanTup (approved):
LGTM w/ some minor comments

DanTup:
Test failures are unrelated. One is a known issue with Flutter SDK DAPs I haven't gotten to the bottom of, and the other two are from recent changes and should now be fixed on master.

PR #4055, opened 2022-07-12
Disable launch config warning message if flag enabled

helin24:
This disables the launch config warning, but I'm not sure if this is the flag you were thinking of checking. It's used to structure the run command without `--target`, which seems to indicate that the target isn't the usual kind.


DanTup:
Yep, that was the flag I was thinking off. It's perhaps not as a perfect match as I was thinking, but unless there's a need for one without the other it's probably not worth adding an additional flag.

DanTup:
The flutter_sdk_dap failure is known and unrelated, I haven't yet gotten back to tracking down what's causing the failure.

DanTup (approved):

Issue #4057, opened 2022-07-13
Test Dart SDK searching


helin24:
It's currently difficult to test whether SDK search (see https://github.com/Dart-Code/Dart-Code/blob/194eabf7409eb4d606d54afb97eda1dc43f8cb76/src/extension/sdk/utils.ts) returns the expected SDK. We may want to expose `scanWorkspace` and test this function individually.

PR #4058, opened 2022-07-13
Use a default Dart SDK for bazel projects

helin24:
We want to avoid users needing to set their own SDK for bazel workspaces. Most users should be okay with the default SDK from our config file.

I was hoping to test that this default SDK doesn't override a user-set SDK, but there isn't currently an easy way to do this now. https://github.com/Dart-Code/Dart-Code/issues/4057

### DIFF
--- a/src/extension/sdk/utils.ts
+++ b/src/extension/sdk/utils.ts
@@ -304,13 +304,13 @@ export class SdkUtils {
 			fuchsiaRoot && path.join(fuchsiaRoot, "dart/tools/sdks", dartPlatformName, "dart-sdk"),
 			firstFlutterMobileProject && flutterSdkPath && path.join(flutterSdkPath, "bin/cache/dart-sdk"),
 			config.sdkPath,
-			workspaceConfig.defaultDartSdk,
 		].concat(paths)
 			// The above array only has the Flutter SDK	in the search path if we KNOW it's a flutter
 			// project, however this doesn't cover the activating-to-run-flutter.createProject so
 			// we need to always look in the flutter SDK, but only AFTER the users PATH so that
 			// we don't prioritise it over any real Dart versions.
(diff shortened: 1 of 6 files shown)


src/extension/sdk/utils.ts
```diff
@@ -304,6 +304,7 @@ export class SdkUtils {
 			fuchsiaRoot && path.join(fuchsiaRoot, "dart/tools/sdks", dartPlatformName, "dart-sdk"),
 			firstFlutterMobileProject && flutterSdkPath && path.join(flutterSdkPath, "bin/cache/dart-sdk"),
 			config.sdkPath,
+			workspaceConfig.defaultDartSdk,
```
DanTup, line 307: I presume it's ok for this to take priority over anything on `PATH`? This means if a user wants to override this path they'll need to explicitly set `config.sdkPath` and can't just rely on their SDK being picked up from the `PATH` env variable).

src/shared/utils/workspace.ts
```diff
@@ -42,6 +42,7 @@ export function tryProcessBazelFlutterConfig(logger: Logger, config: WritableWor
 			sdkHome: string | undefined;
```
DanTup, line 42: Looks like this `sdkHome` field is used to set `flutterSdkPath`. Should we rename it to be clearer now there's also a Dart one?

DanTup (approved):
LGTM, added some minor questions/comments

src/shared/utils/workspace.ts
helin24: I think we'll leave this here since this is just the JSON parsing spot (workspace config calls this `flutterSdkHome`) and it's annoying to change the format of the file internally.

src/extension/sdk/utils.ts
helin24: True, it'd be better to make this last.

DanTup (approved):

PR #4068, opened 2022-07-22
Use workspace tools script for debug

helin24:
This adds a generic tools script that we can use to call `debug_adapter`. I'm not sure if the tests pass because I was having trouble running the bazel tests locally, but I'll look at them a bit more to see if I can fix that.

### DIFF
--- a/src/extension/providers/debug_adapter_descriptor_factory.ts
+++ b/src/extension/providers/debug_adapter_descriptor_factory.ts
@@ -10,7 +10,7 @@ import { Context } from "../../shared/vscode/workspace";
 import { config } from "../config";
 
 export class DartDebugAdapterDescriptorFactory implements DebugAdapterDescriptorFactory {
-	constructor(private readonly sdks: DartSdks, private readonly logger: Logger, private readonly extensionContext: Context, private readonly dartCapabilities: DartCapabilities, private readonly flutterCapabilities: FlutterCapabilities, private readonly flutterToolsScript?: CustomScript) { }
+	constructor(private readonly sdks: DartSdks, private readonly logger: Logger, private readonly extensionContext: Context, private readonly dartCapabilities: DartCapabilities, private readonly flutterCapabilities: FlutterCapabilities, private readonly flutterToolsScript?: CustomScript, private readonly cwd?: string) { }
 
 	public createDebugAdapterDescriptor(session: DebugSession, executable: DebugAdapterExecutable | undefined): DebugAdapterDescriptor {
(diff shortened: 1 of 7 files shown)


src/extension/providers/debug_adapter_descriptor_factory.ts
```diff
@@ -53,7 +53,7 @@ export class DartDebugAdapterDescriptorFactory implements DebugAdapterDescriptor
 				args.push("--no-dds");
 
 			this.logger.info(`Running SDK DAP Dart VM: ${executable} ${args.join("    ")}`);
-			return new DebugAdapterExecutable(executable, args);
+			return new DebugAdapterExecutable(executable, args, {cwd: "/google/src/cloud/helinx/head/google3"});
```
DanTup, line 56: I think this needs changing :-)

src/shared/interfaces.ts
```diff
@@ -55,6 +55,7 @@ export interface WritableWorkspaceConfig {
 	flutterSdkHome?: string;
 	flutterSyncScript?: string;
 	flutterTestScript?: CustomScript;
+	flutterToolsScript?: CustomScript;
```
DanTup, line 58: Is this intended to be used for all Flutter invocations (for example when we run "flutter pub get" or "flutter doctor")? If not, is it better named as `flutterDebugAdapterScript` or similar? (and if so, do we need to do that now?)

DanTup (changes requested):
I don't know if it's the cause of the failures, but there's a hard-coded `cwd` in here. If that doesn't fix the tests I'll take a proper look then.

(the Flutter SDK DAP bot is still broken, I've not had much look getting to the bottom of that yet, but if the failure is the same as on master we don't need to worry about that one for merging this).

src/extension/providers/debug_adapter_descriptor_factory.ts
helin24: oops! thanks

src/shared/interfaces.ts
helin24: This is a little confusing. There are some flutter invocations that can share this, but not all of them. I can change over the ones that can use it in a later change, but this script isn't specific to debugging.

DanTup (approved):
LGTM, thanks!

src/shared/interfaces.ts
DanTup: SGTM. It may be worth a comment about which it can/should be used for so we have it written here somewhere in this codebase (for my benefit - or others that might not have internal knowledge).

helin24: Done, thanks!

PR #4073, opened 2022-07-26
Use remote daemon for showing devices

helin24:
This allows users to set in config a port where the daemon can be accessed so that devices from another machine can be exposed. This assumes the user has started daemon on the other machine and port forwarding has been done.

~I'm not sure how to access a config from `flutter_debug_impl` - it looks like I shouldn't import vscode APIs there because it's not in the extension folder; should I pass that along elsewhere?~

### DIFF
--- a/package.json
+++ b/package.json
@@ -2359,8 +2359,8 @@
 					"dart.daemonPort": {
 						"type": "number",
 						"default": false,
-						"markdownDescription": "EXPERIMENTAL: Daemon port",
-						"scope": "window"
+						"markdownDescription": "EXPERIMENTAL: The port where flutter daemon can be accessed if daemon is run remotely. This setting is intended for use by Google developers.",
+						"scope": "resource"
 					},
 					"dart.previewCommitCharacters": {
 						"type": "boolean",
--- a/package.json
+++ b/package.json
@@ -2356,6 +2356,12 @@
 						"markdownDescription": "EXPERIMENTAL: Whether to look for custom script definitions at `dart/config/ide/flutter.json` in Bazel workspaces. Currently supported for macOS and Linux only.",
 						"scope": "window"
 					},
+					"dart.daemonPort": {
+						"type": "number",
+						"default": false,
+						"markdownDescription": "EXPERIMENTAL: Daemon port",
+						"scope": "window"
+					},
 					"dart.previewCommitCharacters": {
 						"type": "boolean",
 						"default": false,
--- a/src/debug/flutter_debug_impl.ts
+++ b/src/debug/flutter_debug_impl.ts
@@ -1,7 +1,6 @@
 import { ContinuedEvent, Event, OutputEvent } from "@vscode/debugadapter";
 import { DebugProtocol } from "@vscode/debugprotocol";
 import * as path from "path";
-import { workspace } from "vscode";
 import { FlutterCapabilities } from "../shared/capabilities/flutter";
 import { debugLaunchProgressId, flutterPath, restartReasonManual } from "../shared/constants";
 import { DartLaunchArgs } from "../shared/debug/interfaces";
@@ -251,9 +250,8 @@ export class FlutterDebugSession extends DartDebugSession {
 		if (args.args)
 			allArgs = allArgs.concat(args.args);
 
-		const daemonPort = workspace.getConfiguration("dart").get("daemonPort", null);
-		if (!!daemonPort) {
-			allArgs = allArgs.concat(`--daemon-connection-port=${daemonPort}`);
+		if (!!args.daemonPort) {
+			allArgs = allArgs.concat(`--daemon-connection-port=${args.daemonPort}`);
 		}
 
 		execution = {
--- a/src/debug/flutter_debug_impl.ts
+++ b/src/debug/flutter_debug_impl.ts
@@ -1,6 +1,7 @@
 import { ContinuedEvent, Event, OutputEvent } from "@vscode/debugadapter";
 import { DebugProtocol } from "@vscode/debugprotocol";
 import * as path from "path";
+import { workspace } from "vscode";
 import { FlutterCapabilities } from "../shared/capabilities/flutter";
 import { debugLaunchProgressId, flutterPath, restartReasonManual } from "../shared/constants";
 import { DartLaunchArgs } from "../shared/debug/interfaces";
@@ -249,6 +250,11 @@ export class FlutterDebugSession extends DartDebugSession {
 
 		if (args.args)
 			allArgs = allArgs.concat(args.args);
+
+		const daemonPort = workspace.getConfiguration("dart").get("daemonPort", null);
+		if (!!daemonPort) {
+			allArgs = allArgs.concat(`--daemon-connection-port=${daemonPort}`);
+		}
 
 		execution = {
 			args: allArgs,
--- a/src/debug/flutter_debug_impl.ts
+++ b/src/debug/flutter_debug_impl.ts
@@ -250,10 +250,6 @@ export class FlutterDebugSession extends DartDebugSession {
 		if (args.args)
 			allArgs = allArgs.concat(args.args);
 
-		if (!!args.daemonPort) {
-			allArgs = allArgs.concat(`--daemon-connection-port=${args.daemonPort}`);
-		}
-
 		execution = {
 			args: allArgs,
 			executable: execution.executable,
--- a/src/extension/config.ts
+++ b/src/extension/config.ts
@@ -61,7 +61,7 @@ class Config {
 	get customDevToolsUri(): undefined | string { return this.getConfig<null | string>("customDevToolsUri", null); }
 	get dapLogFile(): undefined | string { return createFolderForFile(resolvePaths(this.getConfig<null | string>("dapLogFile", null))); }
 	get dartTestLogFile(): undefined | string { return createFolderForFile(resolvePaths(this.getConfig<null | string>("dartTestLogFile", null))); }
-	get daemonPort(): undefined | string { return this.getConfig<null | string>("daemonPort", null); }
+	get daemonPort(): undefined | number { return this.getConfig<null | number>("daemonPort", null); }
 	get debugExtensionBackendProtocol(): "sse" | "ws" { return this.getConfig<"sse" | "ws">("debugExtensionBackendProtocol", "ws"); }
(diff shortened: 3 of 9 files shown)


src/debug/flutter_debug_impl.ts
```diff
@@ -250,6 +250,10 @@ export class FlutterDebugSession extends DartDebugSession {
 		if (args.args)
 			allArgs = allArgs.concat(args.args);
 
+		if (!!args.daemonPort) {
+			allArgs = allArgs.concat(`--daemon-connection-port=${args.daemonPort}`);
+		}
```
DanTup, line 255: This code here is the legacy DAP, so this change will be lost in the mode to the SDK DAPs.

While we could also duplicate it in the SDK, I think a better way would be to insert it into `toolArgs` in the VS Code extension, which both DAPs accept as a generic way to pass additional flags without specific support in the debug adapter.

See `buildFlutterToolArgs` here:

https://github.com/Dart-Code/Dart-Code/blob/3047a27cbfde9844d1a1bbe20a40780af7e37c40/src/extension/providers/debug_config_provider.ts#L626

This is the code that adds things like `--web-renderer` based on settings. There is also a `buildFlutterTestToolArgs` if you need to include it for test.

(if this flag only works for the internal Flutter script, we should probably also ensure we only inject it if that script is set, so if an internal user opens a standalone project we don't send it to the public daemon and fail).

src/extension/config.ts
```diff
@@ -60,6 +60,7 @@ class Config {
 	get customFlutterDapPath(): undefined | string { return resolvePaths(this.getConfig<null | string>("customFlutterDapPath", null)); }
 	get dapLogFile(): undefined | string { return createFolderForFile(resolvePaths(this.getConfig<null | string>("dapLogFile", null))); }
 	get dartTestLogFile(): undefined | string { return createFolderForFile(resolvePaths(this.getConfig<null | string>("dartTestLogFile", null))); }
+	get daemonPort(): undefined | string { return this.getConfig<null | string>("daemonPort", null); }
```
DanTup, line 63: I think this could use `number` instead of `string`?

src/extension/flutter/flutter_daemon.ts
```diff
@@ -47,15 +48,20 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
...
-			path.join(workspaceContext.sdks.flutter, flutterPath),
-			["daemon"].concat(daemonArgs),
-			workspaceContext.config?.flutterDaemonScript,
-		);
 
-		const flutterAdditionalArgs = config.for(vs.Uri.file(folder)).flutterAdditionalArgs;
-		const args = getGlobalFlutterArgs().concat(flutterAdditionalArgs).concat(execution.args);
-		this.createProcess(folder, execution.executable, args, { toolEnv: getToolEnv() });
+		if (!!config.daemonPort) {
+			this.createNcProcess(config.daemonPort);
```
DanTup, line 53: As above, if the user may set this in their global User Settings, we might need to make this only run for internal projects (eg. based on the presence of the custom Flutter script?).

src/extension/flutter/flutter_daemon.ts
```diff
@@ -47,15 +48,20 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
...
+		} else {
+			const execution = usingCustomScript(
+				path.join(workspaceContext.sdks.flutter, flutterPath),
+				["daemon"].concat(daemonArgs),
+				workspaceContext.config?.flutterDaemonScript,
```
DanTup, line 58: Is this redundant now? Will internal uses always have both `flutterDaemonScript` set and `config.daemonPort` or is it valid to use just the former?

src/extension/flutter/flutter_daemon.ts
```diff
@@ -65,6 +71,16 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
 		}
 	}
 
+	protected createNcProcess(port: string) {
+		vs.window.showInformationMessage(`About to create nc process with port: ${port}`);
```
DanTup, line 75: nit: this might read better as "Created nc process..." since by the time the user sees it, the process will almost certainly have been spawned.

Do we need to worry about `nc` not being available? (I guess users won't set the port number in config unless they're set up to use it, but if there might be situations where it's not there or not on `PATH` we might want to handle the error spawning the process and provide a message using `vs.window.showErrorMessage`).

src/extension/flutter/flutter_daemon.ts
```diff
@@ -65,6 +71,16 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
 		}
 	}
 
+	protected createNcProcess(port: string) {
```
DanTup, line 74: It may be worth some comments on this method explaining roughly what the setup is here.

I _think_ the setup is that the user already has a tunnel set up from the remote machine to the local machine, and this is causing the extension (running in the remote machine) to connect over that tunnel to a deamon on the local machine instead of spawning its own (so that it can discover the local devices)?

src/shared/utils/workspace.ts
```diff
@@ -53,11 +53,11 @@ export function tryProcessBazelFlutterConfig(logger: Logger, config: WritableWor
 			return path.join(bazelWorkspaceRoot, relOrAbsolute);
 		}
 
-		function makeScript(relOrAbsolute: string | undefined, replacesArgs = 1): CustomScript | undefined {
+		function makeScript(relOrAbsolute: string | undefined, replacesArgs = 1, makeFull = true): CustomScript | undefined {
```
DanTup, line 56: I can't see any code passing `false` for `makeFull` here - should there be?

src/debug/flutter_debug_impl.ts
helin24: Ah okay, I forgot to test with SDK DAPs. I added a check based on `customTool` and `daemonPort`.

I think we won't need this for tests because tests use the "device" that's on the same machine as our extension.

src/extension/flutter/flutter_daemon.ts
helin24: Unfortunately this isn't redundant 😞 The `flutterDaemonScript` will be used when a user wants to access devices that are on the same machine as the extension host, using `config.daemonPort` will be for when the devices are on a different machine.

src/extension/flutter/flutter_daemon.ts
helin24: Oops not even sure I want this message. I think I'll delete it as it's more for debugging.

I _think_ `nc` should always be available for users, but `handleError` should log if there's a problem, right? Although would it be hard for users to capture this error message on startup? If users are having problems I can also have them set the log file locations in their settings.

src/shared/utils/workspace.ts
helin24: Old code, thanks

package.json
```diff
@@ -2356,6 +2356,12 @@
 						"markdownDescription": "EXPERIMENTAL: Whether to look for custom script definitions at `dart/config/ide/flutter.json` in Bazel workspaces. Currently supported for macOS and Linux only.",
 						"scope": "window"
 					},
+					"dart.daemonPort": {
+						"type": "number",
+						"default": false,
+						"markdownDescription": "EXPERIMENTAL: Daemon port",
```
DanTup, line 2362: I've (very recently) started trying to highlight settings that are not relevant to most users more clearly. Some examples in 0b2fa58db3c5e7def01e98773824c08f6fe04647. It might be worth adding a note here that this is intended for internal Google developers?

package.json
```diff
@@ -2356,6 +2356,12 @@
...
+						"scope": "window"
```
DanTup, line 2363: If we're gong to read this via the `ResourceConfig` class, I think we should set this to `"resource"`.

src/debug/flutter_debug_impl.ts
DanTup: I think we can (should?) remove this code now, since it's being handled with `toolArgs`? (that way we won't be surprised with any differences between SDK/Dart-Code DAPs).

src/extension/flutter/flutter_daemon.ts
DanTup: Ah, gotcha :)

src/extension/providers/debug_config_provider.ts
```diff
@@ -672,6 +673,13 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 				if (renderer)
 					this.addArgsIfNotExist(args, "--web-renderer", renderer);
 			}
+
+			if (this.wsContext.config.forceFlutterWorkspace && conf.config.has("daemonPort")) {
+				const port = conf.config.get<number|undefined>("daemonPort");
```
DanTup, line 678: This should just be `config.daemonPort` and we should add a getter to `ResourceConfig` (this is the same as what you did originally, but in the `ResourceConfig` class instead of `Config`. One is global, and one supports resource-scoped (essentially workspace-folder/project scoped) settings.

DanTup (approved):
Some minor comments, but LGTM!

If you haven't already, if you rebase on master, the tests should be green there as of yesterday.

src/extension/flutter/flutter_daemon.ts
DanTup: Oops, missed this. I think `handleError` might only logging to logs/console. I'm not sure exactly how it'll fail if it's not available. You could perhaps test this by changing `nc` to just some invalid command `notnc` and see how it behaves.

src/extension/providers/debug_config_provider.ts
helin24: Oh is this why I was having trouble accessing `daemonPort` before? I'd only had it in the first config.

src/extension/flutter/flutter_daemon.ts
helin24: I added an error message, though I'll probably change this to link to an error message defined internally in the future.

src/extension/providers/debug_config_provider.ts
DanTup: I wouldn't have expected so. The only time these two should be different is in a multi-root workspace. Otherwise, it should always ready from `.vscode/settings.json` at the top level first, and then user settings if it's not defined there.

However, remote sessions do have some differences with config (in part to stop cross-platform paths being roamed, like trying to use Windows paths set in settings on a Linux remote). I don't know the full details, but it's possible that affected this (or that its presence and/or `"scope"` setting in our `package.json` affects it).

PR #4078, opened 2022-07-29
Translate folder paths for setting DevTools root

helin24:
The current project folder finding method does not work for internal projects that don't use pubspec files, but we need to return these folders with their relative paths for DevTools root to be set correctly. I think we could eventually do this in DDS, but this may be the best option to have this in the next release (I'm not sure and would be open to doing something less hard coded).

### DIFF
--- a/src/extension/commands/debug.ts
+++ b/src/extension/commands/debug.ts
@@ -62,7 +62,7 @@ export class DebugCommands implements IAmDisposable {
 	private suppressFlutterWidgetErrors = false;
 
 	constructor(private readonly logger: Logger, private context: Context, private workspaceContext: DartWorkspaceContext, readonly dartCapabilities: DartCapabilities, readonly flutterCapabilities: FlutterCapabilities, private readonly analytics: Analytics, pubGlobal: PubGlobal, flutterDaemon: IFlutterDaemon | undefined) {
-		this.vmServices = new VmServiceExtensions(logger, this);
+		this.vmServices = new VmServiceExtensions(logger, this, workspaceContext);
 		this.devTools = new DevToolsManager(logger, workspaceContext, this, analytics, pubGlobal, dartCapabilities, flutterCapabilities, flutterDaemon);
 		this.disposables.push(this.devTools);
 		this.debugOptions.name = "Dart Debug Options";
@@ -552,7 +552,7 @@ export class DebugCommands implements IAmDisposable {
 	}
 
 	private async handleCustomEventWithSession(session: DartDebugSessionInformation, e: vs.DebugSessionCustomEvent) {
-		this.vmServices.handleDebugEvent(session, e, this.workspaceContext.config.forceFlutterWorkspace)
+		this.vmServices.handleDebugEvent(session, e)
 			.catch((e) => this.logger.error(e));
 
 		if (e.event === "dart.webLaunchUrl") {
--- a/src/extension/commands/debug.ts
+++ b/src/extension/commands/debug.ts
@@ -61,7 +61,7 @@ export class DebugCommands implements IAmDisposable {
 	public readonly devTools: DevToolsManager;
 	private suppressFlutterWidgetErrors = false;
 
-	constructor(private readonly logger: Logger, private context: Context, workspaceContext: DartWorkspaceContext, readonly dartCapabilities: DartCapabilities, readonly flutterCapabilities: FlutterCapabilities, private readonly analytics: Analytics, pubGlobal: PubGlobal, flutterDaemon: IFlutterDaemon | undefined) {
+	constructor(private readonly logger: Logger, private context: Context, private workspaceContext: DartWorkspaceContext, readonly dartCapabilities: DartCapabilities, readonly flutterCapabilities: FlutterCapabilities, private readonly analytics: Analytics, pubGlobal: PubGlobal, flutterDaemon: IFlutterDaemon | undefined) {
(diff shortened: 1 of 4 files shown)


DanTup:
I'm not sure I understand the requirement here, though it seems like this change would affect a lot more than DevTools. This function (`getAllProjectFolders `) is used in a lot of places that need to know what the paths are to all your project roots and callers expect it to return file system paths. It doesn't seem safe to me to start returning `google3:` URIs to all of those places.

Can you describe in more detail what the goal is? If the issue is that we're not locating the roots (because there's no pubspec) then I think we want to update `findProjectFolders` which is what detects the roots (by looking for a pubspec, or special trigger files we use for creating new projects).

If somewhere also needs URIs in the form `google3:`, I think we should restrict that translation to exactly where it's needed to avoid other callers needing to worry about getting back URIs (instead of file paths) from these functions. (I think we should also restrict this to only in internal workspaces to reduce the change of false positives because of the file path matching).

src/extension/commands/debug.ts
```diff
@@ -552,7 +552,7 @@ export class DebugCommands implements IAmDisposable {
 	}
 
 	private async handleCustomEventWithSession(session: DartDebugSessionInformation, e: vs.DebugSessionCustomEvent) {
-		this.vmServices.handleDebugEvent(session, e)
+		this.vmServices.handleDebugEvent(session, e, this.workspaceContext.config.forceFlutterWorkspace)
```
DanTup, line 555: I think it might be neater to pass the whole wsContext or config into the constructor (up on line 65) so that it's just generally available to that class.

src/extension/flutter/vm_service_extensions.ts
```diff
@@ -86,10 +84,20 @@ export class VmServiceExtensions {
 
 	// TODO: Remove this function (and the call to it) once the fix has rolled to Flutter beta.
 	// https://github.com/flutter/flutter-intellij/issues/2217
```
DanTup, line 86: We should remove this TODO now since it's no longer relevant.

src/extension/flutter/vm_service_extensions.ts
```diff
@@ -52,13 +52,13 @@ export class VmServiceExtensions {
...
 	/// Handles an event from the Debugger, such as extension services being loaded and values updated.
-	public async handleDebugEvent(session: DartDebugSessionInformation, e: vs.DebugSessionCustomEvent): Promise<void> {
+	public async handleDebugEvent(session: DartDebugSessionInformation, e: vs.DebugSessionCustomEvent, forceFlutterWorkspace?: boolean): Promise<void> {
 		if (e.event === "dart.serviceExtensionAdded") {
 			this.handleServiceExtensionLoaded(session, e.body.extensionRPC as VmServiceExtension, e.body.isolateId as string | null | undefined);
 
 			try {
 				if (e.body.extensionRPC === VmServiceExtension.InspectorSetPubRootDirectories) {
-					const projectFolders = await getAllProjectFolders(this.logger, getExcludedFolders, { requirePubspec: true, searchDepth: config.projectSearchDepth });
+					const projectFolders = await getAllProjectFolders(this.logger, getExcludedFolders, { requirePubspec: !forceFlutterWorkspace, searchDepth: forceFlutterWorkspace ? 1 : config.projectSearchDepth, forceFlutterWorkspace });
```
DanTup, line 61: Are these arguments (ignoring pubspec etc.) specific to this call, or do we want to use them everywhere we look for projects? There are quite a few places that call `getAllProjectFolders` and I suspect they should all behave the same. If so (but we don't think we need to fix this now), we could add a TODO. Probably the fix is to make this function simpler so everyone that calls it doesn't need to provide all these arguments explicitly.

src/shared/utils/fs.ts
```diff
@@ -216,7 +216,7 @@ export async function findProjectFolders(logger: Logger, roots: string[], exclud
 	const projectFolderPromises = allPossibleFolders.map(async (folder) => ({
 		exists: options && options.requirePubspec
 			? await hasPubspecAsync(folder)
-			: await hasPubspecAsync(folder) || await hasCreateTriggerFileAsync(folder) || await isFlutterRepoAsync(folder),
+			: options.forceFlutterWorkspace || await hasPubspecAsync(folder) || await hasCreateTriggerFileAsync(folder) || await isFlutterRepoAsync(folder),
```
DanTup, line 219: If `forceFlutterWorkspace` is set to `true` but `searchDepth` is not set to `1`, I think the results here would be unexpected (they'd include every folder at each of the `searchDepth` levels? If so, perhaps we should name this flag `onlyWorkspaceRoots` or similar and ignore `searchDepth`?

src/shared/vscode/utils.ts
```diff
@@ -62,7 +62,7 @@ function getAnalysisOptionsExcludedFolders(
 export async function getAllProjectFolders(
 	logger: Logger,
 	getExcludedFolders: ((f: WorkspaceFolder | undefined) => string[]) | undefined,
-	options: { sort?: boolean; requirePubspec?: boolean, searchDepth: number, workspaceFolders?: WorkspaceFolder[] },
+	options: { sort?: boolean; requirePubspec?: boolean, searchDepth: number, workspaceFolders?: WorkspaceFolder[], forceFlutterWorkspace?: boolean },
```
DanTup, line 65: (same comment as above, maybe something like `onlyWorkspaceRoots?: boolean`?)

DanTup (approved):

PR #4082, opened 2022-08-01
Remove preview bazel option

helin24:
Is it okay to remove this option entirely? It looks like checking for Fuchsia workspaces is similar, so hopefully it's not too much work to always check for a bazel workspace.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2350,12 +2350,6 @@
 						"description": "Whether to normalize file casings before sending them to the LSP server. This may fix issues with file_names lints not disappearing after renaming a file if the VS Code API continues to use the original casing.",
 						"scope": "window"
 					},
-					"dart.previewBazelWorkspaceCustomScripts": {
-						"type": "boolean",
-						"default": false,
-						"markdownDescription": "EXPERIMENTAL: Whether to look for custom script definitions at `dart/config/ide/flutter.json` in Bazel workspaces. Currently supported for macOS and Linux only.",
-						"scope": "window"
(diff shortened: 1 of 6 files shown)


DanTup:
> Is it okay to remove this option entirely? 

That probably depends on whether there's any chance this behaviour could negatively impact anything (something I don't know the answer to, but perhaps you do). Usually when removing a preview flag like this (for ex. with LSP, and soon DAP) I will instead rename it (to not have preview) and change the default to true. This has the same effect (opting everyone in by default) but still leaves a way to quickly opt-out if reports come in that something isn't working.

We could do the same here (and note in the setting description that this is a temporary setting and if anyone finds themselves needing to set it they should file an issue before it is removed), but you have more context than me - if you don't think it could break anyone, I'm also happy to just remove it.

Note: I'm shipping a release today but don't want to merge this without your response, so this won't be included - but we can do another release after merging it if there's value in having it shipped (we can also ship it in a pre-release version first if you'd like to have some other users test it before it rolls out.. unfortunately it's not something I can really help test!).


> hopefully it's not too much work to always check for a bazel workspace.

Yeah, this is a minor amount of work compared to what else happens during startup, so I wouldn't be concerned about the extra checks.

helin24:
@DanTup thanks for the extra info! It sounds like we're not worried about the check generating work, so let's go ahead and remove this flag. It won't break anyone using VSCode currently (as no one is using it internally yet :) )

PR #4117, opened 2022-08-23
Fix daemon not ending / restarting on user request

helin24:
We were previously always sending a daemon shutdown request, which would kill the shutdown process entirely if the daemon never started up.

This also shows the no devices message if the daemon doesn't start up for internal cases instead of the more generic daemon failed to start message.


DanTup (approved):
LGTM. The test failures are because I committed a `.only()` in a test that your changes are based on. I've since fixed on master, but the tests there are still red so I need to look at, it's fine to land this as they're unrelated.

PR #4146, opened 2022-09-06
Update dart.json

d-khomenko:
final > var

### DIFF
--- a/snippets/dart.json
+++ b/snippets/dart.json
@@ -54,7 +54,7 @@
 			"prefix": "for",
 			"description": "Insert a for loop.",
 			"body": [
-				"for (var i = 0; i < ${1:count}; i++) {",
+				"for (final i = 0; i < ${1:count}; i++) {",
 				"  $0",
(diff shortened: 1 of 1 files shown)


DanTup:
> final > var

While I personally agree, I believe this is subjective. Fortunately what you want is already supported - if you enable the `prefer_final_locals` lint and are on a recent SDK, your snippets will automatically use `final`. This is handled by the Dart analysis server, as the snippets you've modified here are legacy versions for older SDKs.

Sep-06-2022 14-56-31

(FWIW, one of the changes made here produces invalid code, you can't use `final` in a `for` loop like that, as you cannot increment the counter).

I'll close this, but if you're on a recent SDK and enabling the `prefer_final_locals` lint does not cause snippets to be inserted with `final`, please let me know!

d-khomenko:
Oops, sorry, automatically changed, yes, lint added, everything works)

DanTup:
Great, thanks for confirming! :)

PR #4158, opened 2022-09-14
WIP use local extension for starting daemon

helin24:


src/extension/extension.ts
```diff
@@ -271,7 +272,25 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean
 			};
 		}
 
-		flutterDaemon = new FlutterDaemon(logger, workspaceContext as FlutterWorkspaceContext, flutterCapabilities, runIfNoDevices);
+		if (workspaceContext.config.forceFlutterWorkspace) {
+			let resultFromLocalExtension = null;
+			try {
+				resultFromLocalExtension = await vs.commands.executeCommand<string>("dartlocaldevice.startDaemon", workspaceContext.config.flutterToolsScript?.script, "expose_devices", workspaceContext.config.flutterSdkHome);
```
DanTup, line 278: It may be worth making these args an object (`[{ flutterToolsScript: xxx, flutterSdkHome: xxx}]` etc.) so if in future they need to change it'll be a little easier (because they're named on the other end, rather than just being indexes).

DanTup, line 278: It might also be worth either having a timeout on this, or showing a progress indicator that allows the user to stop waiting for it if it takes too long (since if we get stuck here, it'll prevent the rest of the extension activating).

src/extension/extension.ts
```diff
@@ -271,7 +272,25 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean
...
+			} catch (e) {
+				// Command may not be available.
+			}
+			if (resultFromLocalExtension !== null) {
+				const resultMessage = resultFromLocalExtension.toString();
+				const results = resultMessage.match(/Device daemon is available on remote port: (\d+)/i);
+				if (results !== null && results?.length > 1) {
+					portFromLocalExtension = parseInt(results[1]);
+				} else if (resultMessage !== null) {
+					vs.window.showErrorMessage(resultMessage);
```
DanTup, line 288: Generally I try to always write to the logger (`logger.error()`) when we show errors. Sometimes users don't notice the toast notifications and VS Code hides them after some number of seconds (5-10?). By writing them to the logger they'll show up both in the dev console (Help -> Toggle Developer Tools) and also if they're logging (with the Capture Logs command).

DanTup (approved):
Some minor comments (that you're welcome to ignore or do in another PR), but lgtm!

src/extension/extension.ts
helin24: I moved the timeout function to utils; let me know if you intended to keep that test-specific.

helin24:
Oops are the test failures due to this change?

DanTup:
> Oops are the test failures due to this change?

I don't think so. One of them seems to be occurring on master too (taking a look now) and the other (protobuf path mismatch) looks like something I thought I'd investigated recently, but can't find any evidence of fixing it or filing an SDK issue. I'll dig into this more after fixing the first one.

(Feel free to land in the meantime if you're ready)

DanTup:
Tests on master are green now. I think maybe the other failure seen here is something I'd previously fixed on master, as this branch is a little out of date now. You could rebase on master if you want to see them all go green before merging, but I'm fairly sure it'll be green once merged.

src/extension/extension.ts
helin24: Ah makes sense. I'll keep that mind and update depending on user feedback.

PR #4178, opened 2022-09-26
Infer a default CWD for workspaces with multiple folders if they have a common ancestor

DanTup:
@helin24 can you verify this works for your case?


helin24 (approved):
This works for my cases! Thanks!

PR #4182, opened 2022-09-27
Pass automatic daemon port to debug config

helin24:
I forgot to pass the port from the local device extension to debug config (oops!). Is there a better way to pass this through these functions?

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -297,7 +297,7 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean 
 
 		flutterDaemon = new FlutterDaemon(logger, workspaceContext as FlutterWorkspaceContext, flutterCapabilities, runIfNoDevices, portFromLocalExtension);
 
-		deviceManager = new FlutterDeviceManager(logger, flutterDaemon, config, workspaceContext, runIfNoDevices);
+		deviceManager = new FlutterDeviceManager(logger, flutterDaemon, config, workspaceContext, runIfNoDevices, portFromLocalExtension);
 
 		context.subscriptions.push(deviceManager);
 		context.subscriptions.push(flutterDaemon);
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -522,7 +522,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 		debugConfig.additionalProjectPaths = debugConfig.additionalProjectPaths || vs.workspace.workspaceFolders?.map((wf) => fsPath(wf.uri));
 		debugConfig.args = debugConfig.args || [];
 		debugConfig.vmAdditionalArgs = debugConfig.vmAdditionalArgs || conf.vmAdditionalArgs;
-		debugConfig.toolArgs = await this.buildToolArgs(debugType, debugConfig, conf);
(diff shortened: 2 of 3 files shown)


src/extension/providers/debug_config_provider.ts
```diff
@@ -522,7 +522,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 		debugConfig.additionalProjectPaths = debugConfig.additionalProjectPaths || vs.workspace.workspaceFolders?.map((wf) => fsPath(wf.uri));
 		debugConfig.args = debugConfig.args || [];
 		debugConfig.vmAdditionalArgs = debugConfig.vmAdditionalArgs || conf.vmAdditionalArgs;
-		debugConfig.toolArgs = await this.buildToolArgs(debugType, debugConfig, conf);
+		debugConfig.toolArgs = await this.buildToolArgs(debugType, debugConfig, conf, deviceManager?.portFromLocalExtension);
```
DanTup, line 525: I don't think you need to pass this all the way down, the implementation of `buildFlutterToolArgs` is inside this class, so it can probably read `this.deviceManager?.portFromLocalExtension` directly?

src/extension/providers/debug_config_provider.ts
```diff
@@ -680,7 +680,9 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 					this.addArgsIfNotExist(args, "--web-renderer", renderer);
 			}
 
-			if (this.wsContext.config.forceFlutterWorkspace && conf.daemonPort) {
+			if (this.wsContext.config.forceFlutterWorkspace && portFromLocalExtension) {
+				this.addArgsIfNotExist(args, "--daemon-connection-port", portFromLocalExtension.toString());
+			} else if (this.wsContext.config.forceFlutterWorkspace && conf.daemonPort) {
```
DanTup, line 685: Maybe worth combining these two ports into a single variable to slightly reduce duplication of the arg?

```dart
const daemonPort = this.deviceManager?.portFromLocalExtension ?? conf.daemonPort;
if (this.wsContext.config.forceFlutterWorkspace && daemonPort) {
	this.addArgsIfNotExist(args, "--daemon-connection-port", daemonPort.toString());
}
```

src/extension/extension.ts
```diff
@@ -297,7 +297,7 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean
 
 		flutterDaemon = new FlutterDaemon(logger, workspaceContext as FlutterWorkspaceContext, flutterCapabilities, runIfNoDevices, portFromLocalExtension);
 
-		deviceManager = new FlutterDeviceManager(logger, flutterDaemon, config, workspaceContext, runIfNoDevices);
+		deviceManager = new FlutterDeviceManager(logger, flutterDaemon, config, workspaceContext, runIfNoDevices, portFromLocalExtension);
```
DanTup, line 300: It might be a little clearer to name this something like `daemonPortOverride` which describes what it is rather than just where it's from.

DanTup (approved):
Some minor nits/comments, but LGTM!

src/extension/extension.ts
helin24: I renamed this in the device manager instead

DanTup: Those changes look good, thanks!

PR #4195, opened 2022-10-03
Check for workspace directory shared parent if no default cwd WIP

helin24:


src/extension/providers/debug_config_provider.ts
```diff
@@ -88,7 +88,8 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 				defaultCwd = fsPath(folder.uri);
 				logger.info(`Setting folder/defaultCwd based on open file: ${defaultCwd}`);
 			}
-		} else if (!folder && vs.workspace.workspaceFolders && vs.workspace.workspaceFolders.length >= 1) {
+		}
+		if (!defaultCwd && !folder && vs.workspace.workspaceFolders && vs.workspace.workspaceFolders.length >= 1) {
```
DanTup, line 92: I'm not sure if this will cover all cases. If we call into the first block (because we have an open file and no `defaultCwd`), we may infer a cwd from the open file, which will then cause us to skip the second block (because `!defaultCwd` is not set).

I feel like in the first block, `openFile` should probably really be something like `debugConfig.program ?? openFile` so we only ever infer from the open file if there is not an explicit program.

I'm gonna try and add some better tests around this (since we don't seem to cover all the cases) and then maybe it'll be easier to understand the best way.

helin24:
Closing in favor of #4197

PR #4196, opened 2022-10-04
Update bazel tests to use two workspace folders

DanTup:
Previously the bazel tests used a single folder, but this meant the `cwd` for running would always be that folder and didn't allow testing that we'd pick a common ancestor (given we don't infer a good `cwd` from the special `program`).

So this change:

- Moves the bazel root up from `bazel_workspace/flutter_hello_world_bazel` to `bazel_workspace` (by moving the `WORKSPACE` file and the `flutter.json` file, and updating the paths in `flutter.json`)
- Adds a `.code-workspace` file that includes this folder and a new (empty) `flutter_hello_world_bazel_2` alongside it
- Updates tests to open the `.code-workspace` instead of the folder directly and updates the "has opened the correct folders" to detect this workspace being opened
- Moves some `program`/`cwd` massaging into the `configureProgramAndCwd` method (no functional changes, but it seems like this code makes more sense inside there)

Additonally:

- Tweaks the doctor test to not fail locally when using a non-official Flutter fork
- Adds a comment in flutter_run test about the expectation not currently being correct (this should match the bug @helin24 is hitting)

This does not fix any issues, it's just refactoring in preparation so that the "fix" change will be less noisy.

@helin24 FYI. I'll land this if/when it's all green, but pinging just FYI.


DanTup:
This failure seems unrelated (I had it on another branch yesterday), so landing this and will investigate that on master.

PR #4197, opened 2022-10-04
Don't infer cwd from workspace folder when there's an explicit program

DanTup:
@helin24 does this solve the issue you had?

If there's a `program` we will try to infer a `cwd` from it, but if that fails, we should no longer use the workspace folder. We'll only use the workspace folder directly is there was no open file _or_ `program`.

Then, if we still don't have one (which we won't for a special `target` that doesn't have an enclosing `WorkspaceFolder`) we'll look for a common ancestor for all workspace folders.

### DIFF
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -413,7 +413,7 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 			}
 		}
 
-		// Convert to an absolute paths (if possible).
+		// Convert any relative paths to absolute paths (if possible).
 		if (defaultCwd && !path.isAbsolute(defaultCwd) && folder) {
 			debugConfig.cwd = path.join(fsPath(folder.uri), defaultCwd);
 			this.logger.info(`Converted defaultCwd to absolute path: ${defaultCwd}`);
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -369,26 +369,48 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 		const logger = this.logger;
 		const isAttachRequest = debugConfig.request === "attach";
 
-		let defaultCwd = folder ? fsPath(folder.uri) : undefined;
-		if (openFile && !defaultCwd) {
-			folder = workspace.getWorkspaceFolder(Uri.file(openFile));
-			if (folder) {
+		// Try to infer a default working directory:
+		//
+		//   1. Provided by the user
+		//   2. Inferred from the likely entry point (`program ?? openFile`)
+		//   3. From the active workspace folder for this launch configuration (only if no explicit entry point)
+		//   4. A common ancestor from the workspace folders
+		//
+		// The default may be overwritten further down if we locate a project root
+		// while walking up the tree from the `program`.
+
+		let defaultCwd = debugConfig.cwd;
+		if (!defaultCwd) {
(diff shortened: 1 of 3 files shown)


src/extension/providers/debug_config_provider.ts
```diff
@@ -369,35 +369,61 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
...
+				else
+					logger.info(`Unable to infer defaultCwd from open workspace (no common ancestor)`);
 			}
-			if (defaultCwd)
-				logger.info(`Setting defaultCwd based on common ancestor of open folders: ${defaultCwd}`);
-			else
-				logger.info(`Unable to infer defaultCwd from open workspace (no common ancestor)`);
 		}
 
 		// Convert to an absolute paths (if possible).
```
helin24, line 416: path

helin24 (approved):
This fixes the problem we were having! Thanks!

maik-kluwe:
Did you tests the change with opened workspace?
Because it breaks my current configuration:

`{
            // A name for the launch config. This will show in the dropdown on the Run side bar.
            "name": "test_app",
            // This can be "launch" to start an app, or "attach" to attach to an existing app.
            "request": "launch",
            // This should always be "dart" for Dart/Flutter apps.
            // This selects the Dart debugger.
            "type": "dart",
            // Set to "test" in an app with tests to run all tests.
            "program": "lib/main.dart",
            // "debug", "profile" or "release".
            "flutterMode": "debug",
            "args": [
                "-d",
                "chrome",
            ],
        }`
        
If i remove the program it runs perfectly now.

DanTup:
@maik-kluwe can you provide more details about your exact setup? What do you mean by "opened workspace"? Do you mean a `.code-workspace` file? If so, is this launch configuration inside the `.code-workspace` or inside `.vscode/launch.json` inside one of the workspace folders?

If the config above is inside the `.code-workspace` file then this is expected. The issue is that relative paths in files at the workspace level are ambiguous, for example:

```js
{
	"folders": [{
			"path": "/Users/danny/Desktop/dart_sample"
		}, {
			"path": "/Users/danny/Desktop/dart_sample_2"
	}],
	"launch": {
		"version": "0.2.0",
		"configurations": [
			{
				"name": "Dart",
				"type": "dart",
				"request": "launch",
				"program": "bin/main.dart" // Is this path relatve to dart_sample or dart_sample_2 ?
			}
		]
	}
}
```

If you have launch configurations with relative paths in the workspace file, you should use variables to indicate which workspace folder they are relative to:

https://code.visualstudio.com/docs/editor/variables-reference#_variables-scoped-per-workspace-folder

For example:

```js
{
	"folders": [{
			"path": "/Users/danny/Desktop/dart_sample"
		}, {
			"path": "/Users/danny/Desktop/dart_sample_2"
	}],
	"launch": {
		"version": "0.2.0",
		"configurations": [
			{
				"name": "Dart",
				"type": "dart",
				"request": "launch",
				"program": "bin/main.dart",
				"cwd": "${workspaceFolder:dart_sample_2}", // This uses the workspaceFolder path from dart_sample_2
			}
		]
	}
}
```

If this isn't the issue you are seeing, please file a new issue and include specific details that will help me reproduce. Thanks!

DanTup:
@maik-kluwe you are correct - if the launch config is inside the project folder, this should work - I'll take a look shortly. Thanks!

DanTup:
I can repro this - have opened #4239 and will do some digging. Sorry for the bother!

maik-kluwe:
It works 👍

PR #4207, opened 2022-10-06
Ping the daemon WIP

helin24:
@DanTup `promptToReloadExtension` doesn't run activate again, so I'm not sure if there's something else I should use here to run the local device command again. Or perhaps I should move that logic into another area that is run again when `promptToReloadExtension` is called?

### DIFF
--- a/src/extension/flutter/flutter_daemon.ts
+++ b/src/extension/flutter/flutter_daemon.ts
@@ -54,6 +54,7 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements 
 			this.startPing();
 		} else if (workspaceContext.config.forceFlutterWorkspace && config.daemonPort) {
 			this.createNcProcess(config.daemonPort);
+			this.startPing(workspaceContext.config.restartMacDaemonMessage);
 		} else {
 			const execution = usingCustomScript(
 				path.join(workspaceContext.sdks.flutter, flutterPath),
@@ -74,7 +75,8 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements 
 		}
 	}
 
-	public startPing() {
+	public startPing(customMessage?: string) {
+		const message = customMessage ?? "The daemon connection was lost. Reload the extension to restart the daemon.";
 		this.pingIntervalId = setInterval(async () =>  {
 			try {
 				const result = await withTimeout(this.daemonVersion(), "The daemon connection was lost", 10);
@@ -82,7 +84,7 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements 
 				clearInterval(this.pingIntervalId);
 				this.logger.error(e);
 				this.hasShownTerminationError = true;
-				promptToReloadExtension(`The daemon connection was lost. Reload the extension to restart the daemon.`);
+				promptToReloadExtension(message);
 			}
 		}, 60 * 1000);
 	}
--- a/src/extension/flutter/flutter_daemon.ts
+++ b/src/extension/flutter/flutter_daemon.ts
@@ -77,14 +77,14 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements 
 	public startPing() {
 		this.pingIntervalId = setInterval(async () =>  {
 			try {
-				const result = await withTimeout(this.daemonVersion(), "The daemon connection was lost", 1);
+				const result = await withTimeout(this.daemonVersion(), "The daemon connection was lost", 10);
 			} catch (e) {
 				clearInterval(this.pingIntervalId);
 				this.logger.error(e);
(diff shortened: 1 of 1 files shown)


DanTup:
> `promptToReloadExtension` doesn't run `activate` again, so I'm not sure if there's something else I should use here to run the local device command again

Hmm, sounds like something is wrong. `promptToReloadExtension` executes the `_dart.reloadExtension` command:

https://github.com/Dart-Code/Dart-Code/blob/4486b8203bcac6bd3c8cd2db05903b2e87aa6cce/src/extension/utils.ts#L260

And that deactivates and re-activates everything:

https://github.com/Dart-Code/Dart-Code/blob/f4ea9438578efd0b766144a08374c06c5a4dd0a6/src/extension/extension.ts#L165-L169

If it's not happening, I wonder if there are exceptions or something occurring? I would try debugging or adding logging to see if it's not going through those paths.

src/extension/flutter/flutter_daemon.ts
```diff
@@ -50,6 +50,7 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
 
 		if (portFromLocalExtension) {
 			this.createNcProcess(portFromLocalExtension);
+			this.startPing();
```
DanTup, line 53: Should we start this only once we know the daemon is up and running? I'm wondering if otherwise it might fail before it's really set up, and then prompt for a restart?

src/extension/flutter/flutter_daemon.ts
```diff
@@ -72,6 +73,18 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
 		}
 	}
 
+	public startPing() {
+		const intervalId = setInterval(async () =>  {
+			try {
+				const result = await withTimeout(this.daemonVersion(), "The daemon connection was lost", 1);
```
DanTup, line 79: I was a bit concerned by this timeout, but it turns out there are two `withTimeout`s and one of them is in `ms` and one in `s` 😄 (my fault, but the one in ms is just in the legacy debug adapter and not used anywhere else).

Although, perhaps every second is still a bit frequent if it's going over a network. Would doing it every 10s be a much worse experience?

(Good idea to use `version` to avoid needing a new protocol message :-) )

src/extension/flutter/flutter_daemon.ts
```diff
@@ -72,6 +73,18 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
...
+			} catch (e) {
+				clearInterval(intervalId);
```
DanTup, line 81: We should probably ensure this is done on `dispose` too so when we unload we don't get an additional call (and a prompt) before it stops trying to ping. We could pull the `intervalId` up to the class, or do something like this:

```
this.disposables.push({ dispose: () => clearInterval(intervalId) });
```

src/extension/flutter/flutter_daemon.ts
helin24: Ah sorry yeah, the timeout numbers are only for testing. 10s sounds fine to me for this check, and I could do a much larger number for how often this pings, like every minute or something.

src/extension/flutter/flutter_daemon.ts
helin24: I think this should be okay because we only get a port from the local extension after the process has started. If the process has started but there's no response, then it would be appropriate to restart.

src/extension/flutter/flutter_daemon.ts
helin24: I added `intervalId` as a class var

src/extension/flutter/flutter_daemon.ts
```diff
@@ -50,6 +51,7 @@ export class FlutterDaemon extends StdIOService<UnknownNotification> implements
 
 		if (portFromLocalExtension) {
 			this.createNcProcess(portFromLocalExtension);
+			this.startPing();
 		} else if (workspaceContext.config.forceFlutterWorkspace && config.daemonPort) {
 			this.createNcProcess(config.daemonPort);
```
DanTup, line 56: Do we need to startPing() in this branch?

DanTup (approved):

DanTup:
There are some known test failures on the flutter_debug_sdk_dap bot. I've made some progress tracking this down today (we seem to be leaking some `dart devtools` processes on the bots that are slowly consuming all the memory and causing the bots to slow down and fail), but haven't managed to fix it yet. Feel free to land if it's just that bot failing and the non-SDK-DAP version passes.

src/extension/flutter/flutter_daemon.ts
helin24: I think this won't be used very much any more, but it would make sense to add it here too.

PR #4224, opened 2022-10-18
Remove sync script

helin24:
This is no longer needed since we only support running VS Code on a remote machine. (the original purpose was to make sure remote changes were synced on a local machine)


DanTup (approved):
LGTM!

PR #4232, opened 2022-10-22
adding wrap with expanded to flutter refactor commands

georgeman93:
I am constantly wrapping Flutter widgets with Expanded. I wouldn't be surprised if it is a more common action than wrapping with Centre. 

I'm mostly new to typescript and vscode extensions
I am not sure if these even works.

I was able to see debugging statements triggered when I executed commands like Flutter Doctor, but not any of the refactoring actions even though I could see that they were successfully registered in `Dart-Code/src/extension/commands/flutter_outline.ts`
I was hoping to get some guidance as to why this is the case (or if it is a known problem)

I am running 
2017 Macbook Pro Monterey 12.5.1
VSCode 1.72.2
NPM 8.19.2

### DIFF
--- a/package.json
+++ b/package.json
@@ -572,6 +572,11 @@
 				"category": "Flutter"
 			},
 			{
+				"command": "_flutter.outline.refactor.flutter.wrap.expanded",
+				"title": "Wrap with Expanded",
+				"category": "Flutter"
+			},
+			{
 				"command": "_flutter.outline.refactor.flutter.wrap.padding",
 				"title": "Wrap with Padding",
 				"category": "Flutter"
@@ -935,6 +940,10 @@
 				},
 				{
 					"when": "false",
+					"command": "_flutter.outline.refactor.flutter.wrap.expanded"
+				},
+				{
+					"when": "false",
 					"command": "_flutter.outline.refactor.flutter.wrap.padding"
 				},
 				{
@@ -1189,6 +1198,10 @@
 					"command": "_flutter.outline.refactor.flutter.wrap.center"
 				},
 				{
+					"when": "view == dartFlutterOutline && dart-code:widgetSupports:refactor.flutter.wrap.expanded",
+					"command": "_flutter.outline.refactor.flutter.wrap.expanded"
+				},
+				{
 					"when": "view == dartFlutterOutline && dart-code:widgetSupports:refactor.flutter.wrap.padding",
 					"command": "_flutter.outline.refactor.flutter.wrap.padding"
 				},
@@ -1226,6 +1239,10 @@
 				{
 					"when": "view == dartFlutterOutline && viewItem == dart-code:isSelectedWidget && dart-code:widgetSupports:refactor.flutter.wrap.center",
 					"command": "_flutter.outline.refactor.flutter.wrap.center"
+				},
+				{
+					"when": "view == dartFlutterOutline && viewItem == dart-code:isSelectedWidget && dart-code:widgetSupports:refactor.flutter.wrap.expanded",
+					"command": "_flutter.outline.refactor.flutter.wrap.expanded"
 				},
 				{
 					"when": "view == dartFlutterOutline && viewItem == dart-code:isSelectedWidget && dart-code:widgetSupports:refactor.flutter.wrap.padding",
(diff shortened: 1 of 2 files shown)


DanTup:
Thanks for the PR! The changes you've made are related to allowing existing code actions to be accessed from the Flutter Outline view and do not provide an additional refactor. Those refactors come from the Dart Analysis Server, for example:

Wrap with Center:
Implementation: https://github.com/dart-lang/sdk/blob/ea70de31853b20a9f0da1f395b27bb0ffc55de60/pkg/analysis_server/lib/src/services/correction/dart/flutter_wrap.dart#L109-L119
Tests: https://github.com/dart-lang/sdk/blob/ea70de31853b20a9f0da1f395b27bb0ffc55de60/pkg/analysis_server/test/src/services/correction/assist/flutter_wrap_center_test.dart

@bwilkerson are there any general rules for what should/shouldn't be included there? I've been asked about being able to customise this menu before (eg. a way for a user define that they want a "Wrap in Foo" - although interestingly one example given was also Expanded). I can imagine the list could get long if we added all widgets that are useful to _some_ users and I wonder if maybe it's worth making this user-customisable. Given the existing wraps are just a subclass that has a constant string in a getter, having one that takes a String in the constructor and constructing instances from some user-supplied names doesn't seem like it should be too difficult - WDYT? If we were happy for it to be LSP-only we could use existing LSP client config (where we have line length etc.) and not need to find another way to define it (although whether that sort of thing would be better somewhere like analysis_options, I'm not sure).

Another option could be for VS Code to clone the "wrap with generic" assist and replace the placeholder client-side. This would keep things simpler on the server, although at the expense of only supporting this in VS Code.

bwilkerson:
> ... are there any general rules for what should/shouldn't be included there?

No. The primary concern that I'm aware of is the one you mentioned: if we included all of the possible widgets in the list we expect that it would become unusable.

> ... I wonder if maybe it's worth making this user-customisable.

We received some feedback from a recent user survey that suggests that users would like to be able to customize the list, including adding assists for their own widgets. We haven't really discussed the pros and cons of the various options yet, so I'd hold off on implementing it until we've had a chance to convince ourselves that we're implementing the right thing, but I do think it's a good time to start those conversations.

> Given the existing wraps are just a subclass that has a constant string in a getter, having one that takes a String in the constructor and constructing instances from some user-supplied names doesn't seem like it should be too difficult - WDYT?

The existing wraps
- can wrap either one or multiple widgets, depending on the style of the wrapping widget,
- assume that the wrapping widget is defined in the `flutter` package so that we can find the widget and add an import when necessary.

I'd like to preserve those qualities of the operation because I think they add value. In other words, I think we need more than just a single string, but yes, we ought to be able to make it configurable.

> If we were happy for it to be LSP-only we could use existing LSP client config ...

The other implication of that choice is that it becomes a per user configuration option. That might be the right choice, but it might be better if every package could specify the configuration in the analysis options file so that the whole team has access to the same configuration. Or it might be something we want package authors to be able to contribute so that the configuration is shared by all of the users of their package. Or maybe some combination.

There are lots of possibilities that we should consider carefully.

> Another option could be for VS Code to clone the "wrap with generic" assist and replace the placeholder client-side.

I don't think that would give us the best possible UX, so I'd prefer to explore the other options first.

NoriThorleifs:
Definitely something I can see myself using!
I'd even advocate for a "Wrap all children in Expanded" refactor command when you're working with a Column, Row, etc.
But that's an idea for another time :)

DanTup:
I think we should capture the requests here in a proper SDK issue, since this PR won't be merged (it doesn't do what was intended) and isn't a very visible place for discussion.

I've opened https://github.com/dart-lang/sdk/issues/56265 with some notes - please add any details/information that I have missed.

I'll close this, since merging it won't work and the work needs to be done elsewhere.

DanTup:
By the way, I noticed that user-defined snippets in VS Code can use `$TM_SELECTED_TEXT` to wrap a selection. So you can create your own snippets to do this, although they will need invoking/selecting from the command palette (maybe there's a way to optimise this, but I couldn't see anything obvious).

https://github.com/user-attachments/assets/cd71282e-7b4d-43ce-beed-f605efc3ec7e

PR #4241, opened 2022-11-02
Use tools script for internal attach

helin24:
This only required a small change to use the generic tools script, as we don't use a separate script for attach, but hopefully this doesn't break anything else.

### DIFF
--- a/src/extension/providers/debug_config_provider.ts
+++ b/src/extension/providers/debug_config_provider.ts
@@ -586,6 +586,10 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 					: this.wsContext.config.flutterRunScript;
 				debugConfig.customTool = customScript?.script;
 				debugConfig.customToolReplacesArgs = customScript?.replacesArgs;
+			} else if (isAttach && !debugConfig.customTool) {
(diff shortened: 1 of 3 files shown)


src/extension/providers/debug_config_provider.ts
```diff
@@ -710,7 +714,9 @@ export class DebugConfigProvider implements DebugConfigurationProvider {
 				if (renderer)
 					this.addArgsIfNotExist(args, "--web-renderer", renderer);
 			}
+		}
 
+		if (!isAttach || this.wsContext.config.forceFlutterWorkspace) {
```
DanTup, line 719: I don't think this condition is doing anything and could be removed? (that is, we enter this block if not attaching or if `this.wsContext.config.forceFlutterWorkspace` is `true`, but line 721 means we don't do anything unless `this.wsContext.config.forceFlutterWorkspace` is `true` so the `!isAttach` condition isn't adding anything, and once removed this just becomes a dupe of line 721.

So I think we can remove the condition and just always run lines 720-723 (allowing the condition on line 721 to handle this).

DanTup (approved):
Few minor nits, but LGTM!

PR #4337, opened 2023-01-09
Add support for new DAP toolEvents

### LINKED ISSUE #4193: "Inspect Widget" does not navigate to code in IDE when using new DAPs
The new DAPs don't handle the "inspect" event that was used to navigate the editor to code when inspecting widgets.

Perhaps it's worth considering a more specific event for navigating IDEs, since some users user `inspect()` for wanting to dump variables to the debug console and there's some overlap here (the API also requires a number of calls to get the widget summary which complicates things).

@jacob314 what's involved in creating a new event for this? Should I implement the old way in the new DAPs for now, or is it fairly straight-forward? (I think it would likely want to be a broadcast-style event and not a service extension that the DAP registers, to allow for multiple editors to be connected to the debugger and all receive it?).

DanTup:
- [ ] Update old DAP to also support this so that when the old `inspect` call is removed, non-SDK-DAPs still work.

Fixes #4193.

### DIFF
--- a/src/extension/commands/debug.ts
+++ b/src/extension/commands/debug.ts
@@ -553,6 +553,31 @@ export class DebugCommands implements IAmDisposable {
 			this.debugMetrics.text = message;
 			this.debugMetrics.tooltip = "This is the amount of memory being consumed by your applications heaps (out of what has been allocated).\n\nNote: memory usage shown in debug builds may not be indicative of usage in release builds. Use profile builds for more accurate figures when testing memory usage.";
 			this.debugMetrics.show();
+		} else if (event === "dart.toolEvent") {
+			const kind = body.kind;
+			const data = body.data;
+			switch (kind) {
+				case "navigate":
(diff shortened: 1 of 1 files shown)


jacob314:
Can this be landed now?

DanTup:
Technically, yes - but it won't work until the new DDS is rolled into Flutter (which I think is blocked on some other dependencies like dwds being compatible with vm_service 10). I was waiting for that to be done just so I could do a final test with everything landed (since previous testing was done using dependency_overrides etc.).

DanTup:
Scratch that, DDS was already rolled to the required version (I thought it was blocked based on https://github.com/flutter/flutter/pull/118414).

I'll publishing a release shortly, so I'll do that first, then land this before pushing a pre-release. I don't want to merge this in right before the release (and it will only work for those on `master` anyway, who are hopefully on the pre-release version of the extension).

Issue #4424, opened 2023-03-06
Provide telemetry.json and document what's collected on the website


aitor-gomila:
~~**Is your feature request related to a problem? Please describe.**
Users currently *sometimes* know there *is* telemetry, but they do not know which kind of telemetry.~~

~~**Describe the solution you'd like**
A clear message saying VSCode uses telemetry, what is sent, a consent button and a do not consent button. An optional panel in the titlebar showing the user what information is being sent, when and how. 
E.G: a panel in the Help titlebar button, that when clicked, opens a pane, with a table, showing all the requests and what exactly they send, and how much requests have been sent the last week.~~

~~**Describe alternatives you've considered**
Show the user a summary of all the information they've sent, put together.~~

~~**Additional context**
I understand why telemetry can be useful for something as big as VSCode, but I believe there's 2 keys here missing: consent and clarification.~~

EDIT: use vscode's built-in telemetry API.

DanTup:
We used to have our own flag for enabling analytics, but VS Code requested that we switch to using their flag. My understanding was that they wanted a single opt-in for the whole of VS Code (which you should get the first time you run Code) instead of extensions all popping up their own consent notifications.

The discussion was offline, but there is some info about how extensions handle this here:

- https://code.visualstudio.com/api/extension-guides/telemetry
- https://code.visualstudio.com/api/references/vscode-api#env.isTelemetryEnabled

VS Code updated the telemetry settings to be more explicit that this includes extensions:

<img width="1023" alt="Screenshot 2023-03-06 at 16 45 40" src="https://user-images.githubusercontent.com/1078012/223175660-599bdc13-760c-4260-9f00-9eb0e55b2d79.png">

I'm not sure what is best to do here. We could add our own prompt for Dart, but I feel this is not what VS Code wants. @isidorn do you have any advice here? I certainly don't want to appear to collect data without consent - but if it's not clear from VS Code's prompt that it's enabling for extensions maybe we should?

@aitor-gomila I agree it should be much clearer what is collected (probably a page on dartcode.org and clear links from the readme/etc.). In the meantime you can see anything that's collected in https://github.com/Dart-Code/Dart-Code/blob/5e550c1fda028b951c14aa46f9a4b1c16937c777/src/extension/analytics.ts#L108-L145. It's mostly version numbers and some info about which features are used (or which versions of features you use, such as new DAP vs Legacy). The data is sent to an account that only _I_ have access to.

[...]

PR #4425, opened 2023-03-06
Use built-in VSCode API for Telemetry

aitor-gomila:
@DanTup moving #4424 over here.

Using VSCode's built-in logger will be a more concise approach. ~~However, it requires the use of Azure [Application Insights](https://azure.microsoft.com/en-us/products/monitor) for logging instead of the Google-based approach we're using right now.~~

~~I don't like how this API was made - I believe it should be more flexible and Azure should be detached from VSCode.~~

~~However it would be neat to just use this API. What do you think? Should I keep working on it? Is this the approach you/Google/whoever wants?~~

Thanks :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -2320,7 +2320,7 @@
 					"dart.vmAdditionalArgs": {
 						"type": "array",
 						"default": [],
-						"markdownDescription": "Arguments to be passed to the Dart VM when running Dart CLI scripts.\n\nThese arguments appear between \"dart\" and \"run\":\n\n`dart (vmAdditionalArgs) run (toolArgs) bin/main.dart (args)`",
+						"description": "Additional args to pass to the Dart VM when running/debugging command line apps or Dart test scripts.",
 						"scope": "resource",
 						"items": {
 							"type": "string"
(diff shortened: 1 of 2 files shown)


DanTup:
> Using VSCode's built-in logger will be a more concise approach. However, it requires the use of Azure [Application Insights](https://github.com/Dart-Code/Dart-Code/pull/4425) instead of the Google-based approach we're using right now.

My understanding is that we can implement a `TelemetryLogger` to send to Google Analytics and do not need to use Azure Application Insights:

https://code.visualstudio.com/api/references/vscode-api#TelemetryLogger

DanTup:
@aitor-gomila the code is already there in the `analytics.ts` file (the code you removed in this PR :-) ). I think it just needs relocating into a `TelemetrySender` class instead (I incorrectly wrote `TelemetryLogger` above where I meant `TelemtrySender`), and then the implementations of the `analytics.logFoo()` methods should become calls to a `TelemteryLogger`.

So generally I think the work is:

- Create a class that extends `TelemetrySender` that uses the code removed in this PR to send the data to Google Analytics (note: we should preserve the existing `debug` flag that lets you send data to the debug endpoint for validation)
- Update where we create the `Analytics` class to instead first create the sender, then call `createTelemetryLogger()`, then pass that logger into the `Analytics` class
- Update the `logXxx` methods on the `Analytics` class to call the `TelemetryLogger`

aitor-gomila:
I removed that code in order to be able to send a PR, restoring it back :)

Ok, that makes a lot of sense. Thanks for clarifying.

aitor-gomila:
Sorry, moving it to another branch. it was getting dirty up there

PR #4444, opened 2023-03-16
Fix bazel tests not runnable from triangles

helin24:
There's a recently added check for whether tests are pub runnable that caused bazel tests to not be runnable from clicking the triangles, so I added a check of `forceFlutterWorkspace`

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -545,7 +545,7 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean 
 	if (testDiscoverer)
 		context.subscriptions.push(testDiscoverer);
 	const vsCodeTestController = vs.tests?.createTestController !== undefined // Feature-detect for Theia
-		? new VsCodeTestController(logger, testModel, testDiscoverer, workspaceContext.config.forceFlutterWorkspace)
+		? new VsCodeTestController(logger, testModel, testDiscoverer)
 		: undefined;
 	if (vsCodeTestController)
 		context.subscriptions.push(vsCodeTestController);
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -545,7 +545,7 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean 
 	if (testDiscoverer)
 		context.subscriptions.push(testDiscoverer);
 	const vsCodeTestController = vs.tests?.createTestController !== undefined // Feature-detect for Theia
-		? new VsCodeTestController(logger, testModel, testDiscoverer)
+		? new VsCodeTestController(logger, testModel, testDiscoverer, workspaceContext.config.forceFlutterWorkspace)
 		: undefined;
 	if (vsCodeTestController)
 		context.subscriptions.push(vsCodeTestController);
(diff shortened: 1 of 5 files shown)


src/extension/extension.ts
```diff
@@ -545,7 +545,7 @@ export async function activate(context: vs.ExtensionContext, isRestart: boolean
 	if (testDiscoverer)
 		context.subscriptions.push(testDiscoverer);
 	const vsCodeTestController = vs.tests?.createTestController !== undefined // Feature-detect for Theia
-		? new VsCodeTestController(logger, testModel, testDiscoverer)
+		? new VsCodeTestController(logger, testModel, testDiscoverer, workspaceContext.config.forceFlutterWorkspace)
```
DanTup, line 548: I'd consider adding another flag (that's just also set in the same place as this one) that's descriptive specifically for this purpose. That way if the capabilities every change internally (or we want to reuse this for other types of workspaces), it's a bit simpler.

We actually have another flag somewhere that is the opposite of this (which disables it for the Dart VM). Perhaps if we add a specific flag, I could change that to use this (eg. it's `true` here, `false` for the Dart SDK, and `undefined` when we should check for pubspec).

I'm not sure exactly how to describe what it's doing - "pub run test" is a bit misleading here (since it's actually "dart test" and "flutter test" now).. Perhaps `supportsPackageTest` would be reasonable (both `dart test` and `flutter test` use `package:test`, which is what is providing the ability to run specific tests).

src/extension/test/vs_test_controller.ts
```diff
@@ -18,7 +18,7 @@ export class VsCodeTestController implements TestEventListener, IAmDisposable {
 	private nodeForItem = new WeakMap<vs.TestItem, TreeNode>();
 	private testRuns: { [key: string]: { run: vs.TestRun, shouldEndWithSession: boolean } | undefined } = {};
 
-	constructor(private readonly logger: Logger, private readonly model: TestModel, private readonly discoverer: TestDiscoverer | undefined) {
+	constructor(private readonly logger: Logger, private readonly model: TestModel, private readonly discoverer: TestDiscoverer | undefined, private readonly isForceFlutterWorkspace: boolean | undefined) {
```
DanTup, line 21: As above, maybe name something like `supportsPackageTest`.

src/extension/test/vs_test_controller.ts
```diff
@@ -286,7 +286,7 @@ export class VsCodeTestController implements TestEventListener, IAmDisposable {
 			return false;
 		if ((label.startsWith("(setUp") || label.startsWith("(tearDown")) && label.endsWith(")"))
 			return false;
-		if (this.discoverer?.fileTracker.supportsPubRunTest(node.suiteData.path) === false)
+		if (this.discoverer?.fileTracker.supportsPubRunTest(node.suiteData.path) === false && !this.isForceFlutterWorkspace)
```
DanTup, line 289: I think there may be other places calling `supportsPubRunTest` that want the same behaviour as this (such as in the Test CodeLens handler, which adds the Run/Debug links above the Group/Test methods?).

DanTup (approved):

src/extension/test/vs_test_controller.ts
helin24: Ah it looks like `supportsPubRunTest` checks `useVmForTests` from `workspaceConfig`, so I should be able to just add a flag and use it in there as well.

src/extension/extension.ts
helin24: Thanks for the name suggestion! I'll leave it up to you if you want to change other spots to using this.

DanTup:
Thanks - I dropped `useVmForTests` in adf980a6637de5ec424335347e2e56e57384d9b8 and just piggy-back'd on this flag (making `false` mean an explicit no, and `undefined` mean it requires a pubspec). Should just be a no-op, but lmk if anything doesn't look right.

helin24:
LGTM!

Issue #4250, opened 2022-11-06
Debugging flutter tests fail if the descrription string has an interpolated value and a is composed of two concatenated strings


WorikQCI:
**Describe the bug**
If the description of the test has an interpolated value and a is composed of two concatenated strings the test will not run

**To Reproduce**

Write a test like the following and try to run it by clicking the debug option 

![image](https://user-images.githubusercontent.com/95715668/200192354-9cb12fbd-05d4-4108-871d-6f1f2c294c73.png)

```
import 'package:test/test.dart';

void main() async {
  //
  int foo = 1;
  test("Start string $foo " "  end string", () async {
    expect(true, true);
  });
}
```

Get this result:

```
Connecting to VM Service at http://127.0.0.1:64510/Y0ktArhJg9g=/ws
No tests match regular expression "^Start string .* " "  end string( \(variant: .*\))?$".
Exited (1)
```
**Expected behavior**

I expect the test to run.

**Please complete the following information:**
 - Operating System and version:
   MacOS 12.6
 - VS Code version:
 
```
Version: 1.72.2 (Universal)
Commit: d045a5eda657f4d7b676dedbfa7aab8207f8a075
Date: 2022-10-12T22:16:30.254Z
Electron: 19.0.17
Chromium: 102.0.5005.167
Node.js: 16.14.2
V8: 10.2.154.15-electron.0
OS: Darwin arm64 21.6.0
Sandboxed: No
```

 - Dart extension version:
v3.52.1

 - Dart/Flutter SDK version:
`Dart SDK version: 2.18.4 (stable) (Tue Nov 1 15:15:07 2022 +0000) on "macos_arm64"`
 - Target device (if the issue relates to Flutter debugging):

DanTup:
Merging this into #4021 since that should fix a number of issues with non-trivial test names like this one.

DanTup:
Re-opening this for tracking to ensure this issue is fixed by the upcoming support for running tests by lines.

DanTup:
In the next release (and the pre-release version just pushed today - v3.63.20240419), you can choose to have tests executed using their line numbers rather than their names. This fixes a number of issues including this one, caused by not always being able to statically compute the exact name of a test. 

![Test Invocation settings](https://user-images.githubusercontent.com/1078012/233161136-294c72e9-5a28-489b-bfc0-cd786636233c.png)

Please let me know if you find any issues with this functionality!

WorikQCI:
[...]

Issue #4168, opened 2022-09-21
Parens around test name prevents CodeLens links from working


DanTup:
Noted at https://github.com/Dart-Code/Dart-Code/issues/1458#issuecomment-1253608986 by @nilsreichardt.

```dart
test('foo', () {});
```

> No tests match regular expression "^\('foo'\)( \(variant: .*\))?$".

Opening this as a placeholder/reminder to ensure that once #4021 is implemented that it covers this case (which might mean that we just don't pass any test name with the line/col).

DanTup:
I thought this might be easy to fix without the line numbers, but it turns out the server returns the same names for these tests:

```dart
test(("test 6"), () => expect(1, equals(1)));
test('("test 6")', () => expect(1, equals(1)));
```

```
"name": "test(\"(\"test 6\")\")",
"name": "test(\"(\"test 6\")\")",
```

So we'll have to wait for line numbers.

DanTup:
In the next release (and the pre-release version just pushed today - v3.63.20240419), you can choose to have tests executed using their line numbers rather than their names. This fixes a number of issues including this one, caused by not always being able to statically compute the exact name of a test. 

![Test Invocation settings](https://user-images.githubusercontent.com/1078012/233161136-294c72e9-5a28-489b-bfc0-cd786636233c.png)

Please let me know if you find any issues with this functionality!

Issue #4150, opened 2022-09-07
Multiple tests execution


grafovdenis:
**Describe the bug**
Once you're trying to run test, which description contains variable or type used in other test, multiple tests would be executed.
This could be annoying if you're using some kind of BLoC/Redux actions, which types may be located in multiple test descriptions.

Simular issue in `/test` repo: https://github.com/dart-lang/test/issues/1760.
There we found out, that the plugin [could interprete](https://github.com/dart-lang/test/issues/1760#issuecomment-1238381218) expressions like `'is $Object'` incorrectly.

**To Reproduce**
Run `'is $Object'` test.

```dart
import 'package:test/test.dart';

class MyObj {
  final number = 10;

  bool foo() => true;
}

void main() {
  group('$MyObj', () {
    late MyObj myObj;

    setUp(() {
      myObj = MyObj();
    });

    test('is $Object', () {
      expect(myObj, isA<Object>());
    });

    test('foo returns true and is an $Object', () {
      expect(myObj.foo(), isTrue);

      expect(myObj, isA<Object>());
    });

    test('number equals 10', () {
      expect(myObj.number, equals(10));
    });
  });
}
```

**Expected behavior**
Running `'is $Object'` test, via IDE would execute only selected test.

**Screenshots**
<img width="645" alt="image" src="https://user-images.githubusercontent.com/20505376/188862970-fbac166c-df2f-4e71-8e37-88bd4f8bf70a.png">
It seems like plugin runs test like this.
<img width="785" alt="image" src="https://user-images.githubusercontent.com/20505376/188863524-d2c4222c-92de-48fd-9734-32950c659e4e.png">
Where `$Object` is an empty env. Which means, that we try to run test with `'is '` description, which also suits for the second test.

**Please complete the following information:**
 - Operating System and version: macOS 12.3.1 (21E258)
 - VS Code version: 1.71.0
 - Dart extension version: v3.48.2
 - Dart/Flutter SDK version: 2.17.3

DanTup:
There are some known limitations in this area. When we run tests, we cannot use the literal name in the source if it contains variables:

```
% dart test test/dartcode4150_test.dart --plain-name "^\$MyObj is \$Object$"
No tests ran.                                                                                                                                                                                
No tests match "^$MyObj is $Object$".
```

We also cannot tell what `$MyObj` or `$Object` are to substitute their values, since they could be dynamic:

```dart
  final foo = DateTime.now().millisecond;
[...]

Issue #4099, opened 2022-08-10
"No tests match regular expression...", customize regex


jibbers42:
I get the "No tests match regular expression..." error when using [Glados](https://github.com/MarcelGarus/glados) (property-based testing) and attempting to run a test via test code lens.

It seems they modify the name of the test [here](https://github.com/MarcelGarus/glados/blob/503180fc8c1051b7c9bdccb3bd14d7d95135722d/glados/lib/src/glados.dart#L167).

Is there a way for me to modify the regex used to account for their modification, or another solution?
(I'm using Dart SDK without Flutter)

Thanks!

DanTup:
Sorry for the delay, I somehow missed this issue. I think the best solution is probably #4021, which avoids us having to "predict" the names of tests when running (instead, we will run them based on the line number they're on). I haven't had chance to try this out yet, but I'll keep this issue open as a reminder to test this once that work is done.

DanTup:
In the next release (and the pre-release version just pushed today - v3.63.20240419), you can choose to have tests executed using their line numbers rather than their names. This fixes a number of issues including this one, caused by not always being able to statically compute the exact name of a test. 

![Test Invocation settings](https://user-images.githubusercontent.com/1078012/233161136-294c72e9-5a28-489b-bfc0-cd786636233c.png)

Please let me know if you find any issues with this functionality!

jibbers42:
Thank you!

PR #4487, opened 2023-04-05
Add support for running tests by line number instead of name

DanTup:
See https://github.com/Dart-Code/Dart-Code/issues/4021.

- [x] Address `// TODO(dantup): This is right for Flutter, but must check package:test version for Dart too!`
- [x] Remove `const runTestsByLine = true`
- [x] Address `// TODO(dantup): Ensure we don't get here for Dart SDK or Bazel where we shouldn't use "dart run"?`
- [x] Address `// TODO(dantup): WIRE THIS UP!`
- [x] Review if `findMatchingDynamicNode` should be taking lines into account?
- [x] Ensure there are tests that cover multiple tests in one run (`"program": "foo_test.dart?line=1", "args": ["foo_test.dart?line=20"]`)
  - [x] manually tested
- [x] Test all issues noted/referenced in https://github.com/Dart-Code/Dart-Code/issues/4021
- [x] Address `// TODO(dantup): Ensure this never runs in the SDK, because it may produce unwanted package_configs?`
- [x] Revert setting the default to "line" that was used for testing on the bots (c00add9e)
- [x] Ensure Flutter has automated tests in the same way as Dart
  - [x] manually tested

Fixes several issues:

- Fixes https://github.com/Dart-Code/Dart-Code/issues/4021
- Fixes https://github.com/Dart-Code/Dart-Code/issues/4150
- Fixes https://github.com/Dart-Code/Dart-Code/issues/4168
- Fixes https://github.com/Dart-Code/Dart-Code/issues/4099
- Fixes https://github.com/Dart-Code/Dart-Code/issues/4250


src/shared/utils/workspace.ts
```diff
@@ -86,6 +87,10 @@ export function tryProcessBazelFlutterConfig(logger: Logger, config: WritableWor
 		config.localDeviceCommandAdviceMessage = flutterConfig.localDeviceCommandAdviceMessage;
 		config.localMacWarningMessage = flutterConfig.localMacWarningMessage;
 		config.supportsPackageTest = true;
+		// TODO(dantup): Check with helin24 whether this should be enabled. Running
+		//  "dart run test:test --version" is how we check pkg:test capabilities (such
+		//  as whether we can run tests by line number).
+		config.supportsDartRunTest = false;
```
DanTup, line 93: @helin24 we added this `supportsPackageTest` flag for Bazel workspaces to tell Dart-Code that it can run tests using `--name` args even though there's no pubspec and we can't verify `package:test` is available.

I've added another flag to check whether it's safe to run `dart run test:test --version` (because we cannot do this in the SDK, for example) when trying to check package:test capabilities.

Can you confirm whether it's valid (and safe) to run internally? If not, we should leave this flag as `false`, but otherwise it can be `true`. Having `true` is required to detect some new features of `package:test` like being able to run tests using their line numbers (`dart test "foo_test.dart?line=10"`) instead of names to avoid some issues with dynamic/duplicated names.

DanTup: (I'm going to land this because I'd like to publish a pre-release version.. this question isn't urgent, as the default is safe - but we can tweak if it if turns out we can do `dart run test:test` here.. if we can't, then version-specific `package:test` functionality like running by line numbers will be disabled, and we can look at another way of doing this if we want to support that)

helin24: This won't work internally yet - we'd probably need to modify the internal code to parse a line number from similar syntax and/or create a different launch command for internal situations. I'm not sure where this would fall on our priorities to implement; were we thinking of removing the ability to start running tests with text? (I think we had a doc on this and now can't find it..)

DanTup: No plans to remove running tests by name. Depending on how it goes, I might flip the default to using lines in future (to avoid things not working sometimes in ways that aren't obvious to users), but we'd only use it where we know it's supported.

For now I'll just replace the TODO with a comment that this isn't supported, and if at any point it becomes useful to have internally, we can look at changing it then. Thanks!

DanTup: I replaced the TODO with some notes in 32b22cfae095ca49db4088edca8be7f22fc3d0d4. Thanks!

PR #4494, opened 2023-04-15
Round corner for flutter guides

pouyakary:
Today I finally discovered that these flutter guides are a beta preview ! and so I enabled them and thought, why not having round corner? So this is the tiniest PR possible :) making only a simple change that makes these guides a little bit prettier hopefully...


DanTup:
Thanks!

pouyakary:
You're very welcome!

PR #4588, opened 2023-06-16
[package.json] respect users editor.tabSize and related options

chipweinberger:
Dart-Code needs to be very conservative when overriding user settings, as they *take precedence* over user settings. See https://github.com/Microsoft/vscode/issues/41542

> Here are some reasons why editor.tabSize might be not working for you:
> 
> You have installed an extension that uses VS Code API to tweak those settings at runtime. e.g. editorconfig extension. Try with disabling all extensions.

Without this change, a user must change the tabSize every time they open a Dart file. 

Considering we already have Dart Format, I see very little reason for overriding these.

Our organization has for 2 decades been standardized on a tabSize of 4 for _all_ programming languages, so this is very tedious.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1362,9 +1362,6 @@
 		},
 		"configurationDefaults": {
 			"[dart]": {
-				"editor.tabSize": 2,
-				"editor.insertSpaces": true,
-				"editor.detectIndentation": false,
 				"editor.suggest.insertMode": "replace",
 				"editor.defaultFormatter": "Dart-Code.dart-code",
 				"editor.inlayHints.enabled": "offUnlessPressed"


DanTup:
Unfortunately this change will negatively impact a lot of users because the default settings will no longer match the formatter. What we're doing here is the way VS Code excepts per-language defaults to be set.

You can override the settings, but you must do it in the Dart-specific settings for VS Code for it to apply (you can do this either in your User or Workspace settings):

```js
"[dart]": {
	"editor.tabSize": 4,
	"editor.insertSpaces": false,
}
```

You can also get to this via the **Preferences: Configure Language Specific Settings** command.

I'm not sure VS Code does a great job of making how this works clear to users, but it was supposed to have been improved in https://github.com/microsoft/vscode/issues/131707. Unfortunately there isn't currently a better way to do this that wouldn't make the experience worse for a significant number of users.

chipweinberger:
fair enough!

PR #4814, opened 2023-10-26
Add listener for copy events above the devtools frame

CoderDake:
This is to overcome [a bug with VSCode that isn't allowing Devtools to use the copy api.](https://github.com/flutter/devtools/issues/5775)

I've added the allow permissions to the iframe so that if VSCode ever fixes this issue, we should be able to just start utilizing the original copy method right away :)

The complementary code from Devtools can be found in this PR: https://github.com/flutter/devtools/pull/6598


DanTup:
LGTM, thanks!

DanTup:
I've invited you to Dart-Code, feel free to merge this when you're ready.

Issue #4705, opened 2023-08-24
Add logic for VS code to promote other VS code extensions 


kenzieschmoll:
As part of the broader effort to support dart ecosystem extensions, we've designed a mechanism for a pub package to provide extension information as part of its package. For DevTools, this is the precompiled flutter web app that DevTools will embed. For VS Code extensions, the pub package can include a `config.yaml` file that tells the Dart-Code promotion which extension should be promoted:
```
my_package/
   lib/
      ...
   extension/
      vs_code/
          # contains meta data that VS code needs to point to another
          # extension on the VS code marketplace
         config.yaml
```

[package:extension_discovery](https://pub.dev/packages/extension_discovery) has logic to detect available extensions for a given `.dart_tool/package_config.json` file. Since this package is written in Dart, it cannot be used by the Dart-Code extension directly, so it will need to be called into from the Dart-Code extension by proxy of some other Dart service. Eventually, this would be a great fit for the [Dart Tooling Daemon](https://docs.google.com/document/d/1qJNwgg1e_4KzY_fSU3rmSj9CXqtpGC5GmQb0cXkUGfo/edit), but since this does not exist yet, we can use the DevTools server in the interim.

We can add an API to DevTools server that the Dart-Code extension can call, and the devtools_server will call `findExtensions('vs_code', <path/to/package_config.json>)` from [package:extension_discovery](https://pub.dev/packages/extension_discovery) on the Dart-Code extension's behalf.

**Open Question**
The element we still need to figure out is how to handle a monorepo workspace, where there are multiple `package_config.json` files available. 

Related: [Flutter DevTools Extensions](https://docs.google.com/document/d/189A-T-pcHPuds8pOtbzLS8Wl6Hr0AIAIl0aRTPvLKec/edit)

DanTup:
@anderdobo @mosuem FYI - I wonder whether there's value in trying to use this for arb-editor (eg. have the Flutter/intl packages provide the recommendation)? It might need something extra (only show the recommendation if an arb file is opened?), but it might work out better than special handling for one extension?

mosuem:
To recap an offline discussion: We will go with implementing the suggestion the extension itself for now, see https://github.com/Dart-Code/Dart-Code/issues/4719. We will possibly migrate to using this new feature. At the moment, the lack of a file filter ("only show the recommendation on opening `.arb` files") would result in too much noise for Flutter users.

DanTup:
[...]

PR #4847, opened 2023-11-20
Add support for DevTools-server-driven extension recommendations

DanTup:
Adds support for asking DevTools server for any VS Code extension recommendations (which are ready from packages that your workspace depends on).

- [x] Needs tests
- [x] Needs SDK work to land
- [x] Need to verify version number in capabilities
- [x] Need to confirm exact text(/disclaimer?)
- [x] Needs an initial whitelist as part of rollout (empty, can add to later)

image

image

Fixes https://github.com/Dart-Code/Dart-Code/issues/4705


kenzieschmoll:
FYI @vlidholt.

DanTup:
I've added some tests and some settings so user can a) turn extension recommendations off entirely and b) replace the allow-list so that authors of packages/extensions could verify their `extension/vs_code` config locally before being in the list.

This is essentially ready to merge (with an empty allow-list), but due to the recent issues with paths, I think we should gate it on an SDK that has `devtools_shared` v6.0.3 that has the fixes.

@kenzieschmoll how is that version bumped in DDS? Is it coming direct from Pub and not via DEPS in the SDK? Do you already have a plan for when to increase it there?

If we're not ready yet, I could merge this with a very high SDK version and we can move it down once we know which Dart SDK we're happy for it to start being used for.

kenzieschmoll:
> This is essentially ready to merge (with an empty allow-list), but due to the recent issues with paths, I think we should gate it on an SDK that has devtools_shared v6.0.3 that has the fixes. How is that version bumped in DDS? Is it coming direct from Pub and not via DEPS in the SDK? Do you already have a plan for when to increase it there?

I rolled the latest DevTools into the Dart SDK yesterday (including your post message fix for the flutter sidebar). And since the DDS dep on devtools_shared is `^6.0.0`, it should already be picking up `6.0.3`.

DanTup:
I was a bit concerned by not having 6.0.3 in the pubspec to enforce this, but I see you've landed https://github.com/dart-lang/sdk/commit/a86eb8bd113b77b07f588d8772ca000897807c33 now so looks like it's safe to use 3.3.0-247 - thanks!

Issue #4925, opened 2024-01-05
Syntax Highlighting for Markdown Documentation Comments


Number-3434:
Please add syntax highlighting for documentation comments.

For example, the following:

```dart
/// ## Does something useful, involving [value]
///
/// ### Reason for This Method
/// This method is purely intended for demonstrative purposes.
/// Note the lack of any body, and the use of `...`
/// 
/// ### Here's a [table](https://www.markdownguide.org/extended-syntax/#tables):
/// 
/// | Nice Table | Value |
/// |   :---:    | :---: |
/// | Some Text  |  foo  |
/// | More Text  |  bar  |
/// 
/// ### ...And a [definition list](https://www.markdownguide.org/extended-syntax/#definition-lists):
/// 
/// First Term
/// : This is the definition of the first term.
/// 
/// Second Term
/// : This is one definition of the second term.
/// : This is another definition of the second term.
/// 
/// **See also:**
/// - [num], some useful points here
///   - [int], a child of [num]
///   - [double], also a child of [num]
/// - [bool], a basic boolean
void usefulMethod(int value) { ... }
```

...should be coloured like:

```md
## Does something useful, involving [value]

### Reason for This Method
This method is purely intended for demonstrative purposes.
Note the lack of any body, and the use of `...`

### Here's a [table](https://www.markdownguide.org/extended-syntax/#tables):

| Nice Table | Value |
|   :---:    | :---: |
| Some Text  |  foo  |
| More Text  |  bar  |

### ...And a [definition list](https://www.markdownguide.org/extended-syntax/#definition-lists):

First Term
: This is the definition of the first term.

Second Term
: This is one definition of the second term.
: This is another definition of the second term.

**See also:**
- [num], some useful points here
  - [int], a child of [num]
  - [double], also a child of [num]
- [bool], a basic boolean
```
...but using the local syntax highlighting rules.
Rendered as:

### Reason for This Method
This method is purely intended for demonstrative purposes.
Note the lack of any body, and the use of `...`

### Here's a [table](https://www.markdownguide.org/extended-syntax/#tables):

| Nice Table | Value |
|   :---:    | :---: |
| Some Text  |  foo  |
| More Text  |  bar  |

### ...And a [definition list](https://www.markdownguide.org/extended-syntax/#definition-lists):

First Term
: This is the definition of the first term.

Second Term
: This is one definition of the second term.
: This is another definition of the second term.

**See also:**
- [num], some useful points here
[...]

PR #4999, opened 2024-02-21
Add Markdown support to Documentation Comments

RedCMD:
Fix https://github.com/Dart-Code/Dart-Code/issues/4925

image

Simply just `#includes` Markdown

using `meta.embedded.markdown` with `"embeddedLanguages": { "meta.embedded.markdown": "markdown" }`
enables basic language feature support
like `blockComment` and `autoClosingPairs`

### DIFF
--- a/package.json
+++ b/package.json
@@ -95,6 +95,9 @@
 				"path": "./syntaxes/dart.json",
 				"tokenTypes": {
 					"string.interpolated.expression.dart": "other"
+				},
+				"embeddedLanguages": {
+					"meta.embedded.markdown": "markdown"
 				}
 			}
 		],
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -158,7 +158,7 @@
 		"comments-doc": {
 			"patterns": [
 				{
-					"name": "comment.block.documentation.dart",
+					"name": "meta.embedded.markdown comment.block.documentation.dart",
 					"begin": "///",
 					"while": "^\\s*///",
 					"patterns": [
--- a/syntaxes/dart.json
+++ b/syntaxes/dart.json
@@ -72,6 +72,9 @@
 		"dartdoc": {
 			"patterns": [
 				{
+					"include": "text.html.markdown"
+				},
+				{
 					"match": "(\\[.*?\\])",
 					"captures": {
 						"0": {


Number-3434:
Thank you!

RedCMD:
How to enable semantic highlighting?
Can't find any references to it anywhere

VSCode has been setup in a way that semantic highlighting should only ever enchance the TextMate grammars
by highlighting things that TextMate just cant (context aware syntaxes etc)
not completly override the entire document

Number-3434:
@RedCMD It's in `settings.json` - <samp>editor.semanticTo</samp>

RedCMD:
I have that setting enabled already

but even when searching through the code in this repo
I cannot find any reference to `SemanticTokensProvider`

as far as I aware, this extension doesn't provide any semantic highlighting

Number-3434:
Do you see any difference when turning semantic tokens on and off?

DanTup:
@RedCMD 

> VSCode has been setup in a way that semantic highlighting should only ever enchance the TextMate grammars
> by highlighting things that TextMate just cant (context aware syntaxes etc)
> not completly override the entire document

Unfortunately it's not that simple. The LSP server (which is generic and not VS Code-specific) is unaware of what textmate grammar the client may have (if it even has one at all) and therefore provides a full set of semantic highlighting tokens. In VS Code, these overwrite anything from the textmate grammar making it essentially disappear. https://github.com/microsoft/vscode/issues/163292 is a request specifically to support what is necessary to handle this without having to add a lot of non-standard / Code-specific implementation to the LSP server.


> but even when searching through the code in this repo
> I cannot find any reference to SemanticTokensProvider
> 
> as far as I can see, this extension doesn't provide any semantic highlighting

Language functionality comes from the LSP server in the SDK (while there are some providers in this repo, they are only for the legacy protocol which is being phased out). The LSP handler for the semantic tokens is here:

https://github.com/dart-lang/sdk/blob/main/pkg/analysis_server/lib/src/lsp/handlers/handler_semantic_tokens.dart

DanTup:
Isn't that essentially the same request as https://github.com/microsoft/vscode/issues/163292?

Number-3434:
It's very similar, except that my version also allows users and themes to declare the overrides, not in the textmate grammar itself.

I would prefer this option for the customisability aspect.

PR #5040, opened 2024-03-19
Use VM service URI directly for force flutter cases

helin24:
This fixes the problem of DevTools not being able to access a local instance of DDS when the app sources are on a remote machine and Dart-Code is running in the remote extension host.

In the remote extension case for g3 apps, DevTools will be opened on the local machine, and the VM service URI will also be accessible on the local machine. So we don't want to do forwarding of the port; otherwise the result will be available on the remote side instead of local.


src/extension/sdk/dev_tools/manager.ts
```diff
@@ -325,13 +325,13 @@ export class DevToolsManager implements vs.Disposable {
 		// to only use that for a title, we may be able to relax that.
 		if (options.location !== "external") {
 			queryParams.embed = "true";
-			const fullUrl = await this.buildDevToolsUrl(url, queryParams, vmServiceUri);
+			const fullUrl = await this.buildDevToolsUrl(url, queryParams, vmServiceUri, session?.clientVmServiceUri);
```
DanTup, line 328: I would pull this out to a variable (and reuse below) with the same condition as `vmServiceUri` on line 323 so that we don't need to worry about `buildDevToolsUrl` accidentally using it for static tools.

We could actually flatten the implementation in `buildDevToolsUrl` a little then too (removing the outer `if` statement):

```
if (clientVmServiceUri) {
	queryParams.uri = clientVmServiceUri;
} else if (vmServiceUri) {
	const exposedUrl = await envUtils.exposeUrl(vmServiceUri, this.logger);
	queryParams.uri = exposedUrl;
}
```

DanTup (approved):
LGTM with some minor tweaks.

We didn't have any tests specifically checking this, so I've added one here:

https://github.com/Dart-Code/Dart-Code/commit/1312c96ba4d26463d3a02bd3d2b5de276451e02f

If you rebase on master you should be able to add an additional tests here that includes `clientVmServiceUri` in the `debuggerUris` event and then verifies that one is used in the URI instead.

(You can run just those tests by changing `describe` to `describe.only` and then using the "Dart Tests + DAs" launch configuration from the VS Code debug sidebar)

src/extension/utils/vscode/debug.ts
```diff
@@ -6,6 +6,7 @@ import { PromiseCompleter } from "../../../shared/utils";
 export class DartDebugSessionInformation {
 	public observatoryUri?: string;
 	public vmServiceUri?: string;
+	public clientVmServiceUri?: string;
```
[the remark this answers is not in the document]
helin24: thanks, modified a bit

src/extension/sdk/dev_tools/manager.ts
helin24: I ended up not using the additional client URI because I had them backwards - the VM service URI is already accessible on mac, but asking for `exposeUrl` makes it only accessible on linux.

helin24:
@DanTup I misunderstood which URL was accessible on which side when we last talked, so we should actually be using the VM service URI. However, now I'm checking `forceFlutterWorkspace` to decide whether to skip the port forwarding. Is that a reasonable context var to choose?

In this case testing seems a bit trickier, so let me know if you tips here. Is there an example of setting workspace context in a test environment?

src/extension/sdk/dev_tools/manager.ts
DanTup: Hmmm, this sounds a little unsound to me. There are many places that we use this VM Service URI, and only DevTools client is running on the Mac. So if the VM Service URI we get is the raw Mac one (and not accessible from Linux where the extension host is), we may pass it to things on Linux that won't be able to connect (for example we may [give it to DevTools server](https://github.com/Dart-Code/Dart-Code/blob/7a0efbc0ade74e3d617e43b483b96a143705fea4/src/extension/sdk/dev_tools/manager.ts#L464)).

I feel like it may be better if we still had access to both URIs in Dart-Code so we can use the appropriate URI for each use case?

DanTup:
(@helin24 see my comment in the code above)

src/extension/sdk/dev_tools/manager.ts
```diff
@@ -376,8 +376,24 @@ export class DevToolsManager implements vs.Disposable {
...
+			 * the end user machine (eg. Mac) while the extension host is an SSH remote
+			 * (eg. Linux).
+			 *
+			 * `clientVmServiceUri` indicates a URI that is already accessible on the end
+			 * user machine without forwarding. `vmServiceUri` indicates a URI that is
+			 * accessible to the extension host.
+			 *
+			 * If a `clientVmServiceUri` exists, use it directly instead of trying to
+			 * forward a URI from the extension host.
+			 */
```
DanTup, line 390: I would put this comment (at least in part) on the `DartDebugSessionInformation` interface so it'll be visible from other code using these projects.

src/test/dart/devtools/devtools.test.ts
```diff
@@ -50,5 +50,32 @@ describe("DevTools", async () => {
...
+		await vs.commands.executeCommand("dart.openDevTools.external");
+
+		// Verify an attempt was made to open the correct URI.
+		assert.equal(openBrowserCommand.calledOnce, true);
+		const openedUri = openBrowserCommand.args[0][0] as string; // First invocation, first arg.
+		assert.ok(
+			openedUri.includes(debuggerUrisEvent.clientVmServiceUri),
+			`URI did not contain expected VM Service!\nOpened: ${openedUri}\nExpected to contain: ${debuggerUrisEvent.clientVmServiceUri}`,
+		);
+	});
```
DanTup, line 79: I think this should use `assertDevToolsUriWithVmService`. Is there an issue with the encoding when you used it? If so, we should understand whether the encoding is incorrect, or if the other URL is just over-encoded or something and try and make them consistent.

helin24: Oops sorry, I got confused and thought this was simulating the port forwarding from `exposeUrl`.

DanTup (approved):

PR #5175, opened 2024-07-05
Feature: Add command to get packages from all projects

rodrigogmdias:
This Pull Request adds a new command to the DartCode extension for Visual Studio Code, enabling users to `run pub get` for *all Dart projects* within the workspace. This feature streamlines the process of fetching dependencies for multiple projects, enhancing productivity and efficiency.

# Changes
- Added a new command dart.getAllPubGet to execute pub get for all projects.
- Updated the command palette to include the new command.
- Modified the extension activation logic to register the new command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -191,7 +191,7 @@
 			},
 			{
 				"command": "pub.get.all",
-				"title": "Get Packages for all Projects",
+				"title": "Get Packages for All Projects",
 				"category": "Pub",
 				"icon": {
 					"light": "./media/commands/get.svg",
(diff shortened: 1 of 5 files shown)


MarcioFPaludo (approved):

DanTup:
> This feature streamlines the process of fetching dependencies for multiple projects, enhancing productivity and efficiency.

btw, the extension should detect when this is required when you open the project, and it should watch Pubspec for changes to run again.

I don't think having this command is a problem, but if it seems like you need to use this command frequently, there may be something not working correctly (unless you have deliberately turned off those other features).

rodrigogmdias:
Thx and Perfect.
 
I work in monorepo project of a big project. Some times i need run pub get in all because I add a folder with many submodules in my workspace or bacause I change flutter version and sometimes in git pull not working on opening project.
 
For this, I resolved create manual command. 😃

DanTup:
Gotcha, this makes perfect sense then, thanks!

This change will be included in the next release (releases are usually at the start of the month), and may show up in a pre-release version before then (todays pre-release was unfortunately before I merged this so it's not included in that).

Issue #5154, opened 2024-06-25
Workspace name/folder as macro for log


FMorschel:
**Is your feature request related to a problem? Please describe.**
I received instructions to set the extension to save log files. E.g.:

```json
"dart.flutterRunLogFile": "C:\\Dev\\Logs\\flutter_run.txt",
"dart.vmServiceLogFile": "C:\\Dev\\Logs\\vm_service.txt",
"dart.dapLogFile": "C:\\Dev\\Logs\\dap.txt",
```

When looking for these in VSCode settings, I found out about `${name}` being supported inside the path as a macro for the debug session name.

However, I was also instructed to set these settings for each workspace so that different workspaces don't conflict with each other. Which is a little too verbose when having lots of projects that you'd like to log.

**Describe the solution you'd like**
I'd like for the extension to support `${folder}` or `${workspace}` or similar so that I can create a path like `~/dev/logs/${workspace}/${name}`.

**Describe alternatives you've considered**
Manually creating an inner folder for each workspace and defining it to the workspace settings.

**Additional context**
I was also reminded that some workspaces in VSCode can be unnamed. So in this case _I believe_ the files could be simply left under `~/dev/logs/${name}` as for my example.

DanTup:
In the next version you can use `${workspaceName}` and we'll insert the value from `vscode.workspace.name` (or if that's not set, `"unnamed-workspace"` - although I think that only happens if you have no folders open).

The values VS Code gives us are:

- The name (without the extension) of the current `.code-workspace` file (if there is one)
- The name of the folder that's open (if not a workspace)
- "untitled workspace" when you have multiple folders open but not saved as a workspace file

Here's an example from my logs folder after opening a few projects (some of which are code-workspaces):

![image](https://github.com/Dart-Code/Dart-Code/assets/1078012/162f72a6-e4a3-49e8-9e04-70ac97049cc6)

FMorschel:
Hi! Not sure if I'm doing something wrong, but I tested with:

```
[...]\\logs\\${workspaceName}\\${name}\\file.txt
```

And got:

```
[...]\logs\entrega_sunsoft\${name}\file.txt
```

Like, it didn't replace `${name}` with anything.

## Separate

Also, here are all the files I'm setting:

![image](https://github.com/user-attachments/assets/01a4ef54-dc0a-4249-bbd8-787566c5a283)

[...]

PR #5185, opened 2024-07-16
Updating markdown description of LogFiles

FMorschel:
This PR is to address the following part of a comment I made:

> Also, here are all the files I'm setting:
> 
> image
> 
> Although some of the markdown hints you changed it so it would show that `${workspaceName}` would work, only flutterTestLogFile does show that we can also use `${name}`:
> 
> image
> 
> And the `extensionLogFile` shows neither.
> 
> image
> 
Originally posted by @FMorschel in https://github.com/Dart-Code/Dart-Code/issues/5154#issuecomment-2228343112

### DIFF
--- a/package.json
+++ b/package.json
@@ -2347,7 +2347,7 @@
 							"string"
 						],
 						"default": null,
-						"markdownDescription": "The path to a log file for communication between Dart Code and the Analysis Server. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
+						"markdownDescription": "The path to a log file for communication between Dart Code and the Analysis Server. Use `${name}` in the log file name to insert the Debug Session name to prevent concurrent debug sessions overwriting each others logs. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
 						"scope": "machine-overridable"
 					},
 					"dart.toolingDaemonLogFile": {
@@ -2356,7 +2356,7 @@
 							"string"
 						],
 						"default": null,
-						"markdownDescription": "The path to a log file for the `dart tooling-daemon` service, which coordinates between various Dart and Flutter tools and extensions. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
+						"markdownDescription": "The path to a log file for the `dart tooling-daemon` service, which coordinates between various Dart and Flutter tools and extensions. Use `${name}` in the log file name to insert the Debug Session name to prevent concurrent debug sessions overwriting each others logs. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
 						"scope": "machine-overridable"
 					},
 					"dart.dapLogFile": {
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -2338,7 +2338,7 @@
 							"string"
 						],
 						"default": null,
-						"markdownDescription": "The path to a log file for very detailed logging in the Dart Analysis Server that may be useful when trying to diagnose Analysis Server issues. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
+						"markdownDescription": "The path to a log file for very detailed logging in the Dart Analysis Server that may be useful when trying to diagnose Analysis Server issues. Use `${name}` in the log file name to insert the Debug Session name to prevent concurrent debug sessions overwriting each others logs. Use `${workspaceName}` to insert the name of the current workspace in the file path.",
```
DanTup, line 2341: `${name}` isn't supported here, it's only supported in the loges that are debug-session specific (that is DAP, DartTest, FlutterRun, FlutterTest, WebDaemon).

I think as of the current master version, the descriptions should all be accurate.

src/extension/extension.ts
```diff
@@ -183,6 +183,7 @@ export async function activate(context: vs.ExtensionContext, isRestart = false)
 	setupLog(config.analyzerLogFile, LogCategory.Analyzer);
 	setupLog(config.flutterDaemonLogFile, LogCategory.FlutterDaemon);
 	setupLog(config.toolingDaemonLogFile, LogCategory.DartToolingDaemon);
+	setupLog(config.dapLogFile, LogCategory.DAP);
```
DanTup, line 186: This seems to be reverting some changes I just pushed to `master` to support "dynamic" DAP log file names.

src/extension/providers/debug_adapter_logger_factory.ts
```diff
@@ -1,37 +1,23 @@
...
-import { captureLogs, CategoryLogger, EmittingLogger } from "../../shared/logging";
-import { config } from "../config";
-import { insertSessionName } from "../utils";
-import { getLogHeader } from "../utils/log";
+import { LogCategory } from "../../shared/enums";
+import { Logger } from "../../shared/interfaces";
+import { CategoryLogger } from "../../shared/logging";
 
 export class DartDebugAdapterLoggerFactory implements DebugAdapterTrackerFactory {
-	constructor(private readonly logger: EmittingLogger) { }
```
DanTup, line 10: Same here and below, I think you've reverted this PR to before my changes so it's trying to undo them.

FMorschel:
Since I'm not sure how to fix this besides going through every file changed and fixing them by hand, I'll recreate this PR with only the changes for `package.json`.

DanTup:
sgtm - though it's not clear to me what still needs fixing - I thought I'd updated everything to be accurate in my last few changes (I didn't see you were also working on these until afterwards unfortunately!).

FMorschel:
> ${name} isn't supported here, it's only supported in the loges that are debug-session specific (that is DAP, DartTest, FlutterRun, FlutterTest, WebDaemon).
> 
> I think as of the current master version, the descriptions should all be accurate.

I see what you mean then. The only thing my PR would add is more consistency between all of the descriptions and a typo fix. Will rename it then.

PR #5203, opened 2024-07-30
Fix resolvePaths to handle Windows paths with backslashes

FMorschel:
Adding the option to replace `~` for the home directory in paths that use backslashes.

Also, making clearer that only the noted substitutions are allowed and everything else will be left in the path as-is.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2347,7 +2347,7 @@
 							"string"
 						],
 						"default": null,
-						"markdownDescription": "The path to a log file for very detailed logging in the Dart Analysis Server that may be useful when trying to diagnose Analysis Server issues. Use `${workspaceName}` to insert the name of the current workspace in the file path. Use `~` to insert the user's home directory. Only the noted substitutions are supported, others will stay as-is.",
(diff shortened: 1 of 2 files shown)


DanTup:
Thanks!

DanTup (approved):

DanTup:
Hmm, while checking if we had tests for this, I found this comment:

https://github.com/Dart-Code/Dart-Code/blob/bcdaa1e6b6dd13e0ad3f7498a3bf98bfe033ee5d/src/test/dart/util.test.ts#L248-L249

Which led me to this:

https://github.com/Dart-Code/Dart-Code/blob/bcdaa1e6b6dd13e0ad3f7498a3bf98bfe033ee5d/src/shared/utils/fs.ts#L51

I think the reasoning here was that we want to be able to translate in both directions (so we can shorten display paths to use `~` when they're in the home dir), and in order to support teams that might want to commit the SDK paths (if they have a standard location for them, possibly to pin multiple versions), it would be better if the format was fixed across platforms.

So, I think it would be better to revert the change to support `\` and instead note in the docs that `~` can be used for home directory, but the path should always use format slashes even on Windows (`~/flutter/`).

FMorschel:
Alright, will do that. Missed those tests.

DanTup:
Thanks!

Issue #5183, opened 2024-07-16
Launching a new app in VSCode fails quickly with no error in VS Code insiders


natebot13:
**Describe the bug**
Attempting to launch either an old app or a freshly generated new app using VSCode's "Run" option briefly shows the debugger controls, then stops. I've checked extension outputs and vscode's developer console for errors, but nothing reveals any insight. When running from command line with "flutter run", the app launches successfully.

**To Reproduce**
1. Start with a clean checkout of flutter stable (version and system info below)
2. Generate a fresh app
3. Use VSCode to try running it
4. Observe it doesn't launch

**Hints**
I've found that adding the setting `"dart.useLegacyDebugAdapters": true` to my project allows it to be successfully launched by vscode. I assume it's related to something failing with using the non-legacy debug adapter. If I could find some logs for that, I'm sure the issue might be clearer.

<details>

<summary><strong>Workspace Environment</strong></summary>

```text
Dart Code extension: 3.92.0
Flutter extension: 3.93.20240702 (activated)

App: Visual Studio Code - Insiders
App Host: desktop
Version: win 1.92.0-insider

Workspace type: Flutter (LSP)

Dart (3.4.3): C:\src\flutter\bin\cache\dart-sdk
Flutter (3.22.2): C:\src\flutter (Windows (windows-x64/windows))

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`C:\src\flutter\bin\cache\dart-sdk\bin\dart.exe info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.4.3 (stable) (Tue Jun 4 19:51:39 2024 +0000) on "windows_x64"
- on windows / "Windows 10 Pro" 10.0 (Build 22635)
- locale is en-US

#### Project info

- sdk constraint: '>=2.15.0 <3.0.0'
- dependencies: bloc, flutter, flutter_bloc, flutter_highlight, flutter_localizations, google_fonts, highlighter, intl, json_annotation, protobuf, rive
- dev_dependencies: build_runner, flutter_test, json_serializable, pedantic, protoc_builder

#### Process info

| Memory | CPU | Elapsed time | Command line |
| -----: | --: | -----------: | ------------ |
| 111 MB |  -- |              | dart.exe     |
| 626 MB |  -- |              | dart.exe     |
| 103 MB |  -- |              | dart.exe     |
[...]

PR #5204, opened 2024-07-30
Spawn Flutter debug adapter directly with dart.exe

DanTup:
This is a fix (workaround) for a potentially serious issue debugging Flutter on Windows with the incoming VS Code (editor) release. I plan to ship it a Dart-Code pre-release shortly (today) to get some testing and may ship in a stable release tomorrow or Thursday.

@andrewkolos @jwren @helin24 I would appreciate additional eyes over this change before the stable release if possible. It's not a particularly nice workaround but it's currently all I have.

A summary of the issue:

A recent breaking change in NodeJS means spawning a `.bat` shell script on Windows without explicitly setting `shell: true` will fail with an `EINVAL` error. The upcoming release of the VS Code editor (being prepared and likely to ship this or next week) updates to a version of NodeJS that includes that change. Our extension code has always used `shell: true` where necessary, however the Debug Adapter (`flutter debug_adapter`) is not started directly by us in the extension - instead, we pass a `executable: string, args: string[]` to the VS Code editor and it spawns the process. It does not provide the ability for us to tell it to use `shell: true`. I have filed https://github.com/microsoft/vscode/issues/224184 and (open a PR at https://github.com/microsoft/vscode/pull/224204 with a fix) however because it's close to release and my proposed fix was an API change, it will not be merged before the stable VS Code release.

If we do nothing, Windows users will be unable to launch Flutter debug sessions with the upcoming VS Code release (and this is already the case for those using Insider builds of VS Code).

As a workaround, when using the affected version of VS Code (on Windows), I am substituting `"sdk-path\bin\flutter.bat debug_adapter"` with the equivalent command that the shell script runs `"(sdk-path)\bin\cache\dart-sdk\bin\dart --packages="..." (tool-snapshot-path) debug_adapter`. This works for me locally in Insiders, and is passing all of the  (previously failing) CI tests run against insiders.

As I was writing this, VS Code devs agreed to look at another fix (https://github.com/microsoft/vscode/issues/224184#issuecomment-2258617405) so we might not need this - however I'm going to continue with putting this into a pre-release extension for additional testing so if they are unable to ship a fix, we have this to fall back on (and it isn't completely untested).

(I'm merging it so it can go into the pre-release so this PR will be closed, but if you have any feedback or concerns, please still leave them here and I will address).

Fixes https://github.com/Dart-Code/Dart-Code/issues/5183

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -599,7 +599,7 @@ export async function activate(context: vs.ExtensionContext, isRestart = false) 
 	context.subscriptions.push(vs.debug.registerDebugAdapterTrackerFactory("dart", debugLogger));
 	const trackerFactories = [globalEvaluationContext, hexFormatter, forcedDebugMode, removeErrorShowUser, supportUris, launchStatus, debugLogger];
 
-	const debugAdapterDescriptorFactory = new DartDebugAdapterDescriptorFactory(analytics, sdks, logger, extContext, dartCapabilities, flutterCapabilities, workspaceContext, experiments);
+	const debugAdapterDescriptorFactory = new DartDebugAdapterDescriptorFactory(analytics, sdks, logger, extContext, dartCapabilities, flutterCapabilities, vsCodeVersion, workspaceContext, experiments);
 	context.subscriptions.push(vs.debug.registerDebugAdapterDescriptorFactory("dart", debugAdapterDescriptorFactory));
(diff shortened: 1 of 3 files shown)


helin24 (commented):
LGTM

helin24:
Good news 👍 Thanks for the update!

PR #5257, opened 2024-09-10
Add the ability to show notifications when the current SDK is unsupported (or will be soon)

DanTup:
This adds two new fields to `DartCapabilities` for `isUnsupportedNow` and `isUnsupportedSoon` to determine whether to warn the user about unsupported SDKs. This is a necessary step before we can remove any legacy code (such as the legacy analysis server integration, or activating DevTools from pub!).

We always do the SDK check against the Dart SDK version, but if the current Dart SDK comes from Flutter, we will reference "Flutter" and its SDK version in the user-facing messages.

"Unsupported now" messages will always be shown as the intention is that the user will either upgrade their SDK or switch to an older extension.. this may seem annoying, but the extension will be broken in many ways when using an SDK that triggers this, so ignoring it is not an option - you must upgrade SDK or switch to an older extension.

"Unsupported soon" messages will be shown only until dismissed the first time (clicking "More Info" counts as dismissing), but this is tracked by major+minor versions, so dismissing for Dart 2.0 would not prevent seeing the message again in future for 3.0.

Currently looks like this (version number is just the SDK I was testing with, ofc):

Screenshot 2024-09-10 175404

Screenshot 2024-09-10 175318

Where it says "Flutter SDK" and shows a Flutter SDK version number will be "Dart SDK" if the selected SDK is a standalone Dart SDK.

## TODO Before Merging

- [x] agree specific text for each message
- [ ] ~~update extension on https://dartcode.org/sdk-version-compatibility/~~ moved to https://github.com/Dart-Code/Dart-Code/issues/5304
- [x] agree specific SDK version that will become "unsupported soon"
- [x] agree which Dart-Code release this will first ship in
- [x] agree a URL to link to for the "More Info" button
- [x] add a matrix of supported extension / SDK versions (on dartcode.org, or the dart.dev site?)

@jwren @helin24 @itsjustkevin

### DIFF
--- a/src/extension/extension.ts
+++ b/src/extension/extension.ts
@@ -765,7 +765,7 @@ export async function activate(context: vs.ExtensionContext, isRestart = false) 
 
 	// Prompt user for any special config we might want to set.
 	if (!isRestart)
-		void showUserPrompts(logger, extContext, webClient, analytics, workspaceContext, extensionRecommendations);
+		void showUserPrompts(logger, extContext, webClient, analytics, workspaceContext, dartCapabilities, extensionRecommendations);
 
 	// Turn on all the commands.
 	setCommandVisiblity(true, workspaceContext);
--- a/src/extension/user_prompts.ts
+++ b/src/extension/user_prompts.ts
@@ -1,13 +1,14 @@
 import * as fs from "fs";
 import * as path from "path";
 import * as vs from "vscode";
+import { DartCapabilities } from "../shared/capabilities/dart";
 import { DART_CREATE_PROJECT_TRIGGER_FILE, FLUTTER_CREATE_PROJECT_TRIGGER_FILE, installFlutterExtensionPromptKey, isWin, noAction, recommendedSettingsUrl, showRecommendedSettingsAction, useRecommendedSettingsPromptKey, userPromptContextPrefix, yesAction } from "../shared/constants";
 import { LogCategory } from "../shared/enums";
 import { WebClient } from "../shared/fetch";
 import { Analytics, DartProjectTemplate, FlutterCreateCommandArgs, FlutterCreateTriggerData, Logger } from "../shared/interfaces";
 import { fsPath } from "../shared/utils/fs";
 import { checkHasFlutterExtension, extensionVersion, getExtensionVersionForReleaseNotes, hasFlutterExtension, isDevExtension } from "../shared/vscode/extension_utils";
-import { showFlutterSurveyNotificationIfAppropriate } from "../shared/vscode/user_prompts";
+import { showFlutterSurveyNotificationIfAppropriate, showSdkDeprecationNoticeIfAppropriate } from "../shared/vscode/user_prompts";
 import { envUtils, getDartWorkspaceFolders } from "../shared/vscode/utils";
 import { Context } from "../shared/vscode/workspace";
 import { WorkspaceContext } from "../shared/workspace";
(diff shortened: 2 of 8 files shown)


helin24:
The text LGTM.

I'm curious what the matrix of supported extension and SDK versions will look like. What is reasonable to say for extensions prior to the one where we will discontinue support for v3.25 and earlier? "All SDK versions?"

DanTup:
> I'm curious what the matrix of supported extension and SDK versions will look like. What is reasonable to say for extensions prior to the one where we will discontinue support for v3.25 and earlier? "All SDK versions?"

(note: 3.25 was just the current Flutter SDK version I was testing with and isn't due to become unsupported, it's only currently on master 😄 specific versions haven't been published yet)

But it's a good question that I've been wondering myself. I feel like the goal of this information is that a user knows they are on SDK version x and they basically want to know "what version of the extension must I downgrade to for this SDK to fully work?". I don't think we need to track complete version ranges for each (I don't think there's much value in knowing the oldest version of an extension that supports a given SDK, only the newest), so we could just record the final extension version before we consciously remove something that prevents an older SDK from working.

So maybe some table like this (version numbers are just made up here, notes in italic just for explanation):

| Dart SDK | Flutter SDK | Dart Extension |
| ----- | ----- | ----- |
| v0.1 - v0.9 | | Supported until Dart extension 1.10 _(1.11 removed X which SDK <= v0.9 needs)_ |
| v1.0 - v1.11 | v1.0 - v1.8 | Supported until Dart extension 2.3 _(2.4 removed Y Dart SDK <= v1.11 needs)_ |
| v1.12 - v2.4 | v1.9 - v2.1 | Currently supported |

The first column is a non-overlapping set of ranges of the Dart SDK, the second column is the matching Flutter SDKs, and the final column is basically "the newest extension you can use and expect it to work correctly with this SDK". So if you're stuck on an old SDK, you find the row that contains that SDK version, then roll back to the noted version of the extension.

I'm interested in opinions/suggestions though :-)

DanTup:
> I feel like perhaps there's a better word than "supported" to convey "In theory supported, but older versions of SDK probably fail in unpredictable ways" 😆

Yeah, maybe we can come up with something better, or just start the table from a slightly more recent SDK (instead of `<`) that we spend a little time testing against that version of the plugin before we (deliberately) drop support for some versions.

DanTup:
Based on the announcement at https://groups.google.com/g/flutter-announce/c/JQHzM3FbBGI:

> Starting with the release of Dart 3.6, we will deprecate support for Dart SDK versions older than 3.0 

I've updated this to consider Dart SDKs before 3.0 as "unsupported soon" (that is, they will show a warning that only appears until the user interacts with it once, then it won't be shown again) and I will merge this PR once Dart 3.6 releases (so the warning will begin showing up with the next Dart-Code release after then).

I've added a basic page to https://dartcode.org/sdk-version-compatibility/ with a table as noted above, and linked the "More Info" button in the notification to that page (which also contains a link to the flutter-announce post).

The notification (when using a 2.19 Dart SDK) looks like this:

Screenshot 2024-09-24 110530

@jwren @itsjustkevin please let me know if this (including the text here and in the original screenshots at the top, which includes text for when the SDK support is removed - which will happen around the Dart 3.7 release for < 3.0) all looks reasonable to you or if we should change before this merges (which won't be until Dart 3.6 releases).

itsjustkevin:
Thanks @DanTup LGTM!

jwren (approved):
LGTM, we'll coordinate in coming weeks to synchronize IJ and VS Code here.

DanTup:
I've removed the version number from here so I can merge this PR now (without the notification showing up to pre-release users if I push a pre-release build), and have opened https://github.com/Dart-Code/Dart-Code/issues/5304 for the current milestone to put it back prior to shipping.

That means, with the next release (somewhere around the end of this month / start of next month) we'll begin showing the notification.

Issue #3628, opened 2021-10-22
Allow customisation of closing label prefix and font/size


twoco:
Please allow to change the character and / or other styles of the closing labels I see in VS Code. 
Currently they look like comments. This triggers me. 

For example change the `//` slashes to `#` hashtag or `/` single slash or `-` minus, etc..  (just an user input)
Or nothing. I don't think the syntax matters. No need for slashes.
Changing the color and / or opacity would be also great.

Setting: `dart.closingLabels`

![image](https://user-images.githubusercontent.com/53160891/138522493-b016900a-e9bd-46da-8e1a-cf90408f8874.png)

infacto:
The close label indicator feature is great, but it's really annoying that it's hard to distinguish from real comments.

My current workaround is adding this to *settings.json*:

```json
{
  "workbench.colorCustomizations": {
    "dart.closingLabels": "#80808040"
  }
}
```

If we add an option to customize the text, I would suggest prefix and suffix. To allow something like `-- Text` an `(Text)` etc. Or just something like `"dart.closingLabelsString": "/* $1 */"` But only a suffix is also great. Maybe other style can also help. Like font-size? Idk. Just separate from real comments. Or improve Dart / Flutter design making closing labels annotations unnecessary. ...

gazialankus:
> Like font-size? Idk.

YES. See the inlay hints with the small font size? I would love it if I can get these closing labels to look the same. Then my eyes can group them similarly as *not real code but suggestions*. I really like this small font size for the inlay hints, so I'm pretty sure it will work perfectly for me if I can change the font. 
<img width="416" alt="image" src="https://github.com/Dart-Code/Dart-Code/assets/412997/83ad5dce-694a-40f0-982a-47381307117d">

For anyone wondering, here are the settings for the inlay hints: 
```
    "[dart]": {
        "editor.inlayHints.enabled": "on",
        "editor.inlayHints.fontSize": 8,
        "editor.inlayHints.padding": false,
        "editor.inlayHints.fontFamily": "Arial", 
```

DanTup:
I think it's a fair request to be able to customise these further and would accept PRs to do this. The styling is handled here:

https://github.com/Dart-Code/Dart-Code/blob/1ee5959ff4d68510b0c03bbbe1ee971b61d614ca/src/extension/lsp/closing_labels_decorations.ts#L15-L18

And the prefix is handled here:

[...]

PR #5263, opened 2024-09-12
Closing label customisation

FMorschel:
Fixes https://github.com/Dart-Code/Dart-Code/issues/3628#.

I'm not sure the `recreateDecorationType` is the right approach here but without it, I could not make the label update when changing the propriety on the settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1985,6 +1985,29 @@
 						"description": "Whether to show annotations against constructor, method invocations and lists that span multiple lines.",
 						"scope": "window"
 					},
+					"dart.closingLabelsPrefix": {
+						"type": [
+							"null",
+							"string"
+						],
+						"default": null,
+						"markdownDescription": "The prefix to use for closing labels. If `null`, the default ` // ` prefix will be used.",
+						"scope": "window"
+					},
+					"dart.closingLabelsStyle": {
+						"type": "string",
+						"enum": [
+							"normal",
+							"italic"
+						],
+						"enumItemLabels": [
+							"Normal",
+							"Italic"
+						],
+						"default": "normal",
+						"description": "The style to use for closing labels. This setting requires a theme that supports italic text.",
+						"scope": "window"
+					},
 					"dart.completeFunctionCalls": {
 						"type": "boolean",
 						"default": true,
--- a/package.json
+++ b/package.json
@@ -1986,15 +1986,12 @@
 						"scope": "window"
 					},
 					"dart.closingLabelsPrefix": {
-						"type": [
-							"null",
-							"string"
-						],
-						"default": null,
-						"markdownDescription": "The prefix to use for closing labels. If `null`, the default ` // ` prefix will be used.",
+						"type": "string",
+						"default": " // ",
+						"description": "The prefix to use in closing labels.",
 						"scope": "window"
 					},
-					"dart.closingLabelsStyle": {
+					"dart.closingLabelsTextStyle": {
 						"type": "string",
 						"enum": [
 							"normal",
@@ -2005,7 +2002,7 @@
 							"Italic"
 						],
 						"default": "normal",
-						"description": "The style to use for closing labels. This setting requires a theme that supports italic text.",
+						"markdownDescription": "The text style to use in closing labels. Using _Italic_ requires a font that supports italic text.",
 						"scope": "window"
 					},
 					"dart.completeFunctionCalls": {
--- a/src/extension/config.ts
+++ b/src/extension/config.ts
@@ -71,6 +71,8 @@ class Config {
 	get cliConsole(): "debugConsole" | "terminal" | "externalTerminal" { return this.getConfig<"debugConsole" | "terminal" | "externalTerminal">("cliConsole", "debugConsole"); }
 	get closeDevTools(): "never" | "ifOpened" | "always" { return this.getConfig<"never" | "ifOpened" | "always">("closeDevTools", "never"); }
 	get closingLabels(): boolean { return this.getConfig<boolean>("closingLabels", true); }
+	get closingLabelsPrefix(): string { return this.getConfig<string>("closingLabelsPrefix", " // "); }
+	get closingLabelsStyle(): string { console.log("Changed to:" + this.getConfig<string>("closingLabelsStyle", "normal")); return this.getConfig<string>("closingLabelsStyle", "normal"); }
 	get completionBudgetMilliseconds(): number | undefined { return this.getConfig<number | undefined>("completionBudgetMilliseconds", undefined); }
 	get customDartDapPath(): undefined | string { return resolvePaths(this.getConfig<null | string>("customDartDapPath", null)); }
 	get customDevToolsUri(): undefined | string { return this.getConfig<undefined | string>("customDevToolsUri", undefined); }
--- a/src/extension/config.ts
+++ b/src/extension/config.ts
@@ -72,7 +72,7 @@ class Config {
 	get closeDevTools(): "never" | "ifOpened" | "always" { return this.getConfig<"never" | "ifOpened" | "always">("closeDevTools", "never"); }
 	get closingLabels(): boolean { return this.getConfig<boolean>("closingLabels", true); }
 	get closingLabelsPrefix(): string { return this.getConfig<string>("closingLabelsPrefix", " // "); }
-	get closingLabelsStyle(): string { console.log("Changed to:" + this.getConfig<string>("closingLabelsStyle", "normal")); return this.getConfig<string>("closingLabelsStyle", "normal"); }
+	get closingLabelsTextStyle(): string { return this.getConfig<string>("closingLabelsTextStyle", "normal"); }
 	get completionBudgetMilliseconds(): number | undefined { return this.getConfig<number | undefined>("completionBudgetMilliseconds", undefined); }
 	get customDartDapPath(): undefined | string { return resolvePaths(this.getConfig<null | string>("customDartDapPath", null)); }
 	get customDevToolsUri(): undefined | string { return this.getConfig<undefined | string>("customDevToolsUri", undefined); }
--- a/src/extension/decorations/closing_labels_decorations.ts
+++ b/src/extension/decorations/closing_labels_decorations.ts
@@ -4,6 +4,7 @@ import { disposeAll } from "../../shared/utils";
 import { fsPath } from "../../shared/utils/fs";
 import { DasAnalyzerClient } from "../analysis/analyzer_das";
 import { isAnalyzable } from "../utils";
+import { config } from "../config";
 
 export const validLastCharacters = [")", "]"];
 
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -1985,6 +1985,29 @@
 						"description": "Whether to show annotations against constructor, method invocations and lists that span multiple lines.",
 						"scope": "window"
 					},
+					"dart.closingLabelsPrefix": {
+						"type": [
+							"null",
+							"string"
+						],
```
DanTup, line 1992: We should just make this `string` and use `// ` as the `default` value here instead. This makes it easier (or possible) to edit in the settings UI.

package.json
```diff
@@ -1985,6 +1985,29 @@
...
+						"default": null,
+						"markdownDescription": "The prefix to use for closing labels. If `null`, the default ` // ` prefix will be used.",
+						"scope": "window"
+					},
+					"dart.closingLabelsStyle": {
```
DanTup, line 1997: nit: I'd call this `closingLabelsTextStyle`

package.json
```diff
@@ -1985,6 +1985,29 @@
...
+						"enum": [
+							"normal",
+							"italic"
+						],
+						"enumItemLabels": [
+							"Normal",
+							"Italic"
+						],
+						"default": "normal",
+						"description": "The style to use for closing labels. This setting requires a theme that supports italic text.",
```
DanTup, line 2008: Same here, "The text style"

> This setting requires a theme that supports italic text.

nit: "Using _Italic_ request a theme that supports it" (although I didn't realise the theme would matter here?)

src/extension/config.ts
```diff
@@ -71,6 +71,8 @@ class Config {
 	get cliConsole(): "debugConsole" | "terminal" | "externalTerminal" { return this.getConfig<"debugConsole" | "terminal" | "externalTerminal">("cliConsole", "debugConsole"); }
 	get closeDevTools(): "never" | "ifOpened" | "always" { return this.getConfig<"never" | "ifOpened" | "always">("closeDevTools", "never"); }
 	get closingLabels(): boolean { return this.getConfig<boolean>("closingLabels", true); }
+	get closingLabelsPrefix(): string { return this.getConfig<string>("closingLabelsPrefix", " // "); }
```
DanTup, line 74: The default argument passed here should always match the one in the `package.json`. We have code to generate these getters, so they're periodically overwritten with the code-generated values (which take defaults from the manifest).

src/extension/config.ts
```diff
@@ -71,6 +71,8 @@ class Config {
...
+	get closingLabelsStyle(): string { console.log("Changed to:" + this.getConfig<string>("closingLabelsStyle", "normal")); return this.getConfig<string>("closingLabelsStyle", "normal"); }
```
DanTup, line 75: I think this is leftover debug logging

src/extension/decorations/closing_labels_decorations.ts
```diff
@@ -12,15 +13,22 @@ export class ClosingLabelsDecorations implements vs.Disposable {
...
-		after: {
-			color: new vs.ThemeColor("dart.closingLabels"),
-			margin: "2px",
-		},
-		rangeBehavior: vs.DecorationRangeBehavior.ClosedOpen,
-	});
+	private decorationType: vs.TextEditorDecorationType | null = null;
+
+	private recreateDecorationType() {
+		if (this.decorationType) this.decorationType.dispose();
```
DanTup, line 19: nit: newline after the condition

src/extension/decorations/closing_labels_decorations.ts
```diff
@@ -57,27 +65,31 @@ export class ClosingLabelsDecorations implements vs.Disposable {
...
 					range: new vs.Range(this.activeEditor.document.positionAt(r.offset), endOfLine),
-					renderOptions: { after: { contentText: " // " + r.label } },
+					renderOptions: { after: { contentText: closingLabelsPrefix + r.label } },
 				};
 				decorations[endOfLine.line] = dec;
 			}
 		}
 
-		this.activeEditor.setDecorations(this.decorationType, Object.keys(decorations).map((k) => parseInt(k, 10)).map((k) => decorations[k]));
+		this.recreateDecorationType();
```
DanTup, line 82: Recreating this every time is probably overkill (I don't know what work VS Code might end up doing if we keep disposing and creating new kinds), but we could set up a handler in the constructor that listens to config changes similar to we do for formatting:

```ts
		this.disposables.push(workspace.onDidChangeConfiguration((e) => {
			if (e.affectsConfiguration("dart.enableSdkFormatter")) {
				if (config.enableSdkFormatter)
					this.registerAllFormatters();
				else
					this.unregisterAllFormatters();
			}
		}));
```

src/extension/decorations/closing_labels_decorations.ts
```diff
@@ -57,27 +65,31 @@ export class ClosingLabelsDecorations implements vs.Disposable {
...
+		this.activeEditor.setDecorations(this.decorationType!, Object.keys(decorations).map((k) => parseInt(k, 10)).map((k) => decorations[k]));
 	}
 
 	private setTrackingFile(editor: vs.TextEditor | undefined) {
 		if (!editor || !isAnalyzable(editor.document))
 			return;
 
 		if (editor !== this.activeEditor)
-			this.activeEditor?.setDecorations(this.decorationType, []);
+			this.activeEditor?.setDecorations(this.decorationType!, []);
```
DanTup, line 92: I think it's be nice to avoid the `!`s here.. perhaps by making sure it's first set up in the constructor and removing the null option from it.

src/extension/lsp/closing_labels_decorations.ts
```diff
@@ -74,26 +82,30 @@ export class LspClosingLabelsDecorations implements vs.Disposable {
...
 				const dec = {
 					range: new vs.Range(labelRange.start, endOfLine),
-					renderOptions: { after: { contentText: " // " + r.label } },
+					renderOptions: { after: { contentText: closingLabelsPrefix + r.label } },
 				};
 				decorations[endOfLine.line] = dec;
 			}
 		}
 
+		this.recreateDecorationType();
```
DanTup, line 99: (same comments as in the legacy version)

DanTup (commented):
Thanks! I've added some comments (mostly trivial).

package.json
FMorschel: > although I didn't realise the theme would matter here?

I would think so, or at least the font would matter. I can probably say _font_ there as well.

But if you don't believe this is true I could remove this. I could not find a font that didn't had this option to test, if you can, please let me know.

src/extension/lsp/closing_labels_decorations.ts
```diff
@@ -41,6 +51,16 @@ export class LspClosingLabelsDecorations implements vs.Disposable {
 			const filePath = fsPath(td.uri);
 			delete this.closingLabels[filePath];
 		}));
+		this.subscriptions.push(vs.workspace.onDidChangeConfiguration((e) => {
+			if (e.affectsConfiguration("dart.closingLabels") || e.affectsConfiguration("dart.closingLabelsPrefix") || e.affectsConfiguration("dart.closingLabelsTextStyle")) {
```
DanTup, line 55: I don't think we need to check the prefix here, since it's not part of the decoration type.

src/extension/decorations/closing_labels_decorations.ts
```diff
@@ -12,15 +13,24 @@ export class ClosingLabelsDecorations implements vs.Disposable {
...
-		after: {
-			color: new vs.ThemeColor("dart.closingLabels"),
-			margin: "2px",
-		},
-		rangeBehavior: vs.DecorationRangeBehavior.ClosedOpen,
-	});
+	private decorationType!: vs.TextEditorDecorationType;
+	private closingLabelsPrefix!: string;
+
+	private getDecorationType() {
```
DanTup, line 19: nit: I'd call this "update", "build" or "create" or something, since "get" seems to imply it returns something, but it doesn't.

src/extension/decorations/closing_labels_decorations.ts
```diff
@@ -12,15 +13,24 @@ export class ClosingLabelsDecorations implements vs.Disposable {
...
 	private closingLabels?: as.AnalysisClosingLabelsNotification;
 
-	private readonly decorationType = vs.window.createTextEditorDecorationType({
-		after: {
-			color: new vs.ThemeColor("dart.closingLabels"),
-			margin: "2px",
-		},
-		rangeBehavior: vs.DecorationRangeBehavior.ClosedOpen,
-	});
+	private decorationType!: vs.TextEditorDecorationType;
```
DanTup, line 16: Do we need these `!`s here? It's not clear to me what they're doing.

DanTup (commented):
Some very minor comments, but otherwise it's looking good - thanks! :)

src/extension/lsp/closing_labels_decorations.ts
FMorschel: It is part of the configurations. So we don't need to "get" that value every time. If you disagree I can move it back or at least separate the `if`s.

src/extension/decorations/closing_labels_decorations.ts
FMorschel: Without it, since we are updating the property only inside the (current, will rename) `buildDecorationType` we get:

```console
Property 'decorationType' has no initializer and is not definitely assigned in the constructor.
```

The `!` there fixes it. I was getting the same error on the `closingLabelsPrefix` before I initiated it there so this one is not needed.

src/extension/lsp/closing_labels_decorations.ts
FMorschel: It could be something like:

```ts
let updatedWatched = false;
if (e.affectsConfiguration("dart.closingLabelsPrefix")) {
    updatedWatched = true;
    this.closingLabelsPrefix = config.closingLabelsPrefix;
}
if (e.affectsConfiguration("dart.closingLabels") || e.affectsConfiguration("dart.closingLabelsTextStyle")) {
    updatedWatched = true;
    this.decorationType.dispose();
    this.buildDecorationType();
}
if (updatedWatched) {
    this.update();
}
```

src/extension/decorations/closing_labels_decorations.ts
DanTup: Ah, got it - I didn't realise you could silence that in this way. I guess this is necessary unless we assign it in the constructor, but that makes it harder to share the `buildDecorationType` implementation. Keep it is fine.

DanTup (approved):
Thanks, this lgtm - I'll pull it down and do a little testing when I get a min and if it's all good merge it in :-)

DanTup:
This LGTM, thanks!

I found a quirk while testing, but it's not a result of this PR - we have it today (but it's easier to notice now there are more settings you might be customising) - I've opened https://github.com/Dart-Code/Dart-Code/issues/5265 about that.

Fernandomr88:
how to use `dart.closingLabelsPrefix`? couldn't find it in the release notes

DanTup:
@Fernandomr88 it's a VS Code setting that shows up in the user/workspace settings:

image

stan-at-work:
Why not also let us change the color, fontSize and suffix ?

Maybe something like thise:

`// ${className} //`

DanTup:
@stan-at-work you can already change the colour (using `dart.closingLabels` in `workbench.colorCustomizations`). I don't think we can alter the size (the settings we can set are at https://code.visualstudio.com/api/references/vscode-api#ThemableDecorationAttachmentRenderOptions), and there's no reason we couldn't support a suffix, it just hasn't been requested.

PR #5341, opened 2024-11-18
Add inherited widget snippet #5340

### LINKED ISSUE #5340: Add inherited widget snippet
Would like to add to the existing `stless` and `stful` shortcuts an `inhr` shortcut to create Inherited Widgets. The boilerplate for an `InheritedWidget` is probably the hardest to remember out of the `Stateful`/`Stateless`/`Inherited` trinity.

Added to my own user snippets but I think it might as well be in the core snippets.

zbarbuto:
Closes  #5340

https://github.com/user-attachments/assets/81dbcc2a-7db5-4548-a525-49a3264a671b

### DIFF
--- a/snippets/flutter.json
+++ b/snippets/flutter.json
@@ -66,6 +66,28 @@
 				"  }",
 				"}"
 			]
+		},
+		"Flutter Inherited Widget": {
+			"prefix": "inhr",
+			"description": "Insert an InheritedWidget",
+			"body": [
+				"class $1 extends InheritedWidget {",
+				"  const $1({super.key});",
+				"",
+				"  static $1? maybeOf(BuildContext context) {",
+				"    return context.dependOnInheritedWidgetOfExactType<$1>();",
+				"  }",
+				"",
+				"  static $1 of(BuildContext context) {",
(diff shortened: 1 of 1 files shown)


DanTup:
Thanks for the PR!

This file here is actually for legacy snippets (that will be removed in the future). Changes here won't show up unless you specifically disable server snippets (or are on a very old SDK) - something we don't want users to do.

The server-supplied versions of the snippets live in the Dart analysis server here:

https://github.com/dart-lang/sdk/tree/main/pkg/analysis_server/lib/src/services/snippets/dart
https://github.com/dart-lang/sdk/tree/main/pkg/analysis_server/test/services/snippets/dart

If we want to add this, it should be contributed there. I don't use Flutter enough to know how common use of this class is (and therefore whether it makes sense to provide the snippet to all users by default) - @gspencergoog do you have an opinion on this?

zbarbuto:
@DanTup thanks for pointing me in the right direction for where to add this.

In light of what @gspencergoog has said I will close this PR and the original issue for now. If people are interested they can re-open it and I'm happy to contribute in the correct places.

PR #5448, opened 2025-03-20
Prioritize using studio instead of studio.sh when opening Android Studio

davidmartos96:
Right now, when opening Android Studio from VSCode, the following is shown. The "Learn more" link goest to here: https://youtrack.jetbrains.com/articles/SUPPORT-A-56/How-to-handle-Switch-to-a-native-launcher-notification

image


DanTup:
Thanks!

PR #5458, opened 2025-03-26
Provide server with useful diagnostic values to include in exportable reports

DanTup:
These are sent once the server has started, and also if any config change affects any of the interesting sections (which are currently "dart" and "editor"). The reported config is scoped to Dart config (falling back to defaults if no Dart-specific config is set),  which means we can't tell if someone has `formatOnSave` set to something different for YAML files (for example), but if becomes useful, we could split this out to record multiple languages values.

This info is visible in the analyzer diagnostics pages as well as the exported report. For a sample workspace it looks like this:

image

For a single-root workspace, only the `global` section would be present. For a multi-root workspace, we'll add `folder_x` sections for any workspace folder that has settings that differ from the global one.

@bwilkerson @pq the settings included for now are below - let me know if you think there's anything else we should include (we can ofc update this over time). Additionally, I've included the "hostKind", which is a string that helps identify the environment being run in (for example Docker, WSL, Codespaces, or just native VS Code desktop).

```js
const interestingSettings: { [key: string]: string[] } = {
	dart: [
		"onlyAnalyzeProjectsWithOpenFiles", // This affects the analysis roots and whether we change them as files open/close which can affect perf
	],
	editor: [
		"formatOnSave",
		"formatOnPaste",
		"formatOnType",
		"codeActionsOnSave",
	],
};
```

### DIFF
--- a/src/extension/analysis/analyzer.ts
+++ b/src/extension/analysis/analyzer.ts
@@ -13,6 +13,7 @@ import { DartSdks, Logger } from "../../shared/interfaces";
 import { CategoryLogger } from "../../shared/logging";
 import { DartToolingDaemon } from "../../shared/services/tooling_daemon";
 import { fsPath } from "../../shared/utils/fs";
+import { AnalyzerUpdateDiagnosticInformationFeature } from "../../shared/vscode/analyzer_update_diagnostic_information";
 import { ANALYSIS_FILTERS } from "../../shared/vscode/constants";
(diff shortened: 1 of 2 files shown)


pq (approved):

pq:
Looks like a great start. Thank you!

bwilkerson (approved):
That output looks good to me too.

PR #5551, opened 2025-06-16
[experimental] Add support for advertising the Dart SDK MCP server

DanTup:
This uses the new VS Code API `registerMcpServerDefinitionProvider` to allow VS Code to ask us for MCP servers. If we're using a >= Dart 3.9.0-0 SDK and the (currently "hidden") `dart.experimentalMcpServer` setting is enabled, we will return the `dart mcp-server` command.

This allows VS Code to discover our MCP server and tools (which are then available to VS Code and other extensions).

Currently we don't provide a way to pass the DTD URI (at startup, or if we have to restart it, when the MCP server may continue to run - since VS Code controls its lifetime), but this will need adding.

@jakemac53 @kenzieschmoll

### DIFF
--- a/package.json
+++ b/package.json
@@ -797,7 +797,8 @@
 				"inputSchema": {
 					"type": "object",
 					"properties": {}
-				}
+				},
+				"when": "config.dart.experimentalMcpServer"
 			}
 		],
 		"menus": {
--- a/package.json
+++ b/package.json
@@ -786,6 +786,20 @@
 				"label": "Dart SDK MCP Servers"
 			}
 		],
+		"languageModelTools": [
+			{
+				"name": "get_dart_tooling_daemon_dtd_uri",
+				"displayName": "Get the Dart Tooling Daemon (DTD) URI",
+				"userDescription": "Gets the URI to connect to the Dart Tooling Daemon (DTD).",
+				"modelDescription": "Gets the URI to connect to the Dart Tooling Daemon (DTD).",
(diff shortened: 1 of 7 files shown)


DanTup:
@jakemac53 I see we have a [`lm.registerTool` API](https://code.visualstudio.com/api/references/vscode-api#lm.registerTool) that I missed before, which I think is what you were suggesting we use to provide a way for the LLM to ask for the DTD URI?

DanTup:
I pushed a change to register a tool for getting the DTD URI. I asked Copilot to explain my Flutter runtime error, and it successfully called the tool and provided it to the MCP server:

image

image


@jakemac53 could you review the names/descriptions I used and see if you think there should be any tweaks (either for consistency with the MCP server, or that you think would make it clearer to an LLM)?

DanTup:
@jakemac53 I was testing restarting DTD to see how it's handled. It appears that the result of the call to this tool is being cached, so when I ask for runtime errors after restarting DTD, it just reuses the old URI and fails. If I ask it to try again, it works:

image

I wonder if the failure error for "Connect to DTD" could be improved, perhaps hinting that the DTD URI might be stale and the "current" URI provided, to see if that'll cause it to fetch it immediately rather than failing?

(if you want me to file this somewhere, lmk where)

DanTup:
Merging this (behind the experiment flag) and we can follow up with any tweaks in separate commits.

PR #5680, opened 2025-09-02
For testing: Add code for authentication from the DevTools iframe t…

keertip:
…o enable access to web socket connections for VM Service and DTD.

- These changes were tested in FS and was able to open DevTools all the tries. No issues with VM Service connection or the DTD connection
- Did not remove the delay you added before testing.

### DIFF
--- a/src/extension/views/devtools/base_view_provider.ts
+++ b/src/extension/views/devtools/base_view_provider.ts
@@ -217,12 +217,17 @@ export abstract class MySimpleBaseWebViewProvider extends MyBaseWebViewProvider 
 	abstract get pageRoute(): string;
 
 	get pageUrls(): Promise<WebViewUrls | undefined> {
-		return this.devTools.urlFor(this.pageRoute).then((url) =>
-			url
-				? {
-					viewUrl: url,
-					authUrls: undefined,
-				}
-				: undefined);
+		return this.getPageUrls();
+	}
+
+	private async getPageUrls(): Promise<WebViewUrls | undefined> {
+		const url = await this.devTools.urlFor(this.pageRoute);
+		if (!url)
+			return undefined;
+
+		return {
+			viewUrl: url,
+			authUrls: undefined,
+		};
 	}
 }
--- a/src/extension/views/devtools/base_view_provider.ts
+++ b/src/extension/views/devtools/base_view_provider.ts
@@ -1,7 +1,10 @@
 import * as vs from "vscode";
 import { DartCapabilities } from "../../../shared/capabilities/dart";
 import { disposeAll } from "../../../shared/utils";
+import { DartDebugSessionInformation } from "../../../shared/vscode/interfaces";
 import { envUtils } from "../../../shared/vscode/utils";
+import { isFirebaseStudio } from "../../../shared/vscode/utils_cloud";
+import { debugSessions } from "../../commands/debug";
 import { perSessionWebviewStateKey } from "../../extension";
 import { DevToolsManager } from "../../sdk/dev_tools/manager";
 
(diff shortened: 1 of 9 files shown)


src/extension/views/devtools/base_view_provider.ts
```diff
@@ -17,6 +20,25 @@ export abstract class MyBaseWebViewProvider implements vs.WebviewViewProvider {
...
+		url = `https://${url}`;
+		return url;
+
+	}
+
+	protected get currentSession(): DartDebugSessionInformation | undefined {
+		// Get the first debug session if present. Should we get the current session instead? 
+		// Just using the first one works with opening devTools, even when debugging a subsequent session.
+		return debugSessions && debugSessions.length ? debugSessions[0] : undefined;
+	}
```
DanTup: I think we should be able to pass through the same VM Service URL that is going on the queryString so we don't have to worry about the possibility of getting the wrong one here. It might need a little refactoring though (so that where we currently just pass around a URL, we pass around something like a `{ viewUrl: string, vmServiceUrl?: string}`). I will take look and if it works push changes for you to review.

DanTup:
I've pushed some changes. It ended up being more than I'd planned because once I made the set of auto-requiring URLs an array, I needed to create the iframes dynamically, but I think this will make it easier if we need to add more in future (I also expanded it to ensure we we support this for all of the different iframes, whether it's DevTools sidebar, DevTools editor tabs, Property Editor, etc.).

I still need to do some proper testing on FS as so far my testing was all locally with the flags to trigger this hard-coded.

DanTup:
/gemini review
@codex review

src/extension/views/shared.ts
```diff
@@ -0,0 +1,77 @@
...
+	};
+}
+
+/// Given a string like wss://12345-firebase-test-connect-1234567.cluster-abcdef.cloudworkstations.dev/iN6zXu7VpYc=/ws
+/// returns a HTTP(S) url like https://12345-firebase-test-connect-1234567.cluster-abcdef.cloudworkstations.dev/iN6zXu7VpYc=/ws
+/// that can be used inside a hidden iframe to trigger auth cookies so that client apps like DevTools can connect over WebSockets.
+export function computeAuthFrameUri(urlString: string): string {
+	const uri = URL.parse(urlString);
+	if (!uri) return urlString; // We can't map it if it doesn't parse, just pass it through.
+
```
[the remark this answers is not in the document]
DanTup: Looks like [`URL.parse`](https://nodejs.org/api/url.html#urlparseinput-base) has only existed since NPM v22.1.0 (VS Code 1.101 from May 2025). Since we claim to support v1.75.0 we should change this.

DanTup:
@keertip I've done some testing of this PR with my tweaks both on Firebase Studio and locally, and I _think_ everything is working as expected. I was unable to trigger any errors on Firebase Studio.

I have not yet tried removing the startup delay - I suspect we don't need it, but I think we should try to land this as-is first, and then do some testing of that separately.

It would be good to also have you re-test since I think you were triggering the issue more than I was. If you don't want to build locally, there's a preview build from this PR branch here (`dart-code-3.120.0-dev.fs.2.vsix`):

https://drive.google.com/drive/u/0/folders/1PM66L6kN_OVzS9RXgJ0ceWtr46gT_yob

keertip:
@DanTup , changes look good! Will retest it today to confirm that it works.

src/extension/views/devtools/base_view_provider.ts
keertip: 👍

keertip:
Retested, and confirmed that it works as expected! Thanks for shepherding the change.

PR #5683, opened 2025-09-04
Remove service action delay for Firebase Studio

DanTup:
Now we're using iframes to force authentication, this may be unnecessary. In my testing, I'm not hitting any issues without it.


DanTup:
@keertip when you get chance, could you also test this? I'd like a second opinion before I merge, to ensure I don't make anything worse.

I've uploaded a `dart-code-3.120.0-dev.fs.3.vsix` here if that's more convenient than building it.

https://drive.google.com/drive/u/0/folders/1PM66L6kN_OVzS9RXgJ0ceWtr46gT_yob

keertip:
Thanks for this follow up, will test it out today.

keertip:
Tested, and confirmed that it works as expected.

PR #5741, opened 2025-10-07
Fix: Refactored test marker logic to rely on tracked position

### LINKED ISSUE #5739: Reconsider removing "potentially deleted" tests once using new test tracking by default
There were two different issues reported in https://github.com/Dart-Code/Dart-Code/issues/5734:

1. The test markers for tests that aren't run disappear
2. The test markers come back when modifying the file

I'm using that issue to track the second, but this is to track the first. It's actually by-design as the code currently is. Without the new tracking, we don't know if the test was not in the results because a) you stopped running early (or used `@soloTest`) or b) you had deleted the test.

However, with the new test tracking, we could perhaps remove the functionality of deleting "potentially deleted" nodes and instead just rely on the tracking to handle it (remove the node when the tracked position goes away).

It will be easier to do this when we're just using the new tracking by default, so bumping this until then,

abhayiscoding:
Fixed #5739 
- Modified the logic of removeAllPotentiallyDeletedNodes() to skip cleanup when using new tracking.
- Alternatively modified flagSuiteStart() to not mark nodes as potentially deleted when using experimental tracking.
- File: src\shared\test\test_model.ts

### DIFF
--- a/src/shared/test/test_model.ts
+++ b/src/shared/test/test_model.ts
@@ -187,7 +187,8 @@ export class TestModel {
 		suite.getAllGroups().forEach((g) => g.isStale = true);
 		suite.getAllTests().forEach((t) => t.isStale = true);
 
-		if (isRunningWholeSuite && suite) {
+		// not mark nodes as potentially deleted when using experimental tracking:
+		if (isRunningWholeSuite && suite && !this.config.experimentalTestTracking) {
 			this.markAllAsPotentiallyDeleted(suite, TestSource.Result);
 		}
 	}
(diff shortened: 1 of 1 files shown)


abhayiscoding:
Submitted a PR to address this: #5741  
Let me know if any changes are needed!

DanTup:
Thanks! I think we need to ensure there are good tests for this though. There are existing tests that will cover some of this once this behaviour is default, but I think we'll need to expand the tests that use test_reflective_loader to ensure we tests are deleted correctly if you do delete them (and not if you just ran a solo test).

We could add those tests now and just also guard them with this check, but once the new tracking is always enabled, it would (should) allow us to delete a bunch of code related to this tracking of potentially-deleted nodes.

DanTup:
I started looking at this, and it turns out removing this breaks some other scenarios, so for now I'm going to keep this as-is. Details are in https://github.com/Dart-Code/Dart-Code/issues/5739#issuecomment-3607155758

Thanks for the PR anyway!

Issue #5745, opened 2025-10-09
Cannot cancel a batch of "pub get" / "pub upgrade" / "flutter clean" commands


DanTup:
Slightly related to #5740 - if you try to cancel the "pub get" when it's running for all projects, it only cancels the one and then continues with the next.

We should fix things in a way that let you cancel the overall job.

Issue #5740, opened 2025-10-07
Vs code dart extension sends repeated notifications after clicking  `pub upgrade` on flutter/packages.


reidbaker:
**Describe the bug**

**To Reproduce**
1. open flutter/packages/packages in vscode 
2. click `pub upgrade` 
3. See a notification appear for every package as it upgrades. 

**Expected behavior**
Any of the following design patterns 
* One notification the whole time
* One notification the whole time with progress bar or text 
* One notification at the end
* No notification 

**Video**

**Please complete the following information:**

You can run the **Dart: Collect Diagnostic Information** command from the VS Code command palette (`F1`) to easily capture this information or provide it manually.

<details>

<summary><strong>Workspace Environment</strong></summary>

```text
Dart Code extension: 3.120.0
Flutter extension: 3.120.0 (activated)

App: Visual Studio Code
App Host: desktop
Version: mac 1.104.3

Workspace type: Dart, Flutter (LSP)
Workspace name: packages

Dart (3.10.0-268.0.dev): /Users/reidbaker/flutter-work/bin/cache/dart-sdk
Flutter (3.37.0-1.0.pre-386): /Users/reidbaker/flutter-work (macOS (darwin/macos))

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`/Users/reidbaker/flutter-work/bin/cache/dart-sdk/bin/dart info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.10.0-268.0.dev (dev) (Wed Oct 1 09:03:20 2025 -0700) on "macos_arm64"
- on macos / Version 15.7 (Build 24G222)
- locale is en

#### Process info

|   Memory |  CPU | Elapsed time | Command line                                                                     |
| -------: | ---: | -----------: | -------------------------------------------------------------------------------- |
|    34 MB | 0.0% |        01:59 | dart devtools --machine --allow-embedding --dtd-uri ws:<path>/PJfjwGEs7JM=       |
| 12034 MB | 0.0% |        01:59 | dart language-server --protocol=lsp --client-id=VS-Code --client-version=3.120.0 |
|    19 MB | 0.0% |        01:59 | dart tooling-daemon --machine                                                    |
|    79 MB | 0.9% |        01:59 | flutter_tools.snapshot daemon                                                    |

</details>

<details>

<summary><strong>Output from 'flutter doctor'</strong></summary>

[...]

PR #5746, opened 2025-10-09
Improve batch execution of commands

DanTup:
Improves "pub get", "pub upgrade" and "flutter clean" when run in-batch. We now produce a single notification and report the names of the projects as we go through them. This also means that cancellation works correctly.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5745
Fixes https://github.com/Dart-Code/Dart-Code/issues/5740


DanTup:
@codex review
/gemini review

PR #5755, opened 2025-10-17
Show a one-off warning about onlyAnalyzeProjectsWithOpenFiles

DanTup:
See https://github.com/dart-lang/sdk/issues/61742


DanTup:
@codex review
/gemini review

DanTup:
@bwilkerson @jwren any opinions on the text for this notification? (it would be shown only to users with the setting enabled, and once they click the button to open the issue or dismiss the notification, we would not show it again).

Image

The button opens https://github.com/dart-lang/sdk/issues/61742 which has some details about why I think we should deprecate this (TL;DR, it makes initial analysis faster, but can cause long delays as we reset analysis roots as the first/last file in any project is opened/closed during normal use).

bwilkerson:
That message looks good to me.

If I were going to change anything I might suggest removing "you think", just because I don't think it adds to the meaning, but I'm also fine with leaving it in.

DanTup:
> If I were going to change anything I might suggest removing "you think"

Yep, I agree - done. Thanks!

Issue #5549, opened 2025-06-16
New Version always displays notification to perform unnecessary Run 'pub get'


longtimedeveloper:
**Describe the bug**
Each time I open a Flutter project, the extension always displays a pop up notification that says, "Some packages are mission or out of date, would you like to get them now? Button: "Run 'pub get'". Even if I run flutter pub get command, then close VSCode, and then reopen the project, this pop up appears on project open, but shouldn't

**To Reproduce**
1. Open VSCode
2. Using menu, open a recent project.

**Expected behavior**
No pop up notification if the project is actually up to date.

**Screenshots**
[Pop Up Notification](https://raw.githubusercontent.com/longtimedeveloper/puzzle_images/refs/heads/main/bug.JPG)

**Please complete the following information:**

<details>

<summary><strong>Workspace Environment</strong></summary>

```text
Dart Code extension: 3.112.0
Flutter extension: 3.112.0 (not activated)

App: Visual Studio Code
App Host: desktop
Version: win 1.100.3

Workspace type: Dart (LSP)
Workspace name: rock_shared

Dart (3.8.1): D:\flutter\bin\cache\dart-sdk
Flutter (3.32.3): D:\flutter (No device)

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`D:\flutter\bin\cache\dart-sdk\bin\dart.exe info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.8.1 (stable) (Wed May 28 00:47:25 2025 -0700) on "windows_x64"
- on windows / "Windows 10 Pro" 10.0 (Build 19045)
- locale is en-US

#### Project info

- sdk constraint: '^3.8.1'
- dependencies: json_annotation
- dev_dependencies: build_runner, json_serializable, lints

#### Process info

| Memory | CPU | Elapsed time | Command line |
| -----: | --: | -----------: | ------------ |
|  22 MB |  -- |              | dart.exe     |
| 486 MB |  -- |              | dart.exe     |
|  90 MB |  -- |              | dart.exe     |
|  99 MB |  -- |              | dart.exe     |
|  39 MB |  -- |              | dart.exe     |
| 381 MB |  -- |              | dart.exe     |
|  90 MB |  -- |              | dart.exe     |

</details>

DanTup:
[...]

PR #5760, opened 2025-10-20
Update pubspec.lock/package_config.json after running "pub get" as a workaround

DanTup:
... for `pub` not updating these.

This fixes https://github.com/Dart-Code/Dart-Code/issues/5549 but once https://github.com/dart-lang/pub/issues/4672 is fixed we should change the upper bound so we stop doing it for newer SDKs, because it's a bit hacky.


DanTup:
@codex review
/gemini review

PR #5769, opened 2025-10-28
Fix coverage false positives for structural lines like } finally {

tcf909:
## Description

This PR fixes false positive coverage reporting for structural lines like `} finally {` that are incorrectly shown as uncovered in VS Code's coverage visualization.

## Problem

The Dart VM incorrectly generates branch coverage entries (BRDA) for structural lines like `} finally {` without corresponding line coverage entries (DA). This causes the extension to mark these lines as coverable but uncovered, resulting in:
- Red gutter markers on structural lines
- Coverage showing < 100% even when all executable code is covered

### Example

```dart
try {
  // code
} finally {  // ← Line incorrectly shown as uncovered
  cleanup();
}
```

**lcov.info data:**
```
DA:227,1    (line before finally - covered)
BRDA:229,0,0,0  (} finally { - spurious entry!)
DA:230,1    (first line in finally - covered)
```

## Solution

Implements a two-pass parsing approach that filters out spurious BRDA entries:
1. **First pass**: Collect all lines with DA (direct coverage) entries
2. **Second pass**: Process BRDA entries, only including them if:
   - The branch was taken (`taken > 0`), OR
   - There's a corresponding DA entry for that line

This prevents structural lines without DA entries from being marked as coverable.

## Validation

This approach is validated by **the official LCOV tool itself**, which rejects BRDA entries without corresponding DA entries as "inconsistent data":

```perl
# From lcovutil.pm:7575-7585
foreach my $line ($brData->keylist()) {
    # we expect to find a line everywhere there is a branch
    my $lineHit = $lineData->value($line);
    unless (defined($lineHit)) {
        lcovutil::ignorable_error($lcovutil::ERROR_INCONSISTENT_DATA,
              "location has branchcov but no linecov data");
    }
```

Our implementation aligns with this standard by filtering such entries rather than reporting them as coverage gaps.

## Testing

- ✅ Existing test updated to reflect new behavior
- ✅ New test added specifically for the `} finally {` false positive case
- ✅ All tests passing

## Impact

- **Fixes false negatives**: Structural lines no longer show as uncovered
- **Maintains accuracy**: Legitimate uncovered code (with DA entries) still shows correctly  
- **No breaking changes**: Only filters obviously spurious entries

## Related Issues

Similar to dart-lang/sdk#31222 which fixed async function closing brace coverage issues.

### DIFF
--- a/src/shared/test/coverage.ts
+++ b/src/shared/test/coverage.ts
@@ -27,6 +27,8 @@ export class CoverageParser {
 		let sourceFilePath: string | undefined;
 		const coverableLines = new Set<number>();
 		const coveredLines = new Set<number>();
+		const linesWithDirectCoverage = new Set<number>(); // Track lines that have DA entries
+		const branchEntries: Array<{ lineNumber: number, taken: number }> = []; // Store BRDA entries for second pass
 
 		function recordLine(lineNumber: number, taken: number) {
 			coverableLines.add(lineNumber);
@@ -35,6 +37,7 @@ export class CoverageParser {
 			}
 		}
 
+		// First pass: collect all lines and identify which have direct coverage (DA) entries
 		for (const line of record.split("\n").map((l) => l.trim()).filter((l) => l !== "" && !l.startsWith("#"))) {
 			const fields = line.split(":");
 			if (fields.length !== 2) {
@@ -59,6 +62,7 @@ export class CoverageParser {
 						const valueParts = value.split(",").map((p) => p.trim());
 						const lineNumber = parseInt(valueParts[0], 10);
 						const taken = parseInt(valueParts[1], 10);
+						linesWithDirectCoverage.add(lineNumber);
 						recordLine(lineNumber, taken);
 					}
 					break;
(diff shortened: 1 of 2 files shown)


DanTup:
Thank you for the PR! Although, I think if the data we're getting is bad that it would be better to try and improve that in the VM or `package:coverage`.

@liamappelbe is this an issue you are aware of? Do you know if it's an issue in the VM or `package:coverage`? My preference would be to avoid any massaging of the data here because it will get complicated if these assumptions become incorrect (since we need to support multiple SDK/coverage versions in Dart-Code).

liamappelbe:
We often have issues like this in the coverage collection infra. All I can do is fix them one by one as they're reported.

@tcf909 Can you file a bug in the [package:coverage repo](https://github.com/dart-lang/tools/issues)? Is `finally` the only case you've found, or were there others?

DanTup:
I've opened https://github.com/dart-lang/tools/issues/2241 with the description from above.

I'll close this on the assumption that it will be fixed in `pkg:coverage`. We can revisit landing a workaround if that turns out not to be the case for any reason.

@tcf909 if you know of any other places that have this issue (besides finally), please note them on https://github.com/dart-lang/tools/issues/2241 (or file other issues there if they're not directly related).

Thanks!

Issue #5794, opened 2025-11-12
The Dart Tooling Daemon connection has terminated.


SenexTech:
**Describe the bug**
When you build any app using Flutter, in my case I created a new app using the empty app template that comes with Flutter/VS Code.  So to clarify if you create a new app using VS Code (View/Command Pallete/Empty Application) then select Run the app will build successfully and run on my real Android phone. It could be a few minutes latter that the app exits and displays this error. This happens on two different Android phones and two different USB cables. I'm also getting the error 'Unhandled exception:
FileSystemException: writeFrom failed, path = '' (OS Error: The pipe is being closed. , errno = 232)' which may be caused by the daemon exiting?

**To Reproduce**
1. create a new app using VS Code (View/Command Pallete/Empty Application)
2. Select Run the app (start debuging) and theapp will build successfully and run on my real Android phone.
3. Some time latter (could be a couple ofminutes)  I get a popup message saying "The Dart Tooling Daemon connection has terminated."
.

**Expected behavior**
The app should continue to run in debug mode without any errors and stay connected to device

**Please complete the following information:**

You can run the **Dart: Collect Diagnostic Information** command from the VS Code command palette (`F1`) to easily capture this information or provide it manually.

```text
Dart Code extension: 3.122.0
Flutter extension: 3.122.0 (activated)

App: Visual Studio Code
App Host: desktop
Version: win 1.105.1

Workspace type: Flutter (LSP)
Workspace name: flutter_application_1

Dart (3.9.2): C:\Users\pcira\dev\flutter\bin\cache\dart-sdk
Flutter (3.35.7): C:\Users\pcira\dev\flutter (Pixel 7 Pro (android-arm64/android))

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`C:\Users\pcira\dev\flutter\bin\cache\dart-sdk\bin\dart.exe info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.9.2 (stable) (Wed Aug 27 03:49:40 2025 -0700) on "windows_x64"
- on windows / "Windows 11 Pro" 10.0 (Build 26200)
- locale is en-US

#### Project info

- sdk constraint: '^3.9.2'
- dependencies: flutter
- dev_dependencies: flutter_lints, flutter_test

#### Process info

| Memory | CPU | Elapsed time | Command line |
| -----: | --: | -----------: | ------------ |
[...]

PR #5828, opened 2025-12-02
Ping over the DTD WebSocket every 15s to avoid some proxies (like Norton) dropping the connection

DanTup:
Users with Norton 360 installed found that if there is no traffic over the WebSocket for 60s, it gets dropped. This appears to be that the connections are proxied (even if all scanning is disabled) and "idle" connections are dropped.

This forces traffic ever 15s which will prevent them from dropping the connection.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5794


DanTup:
@codex review
/gemini review

DanTup:
@codex review
/gemini review

Issue #5831, opened 2025-12-03
Deleting a test results in the next test marker moving to the wrong line


DanTup:
When I delete this test, the next test getter icon moves to the wrong line.

https://github.com/user-attachments/assets/8f305440-e50a-4ab3-8294-a4864285f4d9

FMorschel:
Typo (title): marker*

FMorschel:
Run the following test file and run `Dart: Sort Members`. See a similar behaviour. Is this the exact same?

```dart
import 'package:test/test.dart';
import 'package:test_reflective_loader/test_reflective_loader.dart';

void main() {
  defineReflectiveSuite(() {
    defineReflectiveTests(ATEst);
  });
}

@reflectiveTest
class ATEst {
  void test_example() {
    // Example test
    expect(1 + 1, 2);
  }

  void test_anotherExample() {
    // Another example test
    // this
    // one
    // is
    // longer
    expect('hello'.toUpperCase(), 'HELLO');
  }

  void test_yetAnotherExample() {
    // Yet another example test
    expect([1, 2, 3].length, 3);
  }
}
```

Note that it doesn't happen if all tests are the same line length.

DanTup:
With my fix, I end up with this:

<img width="706" height="657" alt="Image" src="https://github.com/user-attachments/assets/90f99c9c-d0fa-4991-b6f2-f1d0f64b9e75" />

The top two get swapped around, and their positions removed.

The bug was that when a tracked range (which is a start and end position) was deleted, we first deleted the start position (internally, range tracker is just two position trackers), which caused the first range tracker to dispose itself. This reduced the number of position trackers, but it was done in a callback during the processing, so the code looping over the position trackers skipped one (because the collection was modified).

This resulted in only the _end_ position of the second range being changed (resulting in a range where the start was before the end!) which caused weird things to happen.

PR #5834, opened 2025-12-04
Fix test tracking when ranges are deleted

DanTup:
When processing changes, we were firing the callbacks in-line while enumerating the trackers. A deleted position could result in a tracked range from disposing itself (the start position is deleted first, and it then disposes and untracks the end position) which modified the list of trackers while we were enumerating.

To fix, we store up the callbacks so nothing is modified while we synchronously process the changes.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5831


DanTup:
@codex review

/gemini review

Issue #5909, opened 2026-02-17
The server doesn't start if flutter (maybe dart too?) root is an invalid path


FMorschel:
**Describe the bug**
Once a project has run `pub get` on a different operating system, for example (shared partition), the server never starts.

**To Reproduce**

Mostly a guess, but I think if you have WSL (so no dualboot needed) and Dart installed there. If you run `pub get` on a project from there and then try opening that workspace folder from Windows, you'll see this behaviour.

**Expected behavior**
For the `Run pub get` toast to show up.

**Please complete the following information:**

<details>

<summary><strong>Workspace Environment</strong></summary>

```text
Dart Code extension: 3.129.20260202
Flutter extension: 3.129.20260202 (activated)

App: Visual Studio Code
App Host: desktop
Version: win 1.109.4

Workspace type: Dart (LSP)
Workspace name: essential_lints

Dart (3.12.0-edge.442bafa30af80793a5c638f1b56d5d9ccb3cc16b): C:\Users\felip_0vh5fa6\dart
Flutter (3.38.7): C:\Users\felip_0vh5fa6\.puro\envs\stable\flutter (No device)

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`C:\Users\felip_0vh5fa6\dart\bin\dart.exe info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.12.0-edge.442bafa30af80793a5c638f1b56d5d9ccb3cc16b (main) (Tue Feb 10 06:53:30 2026 -0800) on "windows_x64"
- on windows / "Windows 11 Pro" 10.0 (Build 26200)
- locale is pt-BR

#### Project info

- sdk constraint: '^3.10.0'
- dependencies: analysis_server_plugin, analyzer, essential_lints_annotations, logging, meta
- dev_dependencies: analyzer_testing, frontend_server_client, test, test_reflective_loader, very_good_analysis

#### Process info

| Memory | CPU | Elapsed time | Command line |
| -----: | --: | -----------: | ------------ |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
|   0 MB |  -- |              | dart.exe     |
[...]

PR #5910, opened 2026-02-17
Don't fail with invalid paths in the package map

DanTup:
If you did `pub get` on Linux and get that package_config into Windows, this would throw and stop the extension from starting up.

Fixes https://github.com/Dart-Code/Dart-Code/issues/5909


DanTup:
@codex review

/gemini review

Issue #5954, opened 2026-03-04
ENOTDIR: not a directory, stat '/path/to/project/lib/main.dart/pubspec.yaml'


solace6282:
**Describe the bug**
Flutter extension seem to look for `pubspec.yaml` from file directory given in "program" on `launch.json`'s `configurations` and because "program" points to a a file (i.e. `main.dart`), debug session crashes. Debugging was okay yesterday, probably has something to do with version 3.130.0

**To Reproduce**
1. Open or create any project
2. Create folder `.vscode`
3. Create file `launch.json`, use any configurations
4. Start Debugging (F5)

**Expected behavior**
Debugging should start

**Screenshots**

<img width="1850" height="1029" alt="Image" src="https://github.com/user-attachments/assets/f81d1b24-fcbe-442f-80bf-6397b0f1fd1f" />

**Please complete the following information:**

You can run the **Dart: Collect Diagnostic Information** command from the VS Code command palette (`F1`) to easily capture this information or provide it manually.

 - Operating System and version: Ubuntu 22.04
 - VS Code version: linux 1.99.2
 - Dart extension version: 3.130.0
 - Dart/Flutter SDK version: 3.11.1/3.41.3
 - Target device: Android

solace6282:
Update: Tried downgrading to 3.128.0 and debugging works fine.

GP4cK:
Same issue on MacOS 15.7.4

DanTup:
I don't seem to be able to reproduce this. Is someone able to:

1. Check Help -> Toggle Developer Tools and see if there is a stack trace for this error in the console?
2. Run the **Dart: Capture Debugging Logs** command just before reproducing the issue (then click **Cancel** in the logging notification to stop logging and open the log file)

Can you also confirm _exactly_ what's in the `launch.json` (I was trying with the default one that's created) and confirm that it happens in a newly created project (that's what I'm testing with),

Here's the steps I tried:

- **Flutter: New Project** -> Application
- Click debug sidebar, then **Create launch.json**
- Press `F5` to start debugging

tortamque:
Same here. Downgrading extension to 3.128 solved the issue

oscarpalomar:
Using pre-release version 3.129.20260224 works.
Using pre-release version 3.129.20260226 fails.

solace6282:
> I don't seem to be able to reproduce this. Is someone able to:
> 
[...]

PR #5958, opened 2026-03-04
Catch exceptions when calling fs.statSync()

DanTup:
See https://github.com/Dart-Code/Dart-Code/issues/5954

Some users see an unhandled ENOTDIR error if we end up calling `fs.statSync()` with a path that is a sub-path of a file (which happens because of how we call some of these helpers, assuming they can be passed either files or folders).

I cannot repro this on Win, macOS or Ubuntu locally, nor on the bots, so we'll have to push a pre-release and have someone verify the fix before pushing a stable release.


DanTup:
@codex review

/gemini review

Issue #4617, opened 2023-06-28
Add a test that checks any links in code/constants are valid


DanTup:
See https://github.com/Dart-Code/Dart-Code/issues/4604#issuecomment-1607976296

We could grep all valid hyperlinks out of the `.ts` files, and then check they're valid (I'm not sure how easily we can check fragments). This would help us if any links to the Flutter website etc. changed over time.

queenfbaby979:
#5216

queenfbaby979:
#2255

queenfbaby979:
#5216 ``

Peter080202:
Hey @DanTup, 
I think this issue is still open. A good idea is to create a check in a GitHub action (probably `.github/workflows/build-and-test.yml`) which checks all links, will create a PR later today/tomorrow!

PR #5965, opened 2026-03-06
Add GitHub action to check links with Lychee

Peter080202:
### Description

Introduced `link_check` workflow to automatically verify links in source code, Markdown files and package.json.

**Issue link**: https://github.com/Dart-Code/Dart-Code/issues/4617

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -298,15 +298,14 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Run Lychee Link Checker
-        id: lychee
         uses: lycheeverse/lychee-action@v2
         with:
           args: >
             --root-dir ./
             --exclude-all-private
             --verbose
+            --exclude '\$\%7B[^}]+\%7D'
             --exclude "^file://"
-            --exclude "$%7B.*%7D"
             ./src/**/*.ts
             ./*.md
             ./**/*.license


Peter080202:
@codex review

/gemini review

Peter080202:
@DanTup could you review this one please?

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,43 @@ jobs:
           include-hidden-files: true
           if-no-files-found: ignore
 
+  link_check:
+    name: Check Links (Lychee)
+    runs-on: ubuntu-latest
+    needs: build
```
DanTup: We probably don't need to wait on the build for this, we could just let it run at the same time to speed up the overall completion.

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,43 @@ jobs:
...
+
+    permissions:
+      contents: read
+      issues: write
```
DanTup: I don't think `issue: write` will do anything on its own, it seems to require another action.

My preference would be just to drop this and not create the issues, since I get notifications of build failures anyway.

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,43 @@ jobs:
...
+      - name: Set up Node v${{ matrix.node-version }}
+        uses: actions/setup-node@v6
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: npm install
+        run: npm install
+
+      - name: npm build
+        run: npm run build
```
DanTup: Do we need `npm install` or `npm build`? If it's just scanning local files, I don't think the results of either should matter.

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,43 @@ jobs:
...
+      - name: Run Lychee Link Checker
+        id: lychee
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: >
+            --root-dir ./
+            --exclude-all-private
+            --verbose
+            ./src/**/*.ts
+            ./src/**/*.tsx
```
DanTup: `tsx` is unnecessary since we don't have them, but we might want to consider:

- `.yml` (for ex. we have links in `.github/ISSUE_TEMPLATE/config.yml`)
- `.licence` (for ex. there's a link in `syntaxes\dart.json.license`)

DanTup:
@Peter080202 I've pushed some changes to master that fix some of the failures, but there are two remaining that are URLs with variables in them:

- `return `file:///${encodeURI(file)}`;`
- `await envUtils.openInBrowser(`https://dartcode.org/releases/v${versionLink}/`);`

I think maybe they could be handled with something like `--exclude "://.*$"`?

.github/workflows/build-and-test.yml
Peter080202: Sure

.github/workflows/build-and-test.yml
Peter080202: Makes sense

.github/workflows/build-and-test.yml
Peter080202: Should be safe to remove yes

.github/workflows/build-and-test.yml
Peter080202: Okidoki added them

.lycheeignore
```diff
@@ -0,0 +1,5 @@
+# Ignore all local file URLs
+^file://
+
+# Ignore URLs containing template variables like ${versionLink}
+\$\{.*\}
```
Peter080202: Reference:
https://lychee.cli.rs/recipes/excluding-links/#permanently-excluding-links

I also added `^file://` since I don't think it makes much sense to check local file links

DanTup: This makes sense to me, but would it work as a CLI flag instead? If possible, I'd rather avoid cluttering the repo with another file if it could be kept together with the command.

Thanks!

Peter080202: Yes that works as well! Changed it and also adapted the pattern to exclude URLs containing variables to `$%7B.*%7D` since Lychee parses the literal string in the source code
`https://dartcode.org/releases/v${versionLink}/`
becomes
`https://dartcode.org/releases/v$%7BversionLink%7D/`

DanTup: Thanks! It looks like the comments in the command might be messing things up:

```
    INPUT_ARGS: --root-dir ./ --exclude-all-private --verbose --exclude "^file://"  # Ignore all local file URLs --exclude "$%7B.*%7D" # Ignore URLs containing template variables like ${versionLink} ./src/**/*.ts ./*.md ./**/*.license package.json


error: the following required arguments were not provided:
  <inputs>...

Usage: lychee --mode <MODE> --format <FORMAT> --output <OUTPUT> --root-dir <ROOT_DIR> --exclude-all-private --verbose... --exclude <EXCLUDE> --github-token <GITHUB_TOKEN> <inputs>...

For more information, try '--help'.
```

So we might need to drop the comments (or move them outside of the args).

Peter080202: Ah shit now it aint working at all 🫠

DanTup: Yeah, see my comment above, I think it's just the comments :)

Peter080202: Yes removed them now

DanTup:
Thanks!

Looks like it's still complaining about this:

```
[404] https://dartcode.org/releases/v$%7BversionLink%7D/ | Rejected status code: 404 Not Found (configurable with "accept" option)
```

I'm not sure why, the regex looks reasonable to me. If you don't want to spend more time on this, I'm happy to debug it further later or tomorrow :-)

Thanks again!

DanTup:
@codex review

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,32 @@ jobs:
...
+          args: >
+            --root-dir ./
+            --exclude-all-private
+            --verbose
+            --exclude "^file://"
+            --exclude "$%7B.*%7D"
+            ./src/**/*.ts
+            ./*.md
+            ./**/*.license
+            package.json
```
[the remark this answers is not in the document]
Peter080202: Makes sense, I adapted it again and should work now (tested locally and doesn't show up anymore as error)

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,32 @@ jobs:
...
+
+    permissions:
+      contents: read
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Run Lychee Link Checker
+        id: lychee
```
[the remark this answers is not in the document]
Peter080202: Makes sense

.github/workflows/build-and-test.yml
```diff
@@ -286,6 +286,32 @@ jobs:
...
+        uses: lycheeverse/lychee-action@v2
+        with:
+          args: >
+            --root-dir ./
+            --exclude-all-private
+            --verbose
+            --exclude "^file://"
+            --exclude "$%7B.*%7D"
```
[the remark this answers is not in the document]
Peter080202: Adapted regex

Peter080202:
@DanTup it should work now 🤞

Peter080202:
🎉 
```bash
| Status         | Count |
|----------------|-------|
| 🔍 Total       | 119   |
| ✅ Successful  | 96    |
| ⏳ Timeouts    | 0     |
| 🔀 Redirected  | 16    |
| 👻 Excluded    | 5     |
| ❓ Unknown     | 0     |
| 🚫 Errors      | 0     |
| ⛔ Unsupported | 2     |
```

DanTup (approved):

DanTup:
@Peter080202 perfect, thank you!

PR #6059, opened 2026-05-18
Add initial support for new interactive refactorings

h9jiang:
DO NOT REVIEW

This is an experimental CL that tries to enable interactive command execution from the dart-code.

For golang/go#78553

### DIFF
--- a/src/extension/analysis/analyzer.ts
+++ b/src/extension/analysis/analyzer.ts
@@ -25,7 +25,6 @@ import { envUtils, hostKind, isRunningLocally } from "../../shared/vscode/utils"
 import { WorkspaceContext } from "../../shared/workspace";
 import { Analytics } from "../analytics";
 import { config } from "../config";
-import { AddDependencyCodeActionProvider } from "../providers/add_dependency_code_action_provider";
 import { checkForLargeNumberOfTodos } from "../user_prompts";
 import { promptToReloadExtension } from "../utils";
 import { reportAnalyzerTerminatedWithError } from "../utils/misc";
@@ -34,6 +33,10 @@ import { getDiagnosticErrorCode } from "../utils/vscode/diagnostics";
 import { SnippetTextEditFeature } from "./analyzer_snippet_text_edits";
 import { LegacyRefactors } from "./features/legacy_refactors";
 import { FileTracker } from "./file_tracker";
+import {
+	InteractiveLanguageClient,
+	InteractiveMiddleware
+} from "./form";
 
 // Globals so we only show these errors once per session.
 let hasShownAnalysisServerVersionMismatchError = false;
@@ -63,7 +66,6 @@ export class LspAnalyzer extends Analyzer {
 
 		// Register all language client features.
 		this.client.registerFeature(new CommonCapabilitiesFeature().feature);
-		this.client.registerFeature(new AddDependencyCodeActionProvider(this.client).feature);
 		this.client.registerFeature(new LegacyRefactors(this.logger, this.client).feature);
 		this.client.registerFeature(this.refactors.feature);
 		this.client.registerFeature(this.snippetTextEdits.feature);
@@ -121,7 +123,7 @@ export class LspAnalyzer extends Analyzer {
 		};
 	}
 
-	private buildMiddleware(): ls.Middleware {
+	private buildMiddleware(): InteractiveMiddleware {
 		// Why need this 🤷‍♂️?
 		function isLanguageValuePair(input: any): input is { language: string; value: string } {
 			return "language" in input && typeof input.language === "string" && "value" in input && typeof input.value === "string";
@@ -332,7 +334,7 @@ export class LspAnalyzer extends Analyzer {
 					return results;
 				},
 			},
-		};
+		} as InteractiveMiddleware;
 	}
 
 	public async getDiagnosticServerPort(): Promise<{ port: number } | undefined> {
@@ -389,7 +391,7 @@ export class LspAnalyzer extends Analyzer {
 		);
 	}
 
-	private createClient(logger: Logger, sdks: DartSdks, dartCapabilities: DartCapabilities, wsContext: WorkspaceContext, middleware: ls.Middleware): LanguageClient {
+	private createClient(logger: Logger, sdks: DartSdks, dartCapabilities: DartCapabilities, wsContext: WorkspaceContext, middleware: InteractiveMiddleware): InteractiveLanguageClient {
 		const converters = new LspUriConverters(!!config.normalizeFileCasing);
 		const clientOptions: ls.LanguageClientOptions = {
 			errorHandler: new DartErrorHandler(logger),
@@ -435,7 +437,7 @@ export class LspAnalyzer extends Analyzer {
 			},
 		};
 
-		const client = new LanguageClient(
+		const client = new InteractiveLanguageClient(
 			"dartAnalysisLSP",
 			"Dart Analysis Server",
 			async () => {
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -1,0 +1,1025 @@
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable @typescript-eslint/consistent-indexed-object-style */
+/* eslint-disable arrow-body-style */
+/* eslint-disable no-bitwise */
+/* eslint-disable object-shorthand */
+/* eslint-disable quotes */
+/* eslint-disable @typescript-eslint/no-unsafe-return */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+/* eslint-disable @typescript-eslint/prefer-function-type */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+/* ---------------------------------------------------------
+ * Copyright 2025 The Go Authors. All rights reserved.
+ * Licensed under the MIT License. See LICENSE in the project root for license information.
+ *--------------------------------------------------------*/
+
+import * as vscode from 'vscode';
+import {
+	ExecuteCommandSignature,
+	LanguageClient,
+	LanguageClientOptions,
+	Middleware,
+	RequestType,
+	ServerOptions
+} from 'vscode-languageclient/node';
+import { InitializeParams } from 'vscode-languageserver-protocol';
+
+// ----------------------------------------------------------------------------
+// Form Field Type Definitions
+// ----------------------------------------------------------------------------
+
+// TODO(hxjiang): extend the support for file input type as a subtype of string.
+
+// FormFieldTypeString defines a text input.
+export interface FormFieldTypeString {
+	kind: 'string';
+}
+
+// FileExistence whether the file denoted by a DocumentURI exists.
+//
+// It is a bit set allowing combinations of existence states. For
+// example, New|Existing allows either state.
+export enum FileExistence {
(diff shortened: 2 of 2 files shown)


DanTup:
@h9jiang did you want the bots running on this? (I expect everything would pass, but I don't know if that's what you were trying to test).

FWIW I had just started looking at the server side work for this, but was going to hold off client work in case you were already working on abstracting Go out of these files :)

h9jiang:
Hi Danny,

Thank you so much for your quick response.

> did you want the bots running on this?

Yes, I would like to try running the bots if possible.

> was going to hold off client work in case you were already working on abstracting Go out of these files

This change actually doesn't have any go related code, it is purely just language client code. But I'm still trying it out to see if it is compatible with dart-code.

DanTup:
> Yes, I would like to try running the bots if possible.

I've approved them. I don't know if it'll auto re-run if you push changes, but feel free to ping me if they need running again (or if you'd like me to review anything). 

(I think I've also fixed Gemini reviewing draft PRs!)

DanTup:
@h9jiang Brian mentioned you planned on updating this PR soon with the latest changes. Just FYI, I think I have the Dart server changes done now, so if it's more convenient for you, I could apply your recent Go changes here tomorrow so I can test against the server.

Initially, I only need the capability changes etc., things like `required` handling are less important initially (mainly I want to test that I implemented the server side correctly to match the client code).

h9jiang:
@DanTup  Hi Danny, thank you so much. Because you are reviewing my changes in the vscode-go, I thought I will prepare the CL after your review. 

Here it is, could you give it a try? 

Just to be clear, most of the implementation aligns with the proposal, but the capability is being embedded inside of "experimental". I'm afraid some language client and server may have hard time handling the marshal/unmarshal. 

Let me know if you hit any problems or have any questions.

h9jiang:
In the same time, Gemini seems to have a few comments for me, I will read those and see if I can make Gemini happier. 

Thanks.

DanTup:
@h9jiang thanks! I'll do some testing of this tomorrow.

If I need to make changes, I might merge this PR into a new branch for this feature where we can iterate on it until I'm certain the server implementation all matches correctly. I'll let you know if I hit any issues.

src/extension/analysis/analyzer.ts
```diff
@@ -198,6 +206,21 @@ export class LspAnalyzer extends Analyzer {
...
+
+			interactiveExecuteCommand: async (
+				command: string,
+				args: any[],
+				formAnswers: any[],
+				next: InteractiveExecuteCommandSignature
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-return
+			) => await next(command, args, formAnswers),
+
+			interactiveListEnum: async (param: InteractiveListEnumParams, next: InteractiveListEnumSignature) => await next(param),
```
DanTup: What's the purpose of these middleware stubs? For the original client, middleware is only declared if it needs to do something different to the default, but these seem to just call `next()`. Are they necessary?

src/extension/analysis/analyzer.ts
```diff
@@ -537,7 +560,7 @@ export class LspAnalyzer extends Analyzer {
 			},
 		};
 
-		const client = new LanguageClient(
+		const client = new InteractiveLanguageClient(
```
DanTup, line 440: What are your thoughts on make this a "Feature" instead of a subclass? It might make it easier to combine with other features (if multiple customisations subclassed `LanguageClient`, it would be hard to use multiple of them).

I already use features for some of the things here, although there will still referenced directly in the middleware - however, I realised we can attach middleware in the constructors, so I've started pulling them out (to try and reduce the complexity in this class/file), for ex. https://github.com/Dart-Code/Dart-Code/pull/6071

src/extension/analysis/form.ts
```diff
@@ -0,0 +1,1012 @@
...
+					// we split the flow:
+					//
+					// - "Open Existing": Uses showOpenDialog (Clean UX, no warnings)
+					// - "Create New": Uses showSaveDialog (The "Overwrite" warning
+					// is unavoidable here, but users expect some friction when
+					// "creating" over an existing name, so it is acceptable).
+					const action = await vscode.window.showQuickPick(
+						[
+							{
+								label: '$(file) Open Existing File',
```
DanTup: What about "Select Existing File"? The word Open here felt a little odd when I invoked the existing "Move to File" refactor, because I'm selecting a destination for the class, but the file is opened in a background tab so it doesn't feel much like an "Open".

src/extension/analysis/form.ts
```diff
@@ -0,0 +1,1012 @@
...
+								description: 'Select a file that already exists',
```
DanTup: Should we replace "file" (here and in a few other places here) if the field is for directories?

src/extension/analysis/form.ts
```diff
@@ -0,0 +1,1012 @@
...
+					}
+
+					const uri = await vscode.window.showOpenDialog({
+						canSelectFiles: canSelectFiles,
+						canSelectFolders: canSelectFolders,
+						canSelectMany: false,
+						openLabel: 'Select',
+						defaultUri: defaultUri,
+						filters: canSelectFiles ? filters : undefined,
+						title: field.description || 'Select Existing File'
```
DanTup: Looks like we already use "Select" here, so using it above too would make these consitent.

DanTup (commented):
@h9jiang thanks for this! I had a quick test against the Dart server, and everything seems to work correctly!

I've left some minor comments/questions below I'm interested in your feedback on. I'm happy to make some of these changes myself if you agree with them (although I don't want to diverge from your own implementation if the goal is for `form.ts` to be shared).

src/extension/analysis/analyzer.ts
h9jiang: Ohh, they are not necessary at all, I wrote it here simply to demonstrate how other existing extension author can use this to create their middleware... I will remove them all.

src/extension/analysis/form.ts
h9jiang: Done.

src/extension/analysis/form.ts
h9jiang: Maybe "entry"? or any suggestions?

src/extension/analysis/analyzer.ts
h9jiang: I give it a try on the feature registration based solution and I think it is pretty promising. Just to make sure I understand, we still want to allow "extension authors" to specify their middleware for "interactive execute command", "resolve command" & "interactive list enum". Even after migrating to the feature registration based solution, we still need to give extension authors a place to specify these middleware, is that correct?

In addition, I just noticed that you are using the head version of vscode-languageclient and I'm left behind, so the StaticFeature have some backward incompatible changes. Maybe a simple rename from `dispose()` to `clear()`. But this is pretty small.

Since I'm not very familiar in this area, do you think you can help me create a new PR containing the diff and we can discuss in that CL? How user would use the feature registration based solution .... etc, after we get this merged. Compares to you, my understanding in the vscode side is pretty shallow. :D

h9jiang:
> @h9jiang thanks for this! I had a quick test against the Dart server, and everything seems to work correctly!

@DanTup  Thank you so much. Super happy to see everything is working as we expected. In that case, do you think we can seal our protocol proposal? Basically we agree the proposal are settled, no more changes in the proposed protocol. (we need to release vscode-go very soon, so it is better to have a green signal from you and make sure no more changes in the proposed protocol)

However, we can keep making changes in this CL as you have mentioned we may be able to have a cleaner solution using the feature registration mechanism. But this change is not a proposal change, it is more of a "library" change impacting how the extension author is using form.ts to enable interactive refactoring. If you could put your stamp on the protocol level, we will release the vscode-go next Monday. (Your stamp is very important :D for us).

> I've left some minor comments/questions below I'm interested in your feedback on. I'm happy to make some of these changes myself if you agree with them (although I don't want to diverge from your own implementation if the goal is for `form.ts` to be shared).

How about we get this merged then we can discuss further changes to this library through other PR. I hope we could keep a few goals in term of the lib:
- The form.ts should be language agnostic.
- The form.ts should allow extension author to specify middle ware for 3 LSP methods we defined.
- The form.ts should be easy to integrate with other extensions.

src/extension/analysis/analyzer.ts
DanTup: > Just to make sure I understand, we still want to allow "extension authors" to specify their middleware for "interactive execute command", "resolve command" & "interactive list enum". Even after migrating to the feature registration based solution, we still need to give extension authors a place to specify these middleware, is that correct?

I think this would still work, although I'm not sure if you'd get completion on the middleware object without a cast (because it would still be the default LSP middleware type). I don't know if that's a deal-breaker though. I would have to do some testing to verify though.

> In addition, I just noticed that you are using the head version of vscode-languageclient and I'm left behind, so the StaticFeature have some backward incompatible changes.

Ah, yes - we did this to get a new feature to prevent redundant didOpen/didClose events if you Ctrl+Hover over symbols.

> do you think you can help me create a new PR containing the diff and we can discuss in that CL?

Yes, absolutely! I'm out tomorrow, but on Monday I will merge this PR into a new branch for this feature, then it'll be easier for us to make smaller PRs to iterate on it (starting with a PR I'll open to make this change, so we can review and compare it).

src/extension/analysis/form.ts
DanTup: I like using "file"/"file or folder"/"folder" as you added in `resourceName`, it would be clearer than "entry"?

DanTup:
@h9jiang 

> In that case, do you think we can seal our protocol proposal? Basically we agree the proposal are settled, no more changes in the proposed protocol.

The only thing I was going to ask about after implementing some more in the server today, was the idea of adding an optional regex for validation. Currently I think the validation failures are not the best UX (`showWarningMessage`) and I think most cases where we fail validation, it will be be strings that need to be valid identifiers. If we added an optional regex, we could use the built-in validation in the input box, which would give real-time feedback and in the input box rather than in the bottom right corner.

But, this could be an optional field though (and could be specified as optional for clients to honour, so servers are also responsible for validating server-side using the current mechanism), so it would be completely non-breaking and therefore I don't think needs to hold up finalizing the current protocol and you shipping.

I think it's fair to say the spec as we have today is good, and any future changes need to be non-breaking (eg. optional on both sides, or guarded by capabilities).

h9jiang:
> I think it's fair to say the spec as we have today is good, and any future changes need to be non-breaking (eg. optional on both sides, or guarded by capabilities).

An optional field would be great. Thank you so much for your code / design review and your stamp on this.

> Currently I think the validation failures are not the best UX (showWarningMessage) and I think most cases where we fail validation, it will be be strings that need to be valid identifiers. If we added an optional regex, we could use the built-in validation in the input box, which would give real-time feedback and in the input box rather than in the bottom right corner.

We had an ideal of introducing something called "restriction" or "pattern". But did not land it here. As this is more of an "early" signal that is available to the user. The language server need to run further validation outside of simple pattern matching, like language specifics (like is the ident duplicate of existing, is the ident legal at this position...). Those validation can only be done through the server.

DanTup:
@h9jiang I'm going to merge this into a new branch (`interactive-forms`) so it's easier to iterate with additional PRs.

(I will also look at Gemini's latest feedback in a separate PR)

PR #6098, opened 2026-06-22
Add tests for Interactive Forms + minor validation tweaks

DanTup:
This adds some tests for `form.ts`, plus some minor changes to the validation for consistency, including:

- consider empty strings as "no answer" (eg. fail validation if required)
- treat empty inputs as "null" answer for optional fields
- validate individual parts of a list input

### DIFF
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -782,7 +782,28 @@ export class InteractiveFormsFeature implements StaticFeature {
 	}
 
 	/**
+	 * Validates a string, returning a user-facing error message if it's not valid.
+	 */
+	private validateString(text: string, required: boolean): string | null {
+		if (required && text.trim() === "")
+			return "Please enter a value";
+		return null;
+	}
+
+	/**
+	 * Validates a number, returning a user-facing error message if it's not valid.
+	 */
+	private validateNumber(text: string, required: boolean): string | null {
+		if (text.trim() === "")
+			return required ? "Please enter a number" : null;
+		return !Number.isFinite(Number(text)) ? "Please enter a valid number" : null;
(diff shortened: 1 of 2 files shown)


src/extension/analysis/form.ts
```diff
@@ -992,22 +1020,43 @@ export class InteractiveFormsFeature implements StaticFeature {
...
-						return undefined;
+						return undefined; // Cancelled.
 					}
 
-					// If empty input, return empty list
 					if (rawList.trim() === '') {
-						return [];
+						// Treat empty as no answer. We shouldn't get here if required because
+						// of validation.
+						return null;
```
DanTup, line 1053: @h9jiang I made this change to keep things consistent (submitting an empty input for an optional field always uses `null` for the answer), however if you've already shipped any server code with optional lists that expect non-null, this might be breaking. If so, we could revert this.

DanTup (commented):
@h9jiang @bwilkerson this PR is mostly adding some tests (which I've tried to keep standalone in case we want to lift this with `form.ts` to a central repo in future), however while adding them I noticed some slight inconsistencies with validation (empty strings, and lists) so there are also some minor changes in `form.ts`.

DanTup:
@h9jiang I'd also like your feedback on these changes to `form.ts`. There are also some tests, but those aren't likely to impact you in the same way as the form changes might.

h9jiang (approved):
I think the CL make perfect sense to me. 

Right now, the vscode-go test against the gopls directly. However the approach you are taking (testing against the dummy language server) is better. 

Thanks for coming up with the robust test!

DanTup:
Great, thanks for the review! 🙂

PR #6108, opened 2026-06-30
Add support for client-side regex validators in interactive forms

### LINKED ISSUE #6092: Support "real-time" validation for Interactive Forms
Currently validation of the new "Interactive Forms" is done after the form is complete and reports validation errors via `window.showWarningMessage`.

Here, I try to add a constructor name of `test test` which triggers the warning and forces me to change it to `test`:

https://github.com/user-attachments/assets/f764907d-dbe8-4033-ad32-450bc479612d

Two concerns I have are:

1. When the window is large, the validation message is a long way away from the input box I completed
2. When I correct the mistake and submit, the warning notification remains visible (because we have no API to dismiss it)

While this experience may be unavoidable in some cases, I think a very common case (validating some kind of identifier) could be improved by using the `validateInput` method on `InputBoxOptions`.

Some possible options (which I'll try to flesh out more below):

1. Call `/resolve` in `valudateInput` and just read the error message from the form (but this is weird because we'd ignore any other changes in the form)
2. Add a new request like `/resolve` that is `/validate`, for example `command/validate` which works like `/resolve` but **only** validates and does not modify the form
3. The form fields have optional validation properties that can include a regex and an error message if the input does not match
4. The form fields have optional validation properties that can include a method/command that should be called to validate that input (it would need to be defined whether the whole form or just this value is passed... the whole form is more flexible, but individual values would allow for more general simple validators like `validateClassName` and `validateConstructorName`)

(options 3+4 could also be combined, allowing a server to provide either)

DanTup:
Fixes https://github.com/Dart-Code/Dart-Code/issues/6092

### DIFF
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -856,7 +856,7 @@ export class InteractiveFormsFeature implements StaticFeature {
 	 * Validating a single value in this order means we can stop validating on
 	 * the first validation message.
 	 */
-	sortedValidators<T extends Validator>(validators: T[]): T[] {
+	private sortedValidators<T extends Validator>(validators: T[]): T[] {
 		return validators.slice().sort((a, b) => b.severity - a.severity);
 	}
 
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -488,6 +488,11 @@ export interface Validator {
 	 * The severity of a violation of this validator.
 	 */
 	severity: ValidationSeverity;
+
+	/**
+	 * The message to show if the answer fails this validator.
+	 */
+	message: string;
 }
 
 /**
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -822,13 +822,13 @@ export class InteractiveFormsFeature implements StaticFeature {
 	private validateString(text: string, fieldType: FormFieldTypeString, { required }: { required: boolean }): vscode.InputBoxValidationMessage | null {
 		if (text.trim() === '') {
 			return required
-				? { message: 'Please enter a value', severity: vscode.InputBoxValidationSeverity.Error }
+				? this.error('Please enter a value')
 				: null;
 		}
 
 		if (fieldType.validators) {
 			for (const validator of this.sortedValidators(fieldType.validators)) {
-				if (validator.kind === "regex") {
+				if (validator.kind === 'regex') {
 					let isMatch: boolean;
 					try {
 						isMatch = new RegExp(validator.pattern).test(text);
@@ -852,6 +852,9 @@ export class InteractiveFormsFeature implements StaticFeature {
 	/**
 	 * Sorts validators so that highest severity validators are first, but
 	 * within each severity they preserve the original order.
+	 *
+	 * Validating a single value in this order means we can stop validating on
+	 * the first validation message.
 	 */
 	sortedValidators<T extends Validator>(validators: T[]): T[] {
 		return validators.slice().sort((a, b) => b.severity - a.severity);
@@ -860,12 +863,16 @@ export class InteractiveFormsFeature implements StaticFeature {
 	/**
 	 * Validates a number, returning a user-facing error message if it's not valid.
 	 */
-	private validateNumber(text: string, { required, isList }: { required: boolean; isList: boolean; }): string | null {
-		if (text.trim() === "")
-			return required ? "Please enter a number" : null;
+	private validateNumber(text: string, { required, isList }: { required: boolean; isList: boolean; }): vscode.InputBoxValidationMessage | null {
+		if (text.trim() === '')
+			return required ? this.error('Please enter a number') : null;
 		return !Number.isFinite(Number(text))
-			? (isList ? "Please enter only valid numbers" : "Please enter a valid number")
+			? this.error(isList ? 'Please enter only valid numbers' : 'Please enter a valid number')
 			: null;
+	}
+
+	private error(message: string): vscode.InputBoxValidationMessage {
+		return { message, severity: vscode.InputBoxValidationSeverity.Error };
 	}
 
 	/**
@@ -1092,7 +1099,7 @@ export class InteractiveFormsFeature implements StaticFeature {
 						ignoreFocusOut: true,
 						validateInput: (text) => {
 							if (text.trim() === '') {
-								return field.required ? 'Please enter at least one item' : null;
+								return field.required ? this.error('Please enter at least one item') : null;
 							}
 							const parts = text.split(',').map((s) => s.trim());
 							if (fieldType.elementType.kind === 'string') {
--- a/src/extension/analysis/form.ts
+++ b/src/extension/analysis/form.ts
@@ -43,7 +43,9 @@ export interface FormFieldTypeString {
 	 * Field validators only validate the input/answer to a field in isolation
 	 * and cannot depend on the answers to other fields. For example a string
 	 * field may have a regex validator, or a numeric field may have a range
-	 * validator.
+	 * validator. Multiple validators of the same kind are allowed.
+	 *
+	 * Answers must pass all (supported) validators to be considered valid.
 	 *
 	 * Clients must ignore validators they do not support.
 	 */
@@ -461,6 +463,64 @@ export interface InteractiveListEnumMiddleware {
 		param: InteractiveListEnumParams,
 		next: InteractiveListEnumSignature
 	) => vscode.ProviderResult<FormEnumEntry[]>;
+}
+
+/**
+ * The severity of a violation of a validator.
+ */
+export enum ValidationSeverity {
+	/**
+	 * An informational message that does not block submission of the value.
+	 */
+	Info = 1,
+	/**
+	 * A warning message that does not block submission of the value.
+	 */
+	Warning = 2,
+	/**
+	 * An error message that prevents submission of the value.
+	 */
+	Error = 3,
+}
+
+export interface Validator {
+	/**
+	 * The severity of a violation of this validator.
+	 */
+	severity: ValidationSeverity;
+}
+
+/**
+ * Validators applicable to string fields.
+ */
+export type StringValidator = RegexValidator /* | FooValidator */;
+
+/**
+ * A regex-based validator that ensures an answer matches a given
+ * regex.
+ */
+export interface RegexValidator extends Validator {
+	kind: 'regex';
+
+	/**
+	 * The regex pattern to validate the input.
+	 *
+	 * The server must only provide regular expressions for the engine supported by the client
+	 * in the `general.regularExpressions` client capability. If the server cannot provide an
+	 * appropriate regular expression it should not provide the regex validator.
+	 */
+	pattern: string;
+
+	/**
+	 * Whether the answer matching the pattern means it is valid (no message
+	 * reported) or invalid (message reported).
+	 */
+	matchIsValid: boolean;
+
+	/**
+	 * The message to show if the answer is not valid according to `pattern` and `matchIsValid`.
+	 */
+	message: string;
 }
 
 export class InteractiveFormsFeature implements StaticFeature {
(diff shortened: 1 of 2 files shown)


DanTup:
@bwilkerson @h9jiang here is my initial proposal for client-side regex validation. Rather than just put TS into an issue, I made the changes in the code, so I could also verify what I'm proposing works. I'm happy to change as necessary to ensure we're all happy with it though.

With these changes, we can show errors and warnings for string fields (and also lists or strings), and only errors block submission.

Here's a video showing how it works (the server changes aren't done, I just cobbled together enough to test/demo). Sorry about the quality, Windows snippet recorder usually does better, but there's weird artifacts here.

https://github.com/user-attachments/assets/0c39f3c3-cb1a-447f-8da6-ed8658e5d4a0

src/extension/analysis/form.ts
```diff
@@ -784,19 +819,53 @@ export class InteractiveFormsFeature implements StaticFeature {
...
+	 */
+	sortedValidators<T extends Validator>(validators: T[]): T[] {
+		return validators.slice().sort((a, b) => b.severity - a.severity);
+	}
+
 	/**
 	 * Validates a number, returning a user-facing error message if it's not valid.
 	 */
-	private validateNumber(text: string, required: boolean): string | null {
+	private validateNumber(text: string, { required, isList }: { required: boolean; isList: boolean; }): string | null {
```
bwilkerson: Does the validation allow for both integer and floating point values? I'm guessing the answer is 'yes' and the server is responsible for rejecting kinds of numbers that it can't handle. (Which is fine, just checking my understanding.)

DanTup: The current implementation does allow either (this is the same before and after this PR, my changes in number validation here are mostly for consistency - although I see I didn't make it completely consistent, so I will tidy this up a bit more tomorrow).

We could add a validator kind for numbers for whole numbers similar to how we added a regex validator for strings here (we could also just add a flag on the number field, but maybe the validator would keep things a bit more general).

I would suggest we do add this if either of us start using numeric fields, because it's a nicer experience to have them validation client-side than only at the "end" of the form, even if it is just one input.

bwilkerson: I agree that we don't need to add them until we have a use case.

DanTup: I updated the test in b5933844 to include a decimal number so this behaviour is clearer.

I do wonder if this behaviour is obvious though - I suspect it just fell out of using JS/TS here - @h9jiang do you have an opinion on this? If you're not already using number (or you assumed it was integers), we could "fix" that now. Otherwise we can keep as-is and decide on how to restrict to whole numbers when someone first uses it.

DanTup:
I pushed some minor tweaks to make `validateNumber()` a bit more consistent with `validateString()` (always return the same structured types instead of bare strings).

I did consider rewriting the built-in validation parts (required-ness, and list handling) to just map on to validators the same way that custom ones do, but it was not that trivial (and also raises questions about the exposed types, since the validators would then include `kind`s that are not intended to be provided by servers), so I decided to abort that for now because I don't want to drag this out for what should be implementation details.

src/extension/analysis/form.ts
```diff
@@ -36,6 +37,17 @@ import {
...
+	/* Validators for this form field.
+	 *
+	 * Field validators only validate the input/answer to a field in isolation
+	 * and cannot depend on the answers to other fields. For example a string
+	 * field may have a regex validator, or a numeric field may have a range
+	 * validator.
+	 *
+	 * Clients must ignore validators they do not support.
+	 */
+	validators?: StringValidator[];
```
h9jiang, line 52: is this a `&` logic or a `|` logic if multiple validator exist. I assumed it is "and".

src/extension/analysis/form.ts
```diff
@@ -1072,3 +1158,37 @@ export class InteractiveFormsFeature implements StaticFeature {
 		}
 	}
 }
+
+// The severity of a violation of a validator.
```
h9jiang: is that possible to move the block up so all the type decl are together on top of the file.

src/extension/analysis/form.ts
```diff
@@ -1072,3 +1158,37 @@ export class InteractiveFormsFeature implements StaticFeature {
...
+
+// Validators applicable to string fields.
+export type StringValidator = RegexValidator /* | FooValidator */;
+
+// A regex-based validator that ensures an answer matches a given
+// regex.
+export interface RegexValidator extends Validator {
+	kind: 'regex';
+
+	// The (JS) regex pattern to validate the input.
```
h9jiang: The LSP has a quite strict rule of regex engine

see https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#regExp

I assume the regex pattern here will adhere to the client capability. It is not necessary to reference the regex client capabilities here (but it would be more clear if we can mention that).

src/extension/analysis/form.ts
h9jiang, line 52: do we allow multiple validator of the same kind? I assumed yes...

src/extension/analysis/form.ts
```diff
@@ -1072,3 +1158,37 @@ export class InteractiveFormsFeature implements StaticFeature {
...
+	pattern: string;
+
+	// Whether the answer matching the pattern means it is valid (no message
+	// reported) or invalid (message reported).
+	matchIsValid: boolean;
```
h9jiang: I understand the purpose of the field but I do find it confusing in combination with the `message` below.

The message below says, if the input does not match the pattern.  But if the matchIsValid == false, then message should be positive. if the matchIsValid == true, then message should be negative.

h9jiang (commented):
Thank you so much for raising this PR, I think I agree on idea of introducing the regex validator but the message and matchIsValid is a little confusing to me.

In addition, I would like Alan to review the change to the protocol.

src/extension/analysis/form.ts
DanTup: Correct, I've added this to the comment:

> Answers must pass all (supported) validators to be considered valid.

src/extension/analysis/form.ts
DanTup: Done!

src/extension/analysis/form.ts
DanTup: Oh, I was unaware of this capability and had assumed LSP was fixed. I've updated this comment to:

```
/**
 * The regex pattern to validate the input.
 *
 * The server must only provide regular expressions for the engine supported by the client
 * in the `general.regularExpressions` client capability. If the server cannot provide an
 * appropriate regular expression it should not provide the regex validator.
 */
```

Does this make sense to you?

src/extension/analysis/form.ts
DanTup: Yep, I've added:

> Multiple validators of the same kind are allowed.

src/extension/analysis/form.ts
DanTup: Good catch - I wrote the comment on `message` before I added `matchIsValid`. I've updated the comment on `message` now to:

> The message to show if the answer is not valid according to `pattern` and `matchIsValid`.

DanTup:
@h9jiang thanks for the review! I've addressed your comments, please let me know if you think these changes are sufficient.

> In addition, I would like Alan to review the change to the protocol.

Of course - if you're otherwise happy, do you want to ping him here, or have him review it as a change in the Go repo? Just let me know if I need to change anything or if you're (collectively) happy with this so I can merge here. Thanks!

src/extension/analysis/form.ts
DanTup: Also FYI, I didn't realise this capability was new in LSP 3.18, so this does mean you'll need to be on the latest LSP client for the server to be passed this. I think it's probably better to require this than to make servers assume the client knows ES regex if not provided. Dart-Code is already on the latest LSP client.

src/extension/analysis/form.ts
```diff
@@ -431,6 +465,69 @@ export interface InteractiveListEnumMiddleware {
...
+	 */
+	message: string;
+}
+
+/**
+ * Validators applicable to string fields.
+ */
+export type StringValidator = RegexValidator /* | FooValidator */;
+
+/**
```
adonovan, line 503: Hongxiang asked me to take a look. This CL seems to be mostly a Dart reflection of the agreed protocol changes, so I'm not qualified to say anything intelligent about the code. About the protocol however:

Regular expressions vary between implementations. (Even RE2, used by Go, varies across languages since the original C++ implementation is defined to operate on a byte stream wheres the Go and Java ports operation on a sequence of Unicode code points, so the "any byte" operator has no meaning.)

The LSP of course acknowledges this in its [RegularExpressionsClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#regExp) feature, but it's not clear to me how that feature can realistically work, as it expects both peers (client and server) to check the regular expression engine in use and either support it, emulate it, or (more likely) have a go at at it with the wrong engine, or just refuse to do anything. Given that most (client, server) pairs have a language barrier, it seems almost doomed not to work.

I don't think our protocol changes are really making the LSP regexp situation worse than it already is, but that's my one concern with our use of regexp validation.

DanTup: @adonovan thanks for taking a look!

The validation here is by design entirely optional - it's only for improving the user experience and if either the server or the client dropped the validation, nothing bad should happen (since the server would do the normal server-side validation when the whole form is submitted).

So, my intention was that a server should only provide regular expression validators if it can do so in the format supported by the client (in `RegularExpressionsClientCapabilities`). For the VS Code LSP client, that's ES2020. If the client uses a different regex engine, then the server should either provide the regex patterns in a compatible form, or not provide them at all. (if this isn't clear from the code, I should definitely make this clearer).

It'd be nice if there was a better story for this, but I think it will be difficult to solve more generally (without implemented some very specific rules in the spec, or some parallel to regex) - I'm definitely open to ideas though :-)

DanTup: > This CL seems to be mostly a Dart reflection of the agreed protocol changes, so I'm not qualified to say anything intelligent about the code.

btw, I believed the goal was to share the `form.ts` file, so I think all changes here are potentially relevant for Go. It's not intended that anything here is Dart-specific (although my motivation is to support this in the Dart server).

src/extension/analysis/form.ts
h9jiang: I see, in that case, I need to also upgrade the client to lsp 3.18, but the gopls latest release v0.23.0 is using 3.17 so I will make sure to upgrade the vscode-go to 3.18 after the current vscode-go release so it can work with the next gopls release v0.24.0.

> Does this make sense to you?

Thanks. Yes this make sense to me.

h9jiang (commented):
Hi Danny, I think this looks good. 

I will copy the code form.ts to vscode-go and do some debugging once you merge it. Thank you.

h9jiang (approved):

DanTup:
@h9jiang thanks! I've merged the client changes here and the Dart server changes are at https://dart-review.googlesource.com/c/sdk/+/524020.

src/extension/analysis/form.ts
adonovan: Agreed, I don't think there's much we can do; I just think it is wildly unrealistic for clients and servers to do anything other than execute their own dialect of regular expressions and hope that most server-provided expressions fall in the least-common denominator subdialect.

> the goal was to share the form.ts file, so I think all changes here are potentially relevant for Go

Yeah, I understand that our tools will share and execute parts of the code, I just don't know enough TS to review them competently.

bwilkerson: > I just think it is wildly unrealistic for clients and servers to do anything other than execute their own dialect of regular expressions and hope that most server-provided expressions fall in the least-common denominator subdialect.

I'll just point out that it isn't a requirement that the client and server use the same regular expression, only that they use equivalent regular expressions. It isn't even a requirement that the server use a regular expression at all, though in most cases that will be the easiest way to keep the client and server consistent with each other.

DanTup: I also think it's best for servers to not return these regex validators if the clients regex engine is not one that is likely to be compatible the the expressions provided. In the Dart server we're only going to return these optional validators if the client uses the ECMAScript regex engine, because that's what they're written/tested for. This means clients that use other regex engines will fall back to the original server-validation only.

If there are significant LSP clients that are not using ECMAScript, nothing prevents the server from returning a different set of regular expressions for that engine for those clients (or if the expressions happen to be compatible, the same one).

@bwilkerson perhaps it's worth capturing the regexp engine from client capabilities in analytics?

bwilkerson: That makes sense to me. There are probably other capabilities that it would be useful to capture as well. We should talk off-line. Let me know when you're available.

PR #6135, opened 2026-07-30
Add a command to toggle TODO visibility

### LINKED ISSUE #6118: Create a command for toggling `dart.showTodos` value programatically
When trying out new lints, I often want to see just that lint on the problems tab so I know how many cases there would be for me to fix, and having to open the settings every time to toggle that value to `false` and back to the default again is annoying.

**Describe the solution you'd like**
Create a command for toggling `dart.showTodos` value programatically. So we don't have to keep opening the settings json to edit the value.

Always manually opening the settings json to edit the value for this setting.

DanTup:
This only toggles as a bool though, so you can't set it to show TODO without HACK etc.

Fixes https://github.com/Dart-Code/Dart-Code/issues/6118

### DIFF
--- a/package.json
+++ b/package.json
@@ -111,6 +111,11 @@
 		],
 		"commands": [
 			{
+				"command": "dart.toggleShowTodos",
+				"title": "Toggle Show TODOs",
+				"category": "Dart"
+			},
+			{
 				"command": "flutter.createProject",
 				"title": "Create New Project",
 				"category": "Flutter"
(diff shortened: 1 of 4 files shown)


DanTup:
> The command and multi-root test currently don’t correctly handle/validate workspace-folder scoped settings due to missing resource scoping for configuration inspection/updates.

`showTodos` is window-scoped, so it'll never be workspace-folder scoped.

Issue #5420, opened 2025-02-19
Opening XCode fails


yahu1031:
**Describe the bug**
Opening XCode from "Open in XCode" option fails to open XCode if it was not identified in the terminal/ if XCode is installed in External Drive.

Making use of `-a` argument in open command would fix it.
```bash
open -a XCode <iOS PROJECT PATH>
```

**To Reproduce**
1. Install Xcode in External drive
2. Try opening `ios project` via Termianl/VSCode

**Expected behavior**
Should open XCode

**Please complete the following information:**

**!! ⚠️ PLEASE REVIEW THIS REPORT FOR SENSITIVE INFORMATION BEFORE SHARING ⚠️ !!**

<details>

<summary><strong>Workspace Environment</strong></summary>

```text
Dart Code extension: 3.104.0
Flutter extension: 3.104.0 (activated)

App: Trae
App Host: desktop
Version: mac 1.96.4

Workspace type: Flutter (LSP)
Workspace name: nlearnv4

Dart (3.7.0): /opt/homebrew/Caskroom/flutter/3.29.0/flutter/bin/cache/dart-sdk
Flutter (3.29.0): /opt/homebrew/Caskroom/flutter/3.29.0/flutter (iPhone 16 Pro Max (ios/ios))

```

</details>

<details>

<summary><strong>Output from 'dart info'</strong></summary>

`/opt/homebrew/Caskroom/flutter/3.29.0/flutter/bin/cache/dart-sdk/bin/dart info`

If providing this information as part of reporting a bug, please review the information
below to ensure it only contains things you're comfortable posting publicly.

#### General info

- Dart 3.7.0 (stable) (Wed Feb 5 04:53:58 2025 -0800) on "macos_arm64"
- on macos / Version 15.3.1 (Build 24D70)
- locale is en-US

#### Project info

- sdk constraint: '^3.5.3'
[...]

PR #6160, opened 2026-08-31
Open Xcode by bundle ID rather than app name

### LINKED ISSUE #6159: Open in Xcode does nothing when Xcode is installed as Xcode-beta.app
**Describe the bug**

`Flutter: Open in Xcode` silently does nothing when the only installed Xcode is `/Applications/Xcode-beta.app`.

`openInXcode` launches Xcode by **application name**:

https://github.com/Dart-Code/Dart-Code/blob/master/src/extension/commands/open_in_other_editors.ts

```ts
const file = path.join(folder, files[0].name);
safeToolSpawn(folder, "open", ["-a", "XCode", file]);
```

`open -a` resolves via LaunchServices by app name, and a beta Xcode is installed as `Xcode-beta.app`, so the lookup fails:

```console
$ open -a "XCode" ios/Runner.xcworkspace
Unable to find application named 'XCode'
$ echo $?
1
```

The `safeToolSpawn` result isn't checked, so nothing surfaces in the UI — the context-menu item just appears to do nothing. The `.xcworkspace`/`.xcodeproj` discovery above it works correctly; only the launch fails.

This affects anyone on an iOS/macOS beta, who generally cannot install a stable Xcode alongside it for the current SDK.

**To Reproduce**

1. Have `/Applications/Xcode-beta.app` as the only installed Xcode (no `/Applications/Xcode.app`).
2. Open a Flutter project.
3. Right-click the `ios` folder in the explorer → **Open in Xcode**.
4. Nothing happens, and no error is shown.

**Expected behavior**

Xcode opens the project, or an error is shown if it genuinely can't be found.

**Proposed solution**

Launch by **bundle ID** instead of app name. Both `Xcode.app` and `Xcode-beta.app` declare `com.apple.dt.Xcode`:

```console
$ defaults read /Applications/Xcode-beta.app/Contents/Info.plist CFBundleIdentifier
com.apple.dt.Xcode
```

```diff
-		safeToolSpawn(folder, "open", ["-a", "XCode", file]);
+		safeToolSpawn(folder, "open", ["-b", "com.apple.dt.Xcode", file]);
```

Verified locally (patched into the installed extension bundle) — it opens `Runner.xcworkspace` in Xcode beta correctly.

`-b` goes through the same LaunchServices database as `-a`, just keyed on bundle ID rather than name, so this should preserve the fix from #5420 (Xcode installed on an external drive) while also handling renamed/beta installs. It also avoids depending on `open -a` being case-insensitive for `"XCode"`.

Happy to raise a PR for this if the approach looks right.

**Please complete the following information:**

 - Operating System and version: macOS 27.0 (26A5421a)
 - VS Code version: Cursor 3.12.30
 - Dart extension version: 3.140.0
 - Dart/Flutter SDK version: Flutter 3.47.2 (stable), Dart 3.13.2
 - Xcode version: 27.0 beta (27A5237l), installed as `Xcode-beta.app`
 - Target device (if the issue relates to Flutter debugging): n/a

10xtushar:
Fixes #6159

`openInXcode` currently launches Xcode by application name:

```ts
safeToolSpawn(folder, "open", ["-a", "XCode", file]);
```

`open -a` resolves via LaunchServices by app name, so this fails when the only installed Xcode is `Xcode-beta.app`:

```console
$ open -a "XCode" ios/Runner.xcworkspace
Unable to find application named 'XCode'
$ echo $?
1
```

Because the `safeToolSpawn` result isn't checked, nothing surfaces in the UI and the context-menu item appears to do nothing at all. This affects anyone working against an iOS/macOS beta, who generally can't install a stable Xcode for the current SDK alongside the beta.

This switches to launching by bundle ID, which both `Xcode.app` and `Xcode-beta.app` declare:

```console
$ defaults read /Applications/Xcode-beta.app/Contents/Info.plist CFBundleIdentifier
com.apple.dt.Xcode
```

`-b` uses the same LaunchServices database as `-a`, just keyed on bundle ID rather than name, so it should continue to work for Xcode installed outside `/Applications` (#5420). It also removes the reliance on `open -a` being case-insensitive for `"XCode"`.

The `.xcworkspace`/`.xcodeproj` discovery above this line is unchanged — only the launch call differs.

**Testing:** verified by patching the compiled `extension.js` in an installed 3.140.0 extension and restarting the editor; **Open in Xcode** then correctly opens `Runner.xcworkspace` in Xcode 27 beta. I haven't run the full extension test suite locally — happy to if you'd like, though there's no existing coverage for `openInXcode`.

### DIFF
--- a/src/extension/commands/open_in_other_editors.ts
+++ b/src/extension/commands/open_in_other_editors.ts
@@ -60,7 +60,7 @@ export class OpenInOtherEditorCommands implements IAmDisposable {
 		}
 
 		const file = path.join(folder, files[0].name);
-		safeToolSpawn(folder, "open", ["-a", "XCode", file]);
+		safeToolSpawn(folder, "open", ["-b", "com.apple.dt.Xcode", file]);
 	}
 
 	private async getAndroidStudioDir(folder: string): Promise<string> {


DanTup (approved):
Thanks!

DanTup:
Bot breakage is unrelated, there was an issue on master when this was branched that is since fixed.

Thanks for the PR!

